PR #8, opened 2024-07-05
Whitelist window

Skyedra:
This patch released under MIT, so if any other server wants to take and adapt for their own purposes, feel free!


Skyedra:
Oh, oops, just realized it doesn't wordwrap if the whitelist message is wrong, will fix

Skyedra:
Ok, fixed the wordwrap ^.^

FoxxoTrystan:
Next time please fill out the description so its would be easier to understand what its does.

Also from looking its just add a whitelist window while also adding a new category to advertize others servers?

Skyedra:
Check with floofmin or dm me on discord :)

PR #69, opened 2024-07-27
Add Thieving Trait Back to Felinids

ghost:
# Changelog

:cl:
- add: add thieving trait back to felinids


FoxxoTrystan:
This is compleatly undeeded, this check was here because felinds prototype by DEFAULT have Theiving, so they dont need to double take the trait, this may cause issues.

FoxxoTrystan:
nevermind its seem the prototype does not have the trait by default.

https://github.com/Fansana/floofstation1/blob/master/Resources/Prototypes/Nyanotrasen/Entities/Mobs/Species/felinid.yml

PR #72, opened 2024-07-31
Init Space Law

ghost:
This PR will bring spacelaw into the game


DJB1gYAPPA:
aggravating and aggression both have no definitions, is this on my end?

PR #84, opened 2024-08-07
Add Craftable Bonfire

ghost:
# Description

This PR makes bonfires craftable

---

# Changelog

:cl:
- add: make bonfires craftable

### DIFF
--- a/Resources/Prototypes/Entities/Structures/Decoration/bonfire.yml
+++ b/Resources/Prototypes/Entities/Structures/Decoration/bonfire.yml
@@ -1,9 +1,12 @@
 - type: entity
-  id: Bonfire
+  id: BonfireLit
   parent: BaseStructure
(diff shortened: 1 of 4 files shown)


Resources/Prototypes/Entities/Structures/Decoration/bonfire.yml
```diff
@@ -1,9 +1,12 @@
 - type: entity
-  id: Bonfire
+  id: BonfireLit
```
FoxxoTrystan, line 2: Bonefire may already be mapped, changing it may cause mapping issues, this should be migrated.

ShadesMars (changes requested):
The only issue i found is the one i mentioned.

I have not tested it and did the correct test i wanted, il do that after scubadiving.

PR #95, opened 2024-08-10
Increases Trait Limit From 5 to 7

ShatteredSwords:
# Description

This PR hopefully changes the default trait limit from 5 to 7. The amount of points you get is left the same. The reason that I am changing it is due to the new lewd trait system. I believe that this server is roleplay-focused enough to where this will not be abused heavily. I believe the point costs of each trait is what balances them all. This is essentially so characters can have tits and a cock along with their normal trait loadouts.

---

# Changelog

:cl:
- tweak: Trait limit changed from 5 to 7.

### DIFF
--- a/Content.Shared/CCVar/CCVars.cs
+++ b/Content.Shared/CCVar/CCVars.cs
@@ -361,7 +361,7 @@ namespace Content.Shared.CCVar
         ///     How many traits a character can have at most.
         /// </summary>
         public static readonly CVarDef<int> GameTraitsMax =
-            CVarDef.Create("game.traits_max", 5, CVar.REPLICATED);
+            CVarDef.Create("game.traits_max", 7, CVar.REPLICATED);
 
         /// <summary>
         ///     How many points a character should start with.


FoxxoTrystan:
The Lewd Trait has points fixed to 0, so its should not inpact this.

This PR is marked as needed discussion.

ShatteredSwords:
> The Lewd Trait has points fixed to 0, so its should not inpact this.
> 
> This PR is marked as needed discussion.

This is not a point issue, but a trait limit issue.

ShatteredSwords:
> The Lewd Trait has points fixed to 0, so its should not inpact this.
> 
> This PR is marked as needed discussion.

WithLewdTrait
withoutlewdtrait
As you can see, the lewd traits still take up a trait slot. Without it I have 4/5 max traits, but with it, I have 5/5.

Fansana:
Ah, thanks I wanted to do this myself but didn't get to it.

FoxxoTrystan:
This should have been commented to avoid merge conflicts mistakes.

VMSolidus:
Actually you should not need to worry at all about commenting CCVar changes! Especially not a change like this. 
Also, you don't need to change the CCVars file at all, you can instead put edits like this in Floof station's TOML file!

FoxxoTrystan:
Oh yeah i forgor about that.

PR #102, opened 2024-08-12
Longer Lasting Aphrodisiacs

ShatteredSwords:
# Description

This changes all the aphrodisiacs metabolism rates to 0.03u/second. This makes 10u of each last longer than 5 minutes.

---

# Changelog

:cl:
- tweak: Our chaplain prayed to the Deity of Lust, and now aphrodisiacs last much much longer

### DIFF
--- a/Resources/Prototypes/Floof/Reagents/Aphrodisiacs/medicine.yml
+++ b/Resources/Prototypes/Floof/Reagents/Aphrodisiacs/medicine.yml
@@ -8,7 +8,7 @@
   color: "#45072f"
   metabolisms:
     Medicine:
-      metabolismRate : 0.5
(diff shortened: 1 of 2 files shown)


SleepyScarecrow:
this makes 30u last like 16.33 repeating minutes just for context

SleepyScarecrow:
i previously 'fixed' this when I first wrote aphros, but that was before I made aluminum flush the system, this should be fine now

ShatteredSwords:
> i previously 'fixed' this when I first wrote aphros, but that was before I made aluminum flush the system, this should be fine now

Awesome, good to know

FoxxoTrystan (approved):

PR #103, opened 2024-08-12
Content.Integration Works Again

ShatteredSwords:
# Description

Brings spider_web.yml to modern EE standards, and completely yeets an unused unloved config file. This will not have a CL, because it's bug busting. If this does not work, I might cry myself to sleep.

---


FoxxoTrystan (approved):

FoxxoTrystan:
Double checked files, and yep, this works, big thanks, merging and re-lunching tests from others PRs.

Thank you and now Build & Test Debug would need to be taken correctly into account.

PR #104, opened 2024-08-12
Add: recipe for making cream from milk

neuPanda:
Description
Add Recipe to make cream from milk

<p>
image
</p>
</details>
:cl:
- add: Added way to make Double Cream Blaster with all natural creams

### DIFF
--- a/Resources/Prototypes/Floof/Recipes/Reactions/drinks.yml
+++ b/Resources/Prototypes/Floof/Recipes/Reactions/drinks.yml
@@ -455,3 +455,13 @@
       amount: 1
   products:
     YeOlHandy: 5
+    
+- type: reaction
+  id: Cream
+  reactants:
+    Sugar:
+      amount: 5
+    Milk:
+      amount: 10
+  products:
+    Cream: 15


ShatteredSwords:
This miiiight get in the way of making cream of coconut in a shaker. 
- type: reaction
  id: CreamOfCoconut
  reactants:
    CoconutWater:
      amount: 1
    Sugar:
      amount: 1
    Milk:
      amount: 1
  products:
    CreamOfCoconut: 3

(I'm not sure how to directly reference the lines of code so here you go)
I don't think CreamOfCoconut is a widely used drink like, at ALL, so it's probably fine.

FoxxoTrystan:
Can you try to make CreamOfCoconut with this PR checkout?

And tell the results?

neuPanda:
> Can you try to make CreamOfCoconut with this PR checkout?
> 
> And tell the results?

missed this one.  rather than seeing how it works and running into the issue we have in chem where if you dont put things in the right order you get fuckups.  just make it so that the coconut cream uses cream and sugar rather than milk and sugar but in a different balance in conjunction with the coconut water to be able to get the same output.  so 1:1:1 ratio is the same.

i just realized bars do not have sugar...

CHEM!!!! I NEED SUGAR!!!!

ShatteredSwords:
> > Can you try to make CreamOfCoconut with this PR checkout?
> > And tell the results?
> 
> missed this one. rather than seeing how it works and running into the issue we have in chem where if you dont put things in the right order you get fuckups. just make it so that the coconut cream uses cream and sugar rather than milk and sugar but in a different balance in conjunction with the coconut water to be able to get the same output. so 1:1:1 ratio is the same.
> 
> i just realized bars do not have sugar...
> 
> CHEM!!!! I NEED SUGAR!!!!

Bars do have sugar, it's in the soda dispenser.

FoxxoTrystan (approved):

PR #107, opened 2024-08-14
Loadout Point Increase And Labcoat For All

ShatteredSwords:
# Description

Adds the labcoat for everyone (the completely white one with no identifying colors), and increases the loadout points to 20. 
**(This NEEDS discussion unless you both agree on it)**

---

# Changelog

:cl:
- add: The lab coat is now available for everyone, because you can be a scientist without working in Epistemics.
- tweak: Loadout points have been increased from 14 to 20.

### DIFF
--- a/Resources/Prototypes/Floof/Loadouts/outerClothing.yml
+++ b/Resources/Prototypes/Floof/Loadouts/outerClothing.yml
@@ -12,8 +12,8 @@
   items:
     - ClothingOuterRedRacoon
 
-- type: loadoutOuterCoatLab # Science is for everyone, not just science
-  id: LoadoutOuter
+- type: loadout # Science is for everyone, not just science
+  id: LoadoutOuterCoatLab
   category: Outer
   cost: 2
   items:
--- a/Resources/Prototypes/Floof/Loadouts/outerClothing.yml
+++ b/Resources/Prototypes/Floof/Loadouts/outerClothing.yml
@@ -11,3 +11,10 @@
   cost: 3
   items:
     - ClothingOuterRedRacoon
+
+- type: loadoutOuterCoatLab # Science is for everyone, not just science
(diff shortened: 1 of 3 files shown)


Resources/Prototypes/Floof/Loadouts/outerClothing.yml
```diff
@@ -11,3 +11,10 @@
   cost: 3
   items:
     - ClothingOuterRedRacoon
+
+- type: loadoutOuterCoatLab # Science is for everyone, not just science
+  id: LoadoutOuter
```
FoxxoTrystan, line 16: Something tells me you put type and id in the wrong way.

FoxxoTrystan (changes requested):
Tail Fails are legits, this need to be fixed.

FoxxoTrystan (approved):

Fansana:
Dunno I don't see a reason to bump the load out points right now.

ShatteredSwords:
> Dunno I don't see a reason to bump the load out points right now.

Currently for the sake of more customization (while EE gets the exclusion groups up) and because I am going to be adding a lot more roleplay-oriented items to Floof itself

ShatteredSwords:
> > Dunno I don't see a reason to bump the load out points right now.
> 
> Currently for the sake of more customization (while EE gets the exclusion groups up) and because I am going to be adding a lot more roleplay-oriented items to Floof itself

That being said, it's fine if you disagree, it's a pretty easy to just PR the lab coat :P

Fansana:
Mmmhh, I mean upping it a little shouldn't hurt. 16 would be fine for now I think. aside from that we can adjust the points required for certain items.

ShatteredSwords:
Please be sure to edit the CL next time, but all good, ty

PR #120, opened 2024-08-19
Leashes

Mnemotechnician:
# Description
This PR adds leashes and the necessary logic for them.

Leash normally is an item craftable in the autolathe from some cloth, steel, and plastic. In order to attach the leash, you need an entity with a LeashAnchor or a clothing that has a LeashAnchor. An example of the former is most animals and all station pets - you can leash them without any tools. An example of the latter is all collars - in order to attach a leash to them, you need to open the strip menu and right-click the anchor item while holding a leash. The leash can be detached by right-clicking the person to whom it's attached. If it's attached via clothing, the respective clothing cannot be removed until the leash is removed (otherwise you get weird behavior).

There are still some bugs, e.g. you can pick up a leashed person, and then the leash joint will stop working until they're put back down, but they should be minor enough (the leash will automatically break if the person manages to get too far away, or moves to a different map via ftl or otherwise, or gets deleted, etc).

# Note
If someone has a better icon sprite, or wants to create inhand sprites, I'll gladly accept those. The current icon is more of a placeholder, since I'm by no means an artist.

<p>

image

https://github.com/user-attachments/assets/84af85d3-b90d-4cc7-bba8-13aea6eb4dad

</p>
</details>

# Changelog
:cl:
- add: A new leash item has been added to autolathe to help you keep your animals and station pets in check. If the station pet in question has hands and can speak, the leash can be attached to their collar via strip menu.


Content.Shared/Floofstation/Leash/Events/LeashAttachDoAfterEvent.cs
FoxxoTrystan, line 1: Can Leash Event not be a single file, pretty sure thats possible.

FoxxoTrystan: Example: https://github.com/Fansana/floofstation1/blob/master/Content.Shared/Forensics/Events.cs

FoxxoTrystan (changes requested):
All looks gud.
Request some suggestions for event and you should look in Floof ENT/Mobs so you can also collar the scug.

Afterall, good job lad!

Mnemotechnician:
Nevermind, now I did.

0x6273:
Found a bug:
- Leash a mob.
- Put the leash item in your bag.
- Unequip the bag and drop it on the ground.
- Visually, the leash is connected to the bag on the ground, but moving around still pulls the mob.
- Control Mob the mob that's leashed and move around and it causes the person that had the leash to move around, even tho it's in a bag that's on the ground.

Memeji (changes requested):
Absolutely loving the progress so far!
Only request I have, is to make the behavior of the person being pulled to look more seamlessly. Instead of them locking to a tile, if possible.
Besides that, looks great!

Mnemotechnician:
> Found a bug: <...>

This is really odd... The bug only happens when you put the leash into an equipped bag and take it off afterwards. This seems like an engine issue; I'll try to recreate the joint whenever the leash gets re-parented...

Mnemotechnician:
Yea no, I cannot fix that. This is an issue with the physics engine, which is out of my scope... However, the leash will still break if the leash entity and the leashed entity get too far away from each other, which compensates it for a bit...  If you can suggest a solution, besides simply disallowing putting leashes into storages (note that they work correctly so long as said storage is not equipped clothing), I can apply it.

FoxxoTrystan (approved):

Memeji (approved):

PR #139, opened 2024-08-23
Couch and Stripper Pole!

Memeji:
# Description

Adds a craftable couch and stripper pole!

---

# Changelog

:cl:
- add: Stripper Pole
- tweak: Corp Couche can be crafted


FoxxoTrystan (approved):
Eggs are kinda unsued and i would have prefer them in another PR that include them fully but okay.

PR #148, opened 2024-08-24
No Zombies For Now

ShatteredSwords:
# Description

This PR simply comments out the Zombie gamemode from appearing in Secret, and it comments out the Zombie outbreak that can happen at ANY time after an hour in survival and possibly extended. The gamemode still exists, but it wouldn't happen without admeme intervention. I believe this would be nice for the sake of this server in particular, as we're more on the roleplay side of things. Unclonable people are also just round removed when they become a zombie. Disaster roleplay could be "fun" but there are better ways to do that than zombies. 

---

# Changelog

:cl:
- remove: Zombies are gone from this sector, for now...

### DIFF
--- a/Resources/Prototypes/GameRules/events.yml
+++ b/Resources/Prototypes/GameRules/events.yml
@@ -429,23 +429,23 @@
     - id: MobGiantSpiderVampireAngry
       prob: 0.01
 
-- type: entity
-  id: ZombieOutbreak
-  parent: BaseGameRule
-  noSpawn: true
-  components:
-  - type: StationEvent
-    earliestStart: 60
-    reoccurrenceDelay: 60
-    minimumPlayers: 15
-    weight: 2
-    duration: 1
-  - type: ZombieRule
-    minStartDelay: 0 #let them know immediately
-    maxStartDelay: 10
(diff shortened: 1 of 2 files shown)


ShatteredSwords:
What the hell is causing all of these "delete all then ghost" build and test errors? What is the entity that survives? Is it you, the ghost?

VMSolidus:
> What the hell is causing all of these "delete all then ghost" build and test errors? What is the entity that survives? Is it you, the ghost?

Current nonsense Heisentest fail, it can be ignored. We currently have zero clue what it is, why its *completely randomly* test failing, or any way to fix it.

ShatteredSwords:
> > What the hell is causing all of these "delete all then ghost" build and test errors? What is the entity that survives? Is it you, the ghost?
> 
> Current nonsense Heisentest fail, it can be ignored. We currently have zero clue what it is, why its _completely randomly_ test failing, or any way to fix it.

Yeah seems about right

FoxxoTrystan (approved):
Changes code wise are correct, This will need approval and discussion from my lad.

Memeji (approved):

PR #172, opened 2024-09-02
Fix Capitalization For Items/Trait

cynical24:
# Description

The default naming scheme is like, all lowercase for items I believe (and it also looks weird compared to the other items in the vendor). Also made Hollow Fangs capital, cause I'm weird and silly.


ShatteredSwords:
Based. While you're at it could you fix Glass Jaw? I believe I forgot to change a 10 to a 7 in the Localization

FoxxoTrystan (approved):

Memeji (approved):

PR #184, opened 2024-09-11
Harpy Instrument Additions

cynical24:
# Description

Adds three new harpy instruments: Piano, Church Organ, and Harp.

---

# Changelog

:cl:
- add: three new harpy instruments


Content.Server/Instruments/SwappableInstrumentComponent.cs
```diff
@@ -11,4 +11,11 @@ public sealed partial class SwappableInstrumentComponent : Component
     /// </summary>
     [DataField("instrumentList", required: true)]
     public Dictionary<string, (byte, byte)> InstrumentList = new();
+
+    // Frontier: harpy instruments
+    /// <summary>
+    /// When true, only the instrument entity itself can swap its sound.
+    /// </summary>
+    [DataField]
+    public bool OnlySetBySelf = true;
```
Mnemotechnician, line 20: This is true by default, meaning that ALL swappable instruments are affected by this change. You didn't modify the yaml prototypes of other instruments, meaning that all swappable instruments (synthesizers, guitars, etc) are affected by this change and cannot be swapped by anyone unless they get possessed by an admin...

You should make it false by default, and set it to true on MapInit in SingerSystem instead.

Mnemotechnician (changes requested):

Content.Server/Instruments/SwappableInstrumentComponent.cs
cynical24: okie dokie

Fansana:
just some general advice, try writing commit messages that are a bit more on topic, later on that might help to narrow down where an important change happend.

Memeji:
What's the status on this PR?

FoxxoTrystan (approved):

ShatteredSwords (approved):
All 3 instruments work after playesting

PR #195, opened 2024-09-20
Remove Loadout Restriction

ghost:
# Description
Removed loadout restrictions for command and security

Description.

---

# Changelog

:cl:
- remove: Removed loadout restrictions for command and security


ShatteredSwords (changes requested):
The best thing to do instead of removing all of those lines is to comment all of them out, just so we don't have to deal with a headache when we pull from upstream and something in the loadout changes

PR #200, opened 2024-09-23
Xenospawners

fenndragon:
# Description

actually added event spawners for quite a few different things that did not have them to the vent critters events 

---

# Changelog

:cl:
- add: Added multiple event spawners for various things like meat, reagent slimes, and even argocytes


Memeji:
We do not want to add these mobs to the vents.

Reagent slimes are for the anomaly.
Flesh creatures are spawned by the anomaly.
Neutral Xenos should only be spawned by admins at their discretion.
Space ticks and Argocyte are on asteroids.

PR #210, opened 2024-09-25
New Mini-Hypospray From Frontier

XavierSomething:
# Description

This adds a mini-hypospray from Frontier that Epistemics can research for medical to use. It's only 5u with a usage delay, so you can only draw and inject 5u of one chemical every 2.5 seconds, which would make malicious usage hard, unless you have like, 4 of them, but then that's a whole other issue of how they got 4 (unless they were in epistemics).

This would add a link between medical and epistemics that would make shifts more interesting, as epistemics has links to every other department besides medical. While Epi CAN research stasis beds, cryopods, and cryostasis tools, not many people would ask for those anyways, so it makes gameplay more dynamic, I think. And I like having my own mini hypo as a medical main.

While I DO think this would be amazing to have as medical, I can already see some balancing concerns, so discussion of how to handle this instead of making all the decisions myself is very much needed.

---

# TODO

- [x] Port over frontier's mini hypo that works as intended.
- [x] Figure out balancing in where to put hypo in epistemic reseach
- [x] Put the hypospray in decided research or it's own research
- [x] Make the actual hypospray recipe (also requires balancing so someone can't make 5 of them and go to town)
- [x] Make sure it's in the dynamic recipes of the lathes

---

<p>

MiniHypo

</p>
</details>

---

# Changelog

:cl: Xavier
- add: New mini hypospray that epistemics can research!


VMSolidus:
I'm apparently gaslighting myself or having a boomer moment, because I could have swore this was PR'ed to EE lmao, and then double checked and realized I was wrong.

XavierSomething:
Made a thread within the discord in "current things worked on" for more wider discussion with the community on balancing, but discussions can be held here if necessary.

Eagle0600:
I'd say that while this limitation makes malicious use hard, it also makes its core functionality very limited. I'd say 10u (with the same usage amount) would be better, as it would allow doctors to load and inject a typical dosage without needing to inject half, run back to the medicine cabinet, and then run back to the patient to inject the other half. You don't want the doctor to have to bring the jug with them (this causes other issues).

VMSolidus:
> I'd say that while this limitation makes malicious use hard, it also makes its core functionality very limited. I'd say a 10u reservoir (with the same usage delay) would be better, as it would allow doctors to load and inject a typical dosage without needing to inject half, run back to the medicine cabinet, and then run back to the patient to inject the other half. You don't want the doctor to have to bring the jug with them (this causes other issues).

From experience, I used to actually just carry bottles in my medical belt. It was more efficient than taking the whole jug. :)

Eagle0600:
> From experience, I used to actually just carry bottles in my medical belt. It was more efficient than taking the whole jug. :)

This is a very good idea for paramedics, but medical doctors shouldn't have to.

XavierSomething:
> I'd say that while this limitation makes malicious use hard, it also makes its core functionality very limited. I'd say a 10u reservoir (with the same usage delay) would be better, as it would allow doctors to load and inject a typical dosage without needing to inject half, run back to the medicine cabinet, and then run back to the patient to inject the other half. You don't want the doctor to have to bring the jug with them (this causes other issues).

Thinking about it, 10u reservoir would be good, since you cant inject all 10u at once, you still are limited by the use delay of 5u at a time, just saves time on the return trip for doctors, I agree with this change!

Fansana:
Could you add a .5 do-after to the use? that way it's harder to abuse as an antag. As for the reservoir I think it's fine. As for the research place is as a t2 epi research.

Eagle0600:
It already has a slow use delay and a reduced capacity that both effect that. A 2.5 delay makes it already pretty much possible to use effectively as am antag and, frankly, makes it difficult for it to compete with regular syringes for regular doctors. Adding a do-after on top of that is excessive and would make it pretty much pointless when it's meant to be better than a syringe.

FoxxoTrystan (approved):

Memeji (approved):

PR #220, opened 2024-10-01
Fix Disposal Pipes Ejection

Fansana:
# Description

This PR fixes an issue with disposal pipes that lead into space. The ejection would use the transform of the map instead of the one from the grid and throw into the wrong direction. 
Another issue this fixes is with disposal bends which would take direction from a previous pipe instead of the current one.

# Changelog

:cl:
- fix: Fix the disposal pipe ejection to space and after pipe bends.


Fansana:
This might also be a candidate for EE.

Memeji (approved):

Suraru:
PR'd to upstream

PR #233, opened 2024-10-06
Increase Character Slots

Memeji:
# Description

Simple PR to increase the number of character slots from 10 to 50.

---

# Changelog

:cl:
- tweak: Increased Number of Character Slots to 50.


FoxxoTrystan (approved):
I would have Prefer 20 to 30.

flyingkarii:
At the end of the day, a character slot limit doesn't change much. It simply lets people who like having different characters do so at their whims.

Memeji:
> I would have Prefer 20 to 30.

I was told to set it to 50 since it won't be much of a problem.

Fansana:
Changed like those don't need a PR, they an be done with a config edit.

PR #244, opened 2024-10-08
Shadowkin

FoxxoTrystan:
# Description

Port PR: https://github.com/Simple-Station/Einstein-Engines/pull/960
This is an early merge and only admins will have access to the specie for the moment, please report all bugs found to @FoxxoTrystan .

Another PR will be made when Shadowkin is released for all players.

---

# Changelog

:cl:
- add: Artificial Bluespace Crystals!

PR #261, opened 2024-10-15
Fix ethereal.swsl not compiling in Compatibility Mode

pissdemon:
# Description

This should fix a regression introduced by #244 that prevents Floofstation from starting when Compatibility Mode is on in the launcher. The launch fail is due to type issues in the shader code. The PR does not change the meaning of the code, just fixes the type issues.

```
ERROR: 0:263: '*' : wrong operand types - no operation '*' exists that takes a left-hand operand of type 'const int' and a right operand of type 'const float' (or there is no acceptable conversion)
ERROR: 0:263: '+' : wrong operand types - no operation '+' exists that takes a left-hand operand of type 'const float' and a right operand of type 'const int' (or there is no acceptable conversion)
ERROR: 0:266: 'assign' : cannot convert from 'const int' to 'highp float'
```

# Changelog

:cl:
- fix: Fixed regression bug that prevents Floofstation from starting when Compatibility Mode is enabled in the SS14 Launcher

### DIFF
--- a/Resources/Textures/Shaders/ethereal.swsl
+++ b/Resources/Textures/Shaders/ethereal.swsl
@@ -64,7 +64,7 @@ void fragment() {
 
     highp vec4 background = zTextureSpec(SCREEN_TEXTURE, ( FRAGCOORD.xy + vec2(w) ) * SCREEN_PIXEL_SIZE );
     highp vec3 hsvBg = rgb2hsv(background.rgb);
-    hsvBg.x *= -1;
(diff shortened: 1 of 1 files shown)


FoxxoTrystan:
Guh, Seriously?

FoxxoTrystan (approved):

FoxxoTrystan:
Note to self: add this fix to https://github.com/Simple-Station/Einstein-Engines/pull/960

PR #286, opened 2024-10-25
Shadowkin Support

FoxxoTrystan:
# Description

Add the final support required for Shadowkins, this is only the game side support, Wiki is still required to be made, This can be merged before/After wiki.

This PR add map support for shadowkins, making it by default and so no need for admin intervention.
Fixes also a bunch of tiny things for kins.

This also bring fixes regarding to Ethereal and Collision when laying down.

---

# TODO

- [x] Map Support.
- [x]  Job Slot.
- [x] Anomaly cannot be antags target.
- [x] Fix Shadowkin Portal Timeout.

---

# Changelog

:cl: FoxxoTrystan
- add: The dark has spread on all the stations (Added Shadowkin Portals)
- fix: Fix Ethereal Collision when Laying down.


Fansana:
For now can you limit the amount of anomaly jobs to two? I wanna roll it out slowly and see what people do. Having to deal with 2 is easier than 20

Fansana (approved):

PR #322, opened 2024-11-04
Cryptobiolin Uplink Injector

VividPups:
# Description

This Make it so Nukies/Syndies can buy a 20u cryptobiolin  hypospray with 20u for 2 TC.
How many times have you been fucked over because someone had mind swap, mass sleep, or some other psionic that instantly ducks you over?
(You could also use it aggressively to insulate your target)
---

image

# Changelog

:cl:
- add: New way for Syndies to deal with those pesky psionic


Memeji (changes requested):
Test Fails are valid.
"UplinkCryoinjector" needs to only be in one file as right now it sees two with the same ID.

VividPups:
> Test Fails are valid. "UplinkCryoinjector" needs to only be in one file as right now it sees two with the same ID.

oh you know what I need to # it then. See the problem is that they arent showing up in the uplink unless its in the based file

VividPups:
Request was made

VividPups:
I will be changing some things to match the OG requested in the discord

VividPups:
Changes been made

Memeji (approved):
Test fail is false.
Those issues should have been resolved.

VividPups:
Is there a reason why it is causing these false fails?

Memeji:
> Is there a reason why it is causing these false fails?

Some prototypes causing issues that were recently fixed.
For some reason not caught by the system until NOW.

Resources/Locale/en-US/store/uplink-catalog.ftl
```diff
@@ -163,6 +163,10 @@ uplink-syndicate-martyr-module-desc = Turn your emagged borg friend into a walki
 uplink-singularity-beacon-name = Singularity Beacon
 uplink-singularity-beacon-desc = A device that attracts singularities. Has to be anchored and powered. Causes singularities to grow when consumed.
 
+# Floof station
+uplink-Cryptobiolininjector-name = Cryptobiolin injector
+uplink-Cryptobiolininjector-desc = Made by SESWC. This item is prefilled with 20u of Cryptobiolin to help you get rid of those pesky psionic weirdo's.
+
```
FoxxoTrystan, line 169: Move this to Floof FTL Folders.

FoxxoTrystan (changes requested):

FoxxoTrystan (approved):

PR #327, opened 2024-11-04
Psionic Hypnosis

FoxxoTrystan:
# Description

Require: https://github.com/Fansana/floofstation1/pull/373

Enable back the Psionic Hypnosis for players as a trait costing 2 and 0 Power Slot, this now fully use the Consent System and is provided with a exit verb and warning messages.

This require the rework/rewrite of Consent system by @Memeji 

---

# TODO

- [x] Psionic Hypnosis
- [x] Consent System

---

# Changelog

:cl:
- add: Added Psionic Hypnosis.


FoxxoTrystan:
> Psionic hypnosis is probably just- not a good idea. At best an ability like this could be used to like, make people say stuff you queue for it, but as a power, even if it only affected someone for a short duration or in roleplay settings, it has far too much potential to cause unintended self antag situations.

This has already been adressed and fixed.

fenndragon (changes requested):
I feel perhaps the consent system setup should be in a separate PR, and while the antag side of this has been addressed, one of the issues that came up was actually the opposite. A captain using it on a person to confirm whether or not they were an antag. (Sort of like a truth serum) 

From what I can tell, this ability is just too powerful if it doesn't have some major stipulations. 

Have you considered putting a timer on it like with mindswap? 

Take hand cuffs for instance as an example of someone whom is in a helpless position, it takes constant attention and intent to maintain the capture of the person, they can try to break free at any point and have to be moved/dragged along to keep them from escaping.

FoxxoTrystan:
> I feel perhaps the consent system setup should be in a separate PR, and while the antag side of this has been addressed, one of the issues that came up was actually the opposite. A captain using it on a person to confirm whether or not they were an antag. (Sort of like a truth serum)
> 
> From what I can tell, this ability is just too powerful if it doesn't have some major stipulations.
> 
> Have you considered putting a timer on it like with mindswap?
> 
> Take hand cuffs for instance as an example of someone whom is in a helpless position, it takes constant attention and intent to maintain the capture of the person, they can try to break free at any point and have to be moved/dragged along to keep them from escaping.

All of those has been adressed arleady and fixed.
Read the PR and this does not change the consent system, check the other PR.
There is warning and rules given when the power is used, so all of the current statement are already adressed.

fenndragon (changes requested):
I did read the PR, and the way it's coded. It addressed the antag side of things but the issue given in the example is the opposite. 

What prevents people from using this to get information while playing security if the person's consent preferences say yes to this power?

FoxxoTrystan:
Test fails has been fixed, this also fix a future test.

PR #337, opened 2024-11-08
N.T.N.F. Scarab

JasmineRickards:
A more compact replacement for the existing NTNF Rook, smaller, more nimble, and most importantly, more docks! Very important for a cruiser class carrier.


Fansana:
why did you close it?

PR #343, opened 2024-11-11
Bionic Changes and Fixes

Memeji:
# Description

Makes it so the bionic arm can't open powered doors.
Gladiators can't have bionics.

---

# Changelog

:cl:
- fix: Gladiators can no longer have bionics.
- tweak: Bionic arm cannot opened powered doors. Cost reduced to 6 instead of 8.

### DIFF
--- a/Resources/Prototypes/Traits/physical.yml
+++ b/Resources/Prototypes/Traits/physical.yml
@@ -365,6 +365,7 @@
       inverted: true
       jobs:
         - Prisoner # Bionics should be "Confiscated" from long term prisoners.
+        - Gladiator # Floof
     - !type:CharacterSpeciesRequirement
       species:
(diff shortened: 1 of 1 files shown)


fenndragon:
Should lower the cost of the arm if it is getting relegated to crowbar instead of jaws of life.

Memeji:
> Should lower the cost of the arm if it is getting relegated to crowbar instead of jaws of life.

Already done

Fansana:
I dunno 6 points for what is essentially a crowbar seems a tad much.

Fansana:
I think 3 or 2 points would be better

fenndragon:
I suggested 5 but. 3 sounds good
2 might be undercutting some other qol traits.

Fansana (approved):

PR #348, opened 2024-11-13
Fix Block Game

ghost:
:cl:
- fix: Fixed block game UI!

### DIFF
--- a/Content.Server/Arcade/BlockGame/BlockGame.cs
+++ b/Content.Server/Arcade/BlockGame/BlockGame.cs
@@ -2,7 +2,6 @@ using Content.Shared.Arcade;
 using Robust.Server.GameObjects;
 using Robust.Shared.Random;
 using System.Linq;
-using Content.Shared.Mood;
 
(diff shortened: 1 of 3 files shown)


Content.Server/Arcade/BlockGame/BlockGame.cs
```diff
@@ -83,8 +82,6 @@ private void InvokeGameover()
         {
             _highScorePlacement = _arcadeSystem.RegisterHighScore(meta.EntityName, Points);
             SendHighscoreUpdate();
-            var ev = new MoodEffectEvent("ArcadePlay");
-            _entityManager.EventBus.RaiseLocalEvent(meta.Owner, ev);
```
Mnemotechnician, line 87: why does this remove the positive moodlet from playing arcades?

Fansana: probably because I'm completely sleep deprived and just copied the ss14 changes.

Fansana:
there we go this should be the fix

Memeji (approved):

PR #370, opened 2024-11-20
Fix PirateRadio Heisentest

sleepyyapril:
# Description

You are now free from the burden of this infamous heisentest.
Also I can PR the engine updates with no issue, if you'd like? Not sure if that's something y'all would want, but it wouldn't be hard. The engine update includes the arrivals visuals PR, so, might be good to get that lmao.


FoxxoTrystan (approved):

FoxxoTrystan:
Engines updates would be nice, after the selective merge.

Memeji (approved):

PR #373, opened 2024-11-22
Consent UI Generated

FoxxoTrystan:
# Description

This PR does not include a changlog as its does not change anything gameplay wise.

For devs, this makes so consent options in the UI are auto-generated by the prototype making it more easy to handle consent options as its generated correctly, also untangle some tiny messes and make the UI way more pretty.

---

<p>

image
image

</p>
</details>


fenndragon (changes requested):
A change log should still be necessary for visual/backend UI changes. Those aren't just for the players.

Memeji:
> A change log should still be necessary for visual/backend UI changes. Those aren't just for the players.

Changelogs are specifically for players.

FoxxoTrystan:
No changelogs are needed as this changes nothing for the players directly.

sleepyyapril:
Is the idea for this being able to do something like _consentManager.GetConsentStatus("key"), similar to ccvars?

FoxxoTrystan:
> Is the idea for this being able to do something like _consentManager.GetConsentStatus("key"), similar to ccvars?

This only changes Consent UI Aspect, not serverside, currently you need to do this:

_consent.HasConsent(uid, "key")

Memeji (approved):

sleepyyapril:
> > Is the idea for this being able to do something like _consentManager.GetConsentStatus("key"), similar to ccvars?
> 
> This only changes Consent UI Aspect, not serverside, currently you need to do this:
> 
> _consent.HasConsent(uid, "key")

Right, but the goal is similar, which is what I meant. Amazing work.

PR #383, opened 2024-11-27
Mining Drill

fenndragon:
# Description

This pull request fixes the mining drill to have it's stats back to expected values and performance. 

---

# Changelog

:cl:
- fix: Fixed mining drill

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Melee/pickaxe.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Melee/pickaxe.yml
@@ -64,9 +64,9 @@
     range: 1.5
     damage:
       types:
-        Blunt: 6
-        Slash: 3
-        Structural: 12
+        Blunt: 2
+        Slash: 1
+        Structural: 15
     bluntStaminaDamageFactor: 4.0
     heavyRateModifier: 1
     heavyRangeModifier: 2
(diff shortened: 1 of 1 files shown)


fenndragon:
> Reduce the blunt and slash damage to match the new attack rate. Something like around ~2 or ~1 for them.

ah- I had not noticed the values for the damage on our drill were not the same as frontier, I will gladly change it to match properly

fenndragon:
> Reduce the blunt and slash damage to match the new attack rate. Something like around ~2 or ~1 for them.

alright, I adjusted the damages to match the drill previously, on frontier it deals 3 "brute" so I put 2 blunt, 1 slash, and it deals 15 structural

Memeji (changes requested):
I would reduce the blunt stamina damage factor to 1 and make the heavy stamina cost somewhere around ~1 or ~2.

Or make the heavy cost 0 but reduce the heavy range modifier to 1

fenndragon:
> I would reduce the blunt stamina damage factor to 1 and make the heavy stamina cost somewhere around ~1 or ~2.
> 
> Or make the heavy cost 0 but reduce the heavy range modifier to 1

these settings are to remove the stamina use from the equation entirely, it's a proof of concept in the form of a "weapon" that's not used against people to see if the change is a welcome one. 

if it goes well I intend to look into making edits to other melee weapons to fix the current system which makes you fall on your face after a few swings.

Memeji:
Will be testing this on my local instance to see the details and how it works out.

Memeji (approved):
Went over these changes, worth a try on live to see how it works out.

FoxxoTrystan (approved):

PR #384, opened 2024-11-27
Milking Machine

Lyroth001:
# Description

Adds a milking machine bcs hot :p

Opening a draft to get feedback on like the code implementation and stuff
Current changes - solutionregeneration can now check if its on an item that needs to be equipped, and if so only generate solution when equipped, clothing now has a bool for if its equipped. Slightly hackey probably but thats why its a draft.
Currently adds a proof of concept version

---

# TODO

- [ ] Get Sprites (I am not a spriter sadly so would appreciate help with this )
- [ ] Add to loot table or something
- [ ] Restrict to characters with penises maybe?
- [x] Add the basic milking machine code

---

I will add this later
<p>

</p>
</details>

---

# Changelog

:cl:
- add: Milking Machines can now be found on the station

### DIFF
--- a/Content.Shared/Clothing/Components/ClothingComponent.cs
+++ b/Content.Shared/Clothing/Components/ClothingComponent.cs
@@ -66,6 +66,9 @@ public sealed partial class ClothingComponent : Component
     [DataField("unisexMask")]
     public ClothingMask UnisexMask = ClothingMask.UniformFull;
 
+    [DataField("isEquipped")]
+    public bool IsEquipped;
+
     /// <summary>
     /// Name of the inventory slot the clothing is in.
     /// </summary>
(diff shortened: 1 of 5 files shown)


Content.Shared/Clothing/Components/ClothingComponent.cs
```diff
@@ -66,6 +66,9 @@ public sealed partial class ClothingComponent : Component
     [DataField("unisexMask")]
     public ClothingMask UnisexMask = ClothingMask.UniformFull;
 
+    [DataField("isEquipped")]
+    public bool IsEquipped;
+
```
Mnemotechnician, line 71: This should not be here. Check `InSlot != null` instead. Introducing tautologies like this will lead to new edge cases in the future that this PR cannot account for.

Lyroth001: Gotcha, couldnt find something like that when I looked but its good to know its there!

Fansana:
closing due to inactivity

PR #389, opened 2024-12-02
Nudity Permit

FoxxoTrystan:
# Description

Replace the loin cloth with the nudity permit, request made by @Memeji .

---

# Changelog

:cl:
- tweak: Replace the loin cloth with the nudity permit.


Mnemotechnician:
Why not keep both?

PR #406, opened 2024-12-15
Adding Body Mass to Health Analyzer UI

LucasTheDrgn:
# Description

On most servers, the body mass of patients is something that can be generally inferred. Crew members won't have significantly adverse affects to normal doses of lep or ins, but station pets will. Except on floof, half the crew members are physically smaller than station pets, and sometimes start to approach the station pets in terms of body mass, which means lep and ins can actually start to overheat or freeze them, even in normal doses. Now, doctors can assess the patients' body mass and identify if they are at risk of adverse affects due to the temperature fluctuations applied by these meds. One more tool in the medbay toolbelt. One oddity I've observed is that the mass doesn't line up with the mass given in character creation, but seeing as my code is pulling directly from the FixturesMass variable, and does properly scale with selected height/width/presence of lightweight trait, it's at least consistent with itself, and only off by a factor of 0.1, e.g. my character that shows up in menu as 100kg is showing 91kg int he analyzer, but if i give them the Lightweight trait they show up as 60kg, as one would expect.

---

<p>

image
image
image

</p>
</details>

---

# Changelog

:cl:
- add: Health Analyzers will now display the patient's body mass, and thus their susceptibility to temperature changing reagents. And throwability.


Fansana:
great addition, gonna merge this right after the upstream one.

Fansana (approved):

PR #429, opened 2024-12-23
Add Codeword Highlighting (#30092)

ghost:
:cl:
- add: Added code word highlight

### DIFF
--- a/Content.Shared/Roles/RoleCodeword/RoleCodewordComponent.cs
+++ b/Content.Shared/Roles/RoleCodeword/RoleCodewordComponent.cs
@@ -1,0 +1,37 @@
+using Robust.Shared.GameStates;
+using Robust.Shared.Prototypes;
+using Robust.Shared.Serialization;
+
+namespace Content.Shared.Roles.RoleCodeword;
+
+/// <summary>
+/// Used to display and highlight codewords in chat messages on the client.
+/// </summary>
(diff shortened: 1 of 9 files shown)


Content.Shared/Roles/RoleCodeword/RoleCodewordComponent.cs
```diff
@@ -0,0 +1,37 @@
...
+public sealed partial class RoleCodewordComponent : Component
+{
+    /// <summary>
+    /// Contains the codewords tied to a role.
+    /// Key string should be unique for the role.
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public Dictionary<string, CodewordsData> RoleCodewords = new();
+
+    public override bool SessionSpecific => true;
```
Mnemotechnician, line 20: Would be better to have `SendOnlyToOwner => true` instead of reinventing the wheel here and in OnCodewordCompGetStateAttempt, but if it works, it works.

Mnemotechnician (approved):

Content.Shared/Roles/RoleCodeword/RoleCodewordComponent.cs
Fansana: I mean we can improve on it but it's straight from wizden so I don't really wanna mess with it.

PR #448, opened 2025-01-01
Frontier Clothes

ghost:
:cl:
- add: Added lots of punk clothing from frontier
- add: Added chaplain clothing: sash Crucifix


Resources/Prototypes/Entities/Objects/Specific/Medical/morgue.yml
```diff
@@ -116,6 +116,7 @@
   - type: Tag
     tags:
     - Trash
+    - Ash # Frontier
```
Mnemotechnician, line 119: I'm not sure we have the ash tag on floof/ee. It's why I added the Burnt tag to the ash entity in my bugfix PR, and it may be the reason tests are failing here.

Fansana: image
ash is there

PR #454, opened 2025-01-05
Pka Salvage Borg Changes 

VividPups:
commit

# Description

Salvage borg had no melee weapon and couldn't really used the pka due to the weapon angel angel-wielding requirement 
this PR gives them a Survival knife (as told by floor)

---

# Changelog

:cl:
- tweak: SESWC tinkered with the salvage borg PKA module and taped a knife to its other arm.

### DIFF
--- a/Resources/Prototypes/Floof/Entities/Objects/Weapons/borgmodules.yml
+++ b/Resources/Prototypes/Floof/Entities/Objects/Weapons/borgmodules.yml
@@ -1,38 +1,17 @@
 - type: entity
-  name: Borg kinetic Accelerator
-  parent: BaseWeaponBatterySmall
   id: WeaponProtoKineticAcceleratorborg
-  description: borg salvage gun
+  parent: WeaponProtoKineticAccelerator
+  name: proto-kinetic accelerator borg
+  noSpawn: true
   components:
-  - type: Sprite
-    sprite: Objects/Weapons/Guns/Basic/kinetic_accelerator.rsi
-    layers:
-    - state: icon
-      map: ["enum.GunVisualLayers.Base"]
-  - type: Clothing
-    sprite: Objects/Weapons/Guns/Basic/kinetic_accelerator.rsi
   - type: Gun
     fireRate: 0.5
-    availableModes:
-    - SemiAuto
+    selectedMode: SemiAuto
     minAngle: 1
     maxAngle: 3
+    availableModes:
+    - SemiAuto
     soundGunshot:
       path: /Audio/Weapons/Guns/Gunshots/kinetic_accel.ogg
-    fireOnDropChance: 0.5
-  - type: ProjectileBatteryAmmoProvider
-    proto: BulletKinetic
-    fireCost: 100
-  - type: BatterySelfRecharger
-    autoRecharge: true
-    autoRechargeRate: 40
-  - type: Battery
-    maxCharge: 1000
-    startingCharge: 1000
-  - type: Appearance
-  - type: Tag
-    tags:
-    - Sidearm
-  - type: StaticPrice
-    price: 750
+    fireOnDropChance: 1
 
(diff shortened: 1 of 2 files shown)


fenndragon (changes requested):
alright so the sprites listed aren't actually necessary! you're overlapping some of the stuff from the parent and not adding states. one moment.

FoxxoTrystan (approved):

Resources/Prototypes/Floof/Entities/Objects/Weapons/borgmodules.yml
```diff
@@ -0,0 +1,20 @@
+- type: entity
+  id: WeaponProtoKineticAcceleratorborg
+  parent: WeaponProtoKineticAcceleratorBase
```
Mnemotechnician, line 3: This should probably inherit from WeaponProtoKineticAccelerator instead. The tests are failing because the -base proto doesn't have an item sprite. <h6>also, please, capitalize the word "borg" in the ID...</h6>

Resources/Prototypes/Floof/Entities/Objects/Weapons/borgmodules.yml
FoxxoTrystan, line 1: A Sprite is required for this entity.

Test fails valid.

FoxxoTrystan (changes requested):

Mnemotechnician (changes requested):

Resources/Prototypes/Floof/Entities/Objects/Weapons/borgmodules.yml
fenndragon: he's getting a new version of the weapon uploaded now. one moment

fenndragon (approved):
looks good to me

Resources/Prototypes/Floof/Entities/Objects/Weapons/borgmodules.yml
```diff
@@ -0,0 +1,38 @@
+- type: entity
+  name: Borg kinetic Accelerator
+  parent: BaseWeaponBatterySmall
```
Mnemotechnician, line 3: Don't do this. This is a lot of code duplication and will become a large maintenance burden in the future. You should inherit from the proto-kinetic accelerator entity and only change the components you want changed.

fenndragon: I was told to handle setup for this, I don't see how or where there's duplication. if you would rather rewrite it yourself feel free.

Mnemotechnician: The original was almost good enough, with the exception that it inherited from the wrong entity (the one it inherited from was abstract), which caused the test fails.

VividPups:
There is one more thing I need to work one and that is to make the magnet for the ore bag work on the borg

fenndragon:
> There is one more thing I need to work one and that is to make the magnet for the ore bag work on the borg

I am uncertain without looking if it's possible to make a bag magnet function without being worn... I will look

VividPups:
> > There is one more thing I need to work one and that is to make the magnet for the ore bag work on the borg
> 
> I am uncertain without looking if it's possible to make a bag magnet function without being worn... I will look

well so the Inderustal Process has a one so it might be as simple as making a new module or just adding something to the existing

VividPups:
> > There is one more thing I need to work one and that is to make the magnet for the ore bag work on the borg
> 
> I am uncertain without looking if it's possible to make a bag magnet function without being worn... I will look

Honestly I think there is a way there is some things I found in the VV however, I really don't want to go diving in it and would like this to be in with the update coming so I think I'm good with it now

Mnemotechnician:
Looks alright now.

PR #471, opened 2025-01-09
Energy Shotgun Fix

fenndragon:
# Description

Fixed Energy Shotgun damage values and sprites 

---

# Changelog

:cl:
- fix: Fixed Energy Shotgun


Resources/Prototypes/Entities/Objects/Weapons/Guns/Projectiles/projectiles.yml
```diff
@@ -1042,9 +1042,21 @@
...
+  name: energy bolt
+  id: BulletEnergyGun
+  parent: BulletEnergyGunLaser
+  noSpawn: true
+  components:
+  - type: Projectile
+    impactEffect: BulletImpactEffectRedDisabler
+    damage:
+      types:
+        Heat: 5 # Slightly more damage than the 17heat from the Captain's Hitscan lasgun
```
Mnemotechnician, line 1055:
```suggestion
        Heat: 5
```

Resources/Prototypes/Floof/Entities/Objects/Weapons/guns.yml
```diff
@@ -138,6 +132,40 @@
...
+    fireCost: 150
+  - type: Item
+    size: Large
+    shape:
+    - 0,0,3,1
+    sprite: Objects/Weapons/Guns/Battery/inhands_64x.rsi
+    heldPrefix: energy
   - type: Battery
     maxCharge: 1200
     startingCharge: 1200
```
Mnemotechnician, line 171: So, two issues. First, this is an exact copy of WeaponEnergyShotgun, except it lacks the BatterySelfRecharger component. Instead of introducing so much code duplication (and room for errors), you should introduce a base type contains all components from WeaponBorgEnergyShotgun, and make both WeaponBorgEnergyShotgun and WeaponEnergyShotgun inherit from it (the former would also need to inherit from BaseGunWieldable and add the BatterySelfRecharger comp)

Second issue is this is a BORG WEAPON, and you provide no way to recharge it. This effectively makes the weapon single-use, because a borg cannot just unequip one of their tools and plug it into a recharger.

Resources/Prototypes/Floof/Entities/Objects/Weapons/guns.yml
```diff
@@ -138,6 +132,40 @@
   - type: StealTarget
     stealGroup: WeaponEnergyShotgun
   - type: GunRequiresWield #remove when inaccuracy on spreads is fixed
+  - type: Battery
+    maxCharge: 1200
+    startingCharge: 1200
+  - type: BatterySelfRecharger
+    autoRecharge: true
+    autoRechargeRate: 24
```
Mnemotechnician, line 140: This is higher than the recharge rate of a normal recharger (24 watts vs 20 when plugged into a regular weapon recharger).

Mnemotechnician (changes requested):

Resources/Prototypes/Floof/Entities/Objects/Weapons/guns.yml
fenndragon: battery self recharger is literally under your comment. 

The creation of another gun with similar properties instead of parenting both off of a single weapon was due to the fact I would have to work backwards due to the way wielding component works. the borg shotgun has fewer things than the shotgun, but parenting the shotgun off of the borg shotgun feels backwards. I would rather just opt for making the most basic entity for a borg shotgun instead of doing that parenting nightmare (and introducing a lot of room for errors as parents aren't immediately apparent to me while coding what applies where.

Resources/Prototypes/Floof/Entities/Objects/Weapons/guns.yml
fenndragon: this is the rate of the energy shotgun on wizden, I made no edits or modifications to its recharge rate.

Mnemotechnician: Then that one's fine for now, we'll see if it causes issues later

Resources/Prototypes/Floof/Entities/Objects/Weapons/guns.yml
FoxxoTrystan: battery self recharger is not in WeaponBorgEnergyShotgun, best still is to indeed parent it indeed.

Mnemotechnician: Yeah, there's no BatterySelfRecharger in this prototype. This is a mistake that can easily be made (and has been made many many many times before) when you copy-paste prototypes instead of using entity parenting.

FoxxoTrystan: There is an BatterySelfRecharger just at the bottom, guh.

Anywho i made the changes myself, double check it in case i fucked up

FoxxoTrystan (approved):

Mnemotechnician (approved):
Looks good now. I'm hoping that someone will test it before it gets merged, though.

FoxxoTrystan:
> Looks good now. I'm hoping that someone will test it before it gets merged, though.

Tested it myself, wild and borgo one.

PR #474, opened 2025-01-10
Increase Artifact Analyzer's Point-to-Glimmer Ratio

Mnemotechnician:
# Description
This makes ExtractRatio property of artifact analyzers (the amount of points needed to raise 1 milipsi of glimmer) a DataField, and changes that property of MachineArtifactAnalyzer to be 2.5 times higher than the default (1000 instead of 400 points per 1 milipsi).

This should make xenoarch more of a viable research method, because as of now anomalies are the go-to due to their incredibly low glimmer generation compared to that of xenoarch.

### **This will require testing.**

# Changelog
:cl:
- tweak: Artifact extractions should now generate 2.5 times less glimmer than before.


FoxxoTrystan (approved):

Memeji (approved):
Let's give it a shot!

PR #476, opened 2025-01-11
Drunkard Accent

mazziechai:
# Description

Adds a "drunkard accent", a trait that makes a character's speech slurred like they're drunk.

There are a few alcoholic characters on Floof that would benefit from being "constantly drunk" without the actual status effect. This could be a good effect for other characters that have a speech impediment.

## Technical notes

This uses a modified version of `SlurredSystem`'s `Accentuate` for the trait with hardcoded values and removing the burping and "huh" additions. The functionality *could* be extracted into a static method, however it would require changing code in non-Floof code when this works perfectly fine; reducing changes to the codebase is preferable.

If the character is drunk past a certain point, the `SlurredSpeech` effect will take precedence over `DrunkardAccent`.

# Changelog

:cl:
- add: Added a "drunkard accent", a trait that makes a character's speech slurred like they're drunk.

### DIFF
--- a/Content.Server/FloofStation/Speech/Components/DrunkardAccentComponent.cs
+++ b/Content.Server/FloofStation/Speech/Components/DrunkardAccentComponent.cs
@@ -1,0 +1,4 @@
+﻿namespace Content.Server.FloofStation.Speech.Components;
+
+[RegisterComponent]
+public sealed partial class DrunkardAccentComponent : Component;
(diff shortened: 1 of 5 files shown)


Mnemotechnician:
This is... suboptimal. You could just give the entity the slurred speech status effect and make this system renew it whenever it runs below a certain threshold (say, 5 minutes of remaining effect time)

mazziechai:
> This is... suboptimal. You could just give the entity the slurred speech status effect and make this system renew it whenever it runs below a certain threshold (say, 5 minutes of remaining effect time).

The unfortunate thing is that this would require either giving a constant drunk level or refactoring the original SlurredSpeech system to not be hard coupled to DrunkSystem. SlurredSystem *always* calculates probability based on the drunk status effect. This is a drunk _accent_, not an *always drunk* trait, so I believe that this suboptimal code is actually the best for avoiding unnecessary changes.

FoxxoTrystan (approved):

PR #480, opened 2025-01-11
Make Visual Disabilities and Cybereyes Mutually Exclusive

Tirochora:
# Description

Adds incompatibility to ensure that someone cannot take negative visual traits while having cyber-eyes systems, and likewise cannot take redundant negative visual traits.

---

<p>

githubphotophobiaimage

</p>
</details>

---

# Changelog

:cl:
- fix: Made Photophobia, Blindness, Nearsightedness, and Cyber-eyes mutually exclusive.

### DIFF
--- a/Resources/Prototypes/Traits/disabilities.yml
+++ b/Resources/Prototypes/Traits/disabilities.yml
@@ -8,6 +8,12 @@
       jobs:
         - Borg
         - MedicalBorg
+    - !type:CharacterTraitRequirement
+      inverted: true
+      traits:
+        - Nearsighted
+        - Photophobia
+        - CyberEyes
   functions:
     - !type:TraitAddComponent
       components:
@@ -23,6 +29,11 @@
       jobs:
         - Borg
         - MedicalBorg
+    - !type:CharacterTraitRequirement
+      inverted: true
+      traits:
+        - Blindness
+        - CyberEyes
   functions:
     - !type:TraitAddComponent
       components:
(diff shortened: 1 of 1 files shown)


FoxxoTrystan (approved):

Mnemotechnician:
I think you may need to add the job-specific cybereyes traits to the exclusion lists. There are specialized cybereyes traits for security and I think medical jobs.

Also, I edited your changelog to be more concise — feel free to adjust it if needed.

Memeji (changes requested):
Agreed with Mnemotechnician on also adding the Sec and Med eyes to the exclusions just in case.

Though I'm unsure if there is a way to have the specialized eyes without the cyber eyes, so it might just be useful for in-case scenarios and glitches.

Tirochora:
The SecHud, Medhud, and Diagnostic Hud add-ons (as well as the Hud Suite) can only be taken if you already have the base Cyber-Eyes system, which is exclusive with visual disabilities as of this PR. As such, I don't see very much point to making those exclusive as well, though I could if you believe it might be necessary.

Memeji:
> The SecHud, Medhud, and Diagnostic Hud add-ons (as well as the Hud Suite) can only be taken if you already have the base Cyber-Eyes system, which is exclusive with visual disabilities as of this PR. As such, I don't see very much point to making those exclusive as well, though I could if you believe it might be necessary.

Yeah, not really necessary to be honest. As I can't think how someone can get the add-ons without the base.

Memeji (approved):

PR #501, opened 2025-01-22
Batonbot and Disablerbot

Timfa2112:
# Description

Adds the Batonbot and the Disablerbot. Both of these will idle around until they encounter hostile wildlife, such as angry slimes, and will then liberally apply their respective weapon until the threat is no more.

The Batonbot has a weakened version of the Baton, applying 15 less Stamins damage and only 4 burn damage
The Disablerbot uses the SMG Disabler bullet with a significant spread at a fire rate of 1
This means that both of them are less powerful than a normal baton or disabler.

They are built with:
- Proximity Sensor
- Security Helmet
- Stunbaton/Disabler
As such requiring co-operation from Security to construct.

Also included are Spawners, for if we want them on stations on round-start.

The recipes necessitated a few new tags, which have been marked with # Floofstation, along with the reason they were added to the entities.

---

<p>

https://github.com/user-attachments/assets/8e6c9861-004c-4138-abcc-f0fe203ed9d0

https://github.com/user-attachments/assets/7c7cd08d-3c8e-4896-8f25-ce8cd5a4fb0e

</p>
</details>

---

# Changelog

:cl: Timfa
- add: Add Batonbot
- add: Add Disablerbot


Mnemotechnician (commented):
Looks nice, although I'd probably prefer them to have way less hitpoints (maybe like 30 or 60 to compensate for them being free firepower)

Also, do you think you could make those emaggable?

Timfa2112:
> Looks nice, although I'd probably prefer them to have way less hitpoints (maybe like 30 or 60 to compensate for them being free firepower)
Lowering HP is no problem at all! 
> 
> Also, do you think you could make those emaggable?
If you teach me how :p I don't know how that works

Mnemotechnician (approved):

FoxxoTrystan (approved):

FoxxoTrystan:
You should make them target "Warden" users

PR #525, opened 2025-02-02
Add Floof NT Company Policy to Guidebook

Forzii:
# Description

Added Standard Operating Procedures (SOP) root entry to the guidebook, currently with only the company policy from the floof wiki, and added Floof folders to put all the files in. Company policy is default collapsed and contains the entirety of the policy in the root entry, plus broken out in children.

Of note, I experimented with setting priority 1 to ensure the rules were above the policy (when the root was Company Policy), but it seems that priority 0 is the default if none is set - so 1 put the policy at the very bottom. I don't think I like the current position below the standard "space station 14" entry, but it's probably fine? Additionally, the behavior claimed in the comments of GuideEntry.cs of children being placed in the order they are referenced seems to be incorrect, though I'm not sure how much the order matters for the child entries of the policy.

Turns out I'm not the only one who noticed the ordering issues, someone is working on fixing it! https://github.com/space-wizards/space-station-14/pull/34960

This is my first ever PR, so please give feedback!

---

# TODO

- [x] Add company policy guidebook entries and supporting changes
- [x] Add SOP root for guidebook and generalize files for future planned additions
- [x] Get feedback on location in guidebook and the files themselves (I ended up moving everything into Floof folders anyway)
- [x] Get feedback on if the entirety of the policy should be in the Company Policy root entry or not (no one seems to care)
- [x] Get feedback on viability of two sources of truth (wiki and guidebook), mitigations, sustainability, etc

---

<p>

image
image
image

</p>
</details>

---

# Changelog

:cl:
- add: Added NT Company Policy to guidebook


Forzii:
So, probably relevant... I want to start adding more of the stuff from the wiki to the guidebook so players don't HAVE to refer to the wiki to get certain information. With that in mind, having "NanoTrasen Company Policy" as a root entry in the guidebook might be somewhat short-sighted. Maybe we should instead nest the company policy under a root entry called "Standard Operating Procedures" like the wiki? 

I could also move companypolicy.yml to prototypes->Floof ->Guidebook and the content to ServerInfo->Floof->Guidebook->SOPs or something like that... again, I'd appreciate feedback! I don't really know when we should/shouldn't use Floof folders. Or if you all don't like this at all, lemme know.

Mnemotechnician:
This introduces an issue as players will now have two sources of trust: the wiki, which can change any second without any prior notice, and the guidebook, which can only change with a release (or an emergency patch), which typically happens once in a week. I think we need the opinion of wiki/SOP maintainers on this one.

Forzii:
> This introduces an issue as players will now have two sources of trust: the wiki, which can change any second without any prior notice, and the guidebook, which can only change with a release (or an emergency patch), which typically happens once in a week. I think we need the opinion of wiki/SOP maintainers on this one.

Hmm, that's a fair point. It's actually part of why I started with the Company Policy, since it has been a relatively stable document. We could partially mitigate it by assigning the source of truth to the wiki and stating such in the root SOP entry (along with a link to the wiki SOP list page). Could also put a last-modified date at the start of each entry, though that then relies on coders/maintainers updating those dates when changes are made. 

A better mitigation would be a workflow or something that automatically pulls and re-formats the pages for the in-game guidebook on a regular basis (weekly?) but I don't foresee anyone doing this (myself included).

Yeah, lets see what the wiki admins think.

Forzii:
Alright, added some red text to try to make it clear that the wiki is still the source of truth and included last-updated dates. Both Seb and Tech seem good with it, feel free to confer with whoever you need to (or let me know who else I should ask for feedback).

Memeji (approved):

PR #528, opened 2025-02-04
Biosynthetic Left Leg Fix

LucasTheDrgn:
# Description

Copy/paste error causing bio-synthetic left lefts to expect right feet as children.

---

# Changelog

:cl:
- fix: Bio-synthetic left legs now properly expect and allow a left foot to be attached.


FoxxoTrystan (approved):

Mnemotechnician (approved):
Wonder if any other body parts are also affected...

PR #529, opened 2025-02-04
ERT/Testing Surgical Tool Crate

LucasTheDrgn:
# Description

As I was testing all of these surgery bugs I often found myself spawning in the same items over and over again every time I needed to reload the instance. I figured not only would it be easier on myself to make a crate prototype, but it could also serve as an ERT Medical option. Includes:
- a souped up medical multitool (because ERT get the fancy tools)
- sterile mask and nitrile glove boxes
- a general anesthetic tank
- a medical breath mask, 
- flatpacks for a medical biofabricator and an operating table
- a multitool to unpack the flatpacks
- 300 biomass to print up to 30 body parts

---

# Changelog

:cl:
ADMIN:
- add: Added an ERT surgical tools crate, which includes new items such as an admeme medical multitool and flatpacks for the med biofab and operating tables.

### DIFF
--- a/Resources/Prototypes/Floof/Entities/Objects/Specific/Medical/surgery.yml
+++ b/Resources/Prototypes/Floof/Entities/Objects/Specific/Medical/surgery.yml
@@ -1,8 +1,8 @@
 - type: entity
-  name: ERT medical multitool
-  id: ERTOmnimedTool
+  name: bluespace medical multitool
+  id: BluespaceOmnimedTool
   parent: BaseToolSurgery
-  suffix: Admeme
+  suffix: debug, DO NOT MAP
   components:
   - type: Sprite
     sprite: Objects/Specific/Medical/Surgery/omnimed.rsi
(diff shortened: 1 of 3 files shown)


FoxxoTrystan (approved):

Resources/Prototypes/Floof/Entities/Objects/Specific/Medical/surgery.yml
```diff
@@ -0,0 +1,34 @@
+- type: entity
+  name: ERT medical multitool
+  id: ERTOmnimedTool
+  parent: BaseToolSurgery
+  suffix: Admeme
```
Mnemotechnician, line 5:
```suggestion
  suffix: debug, DO NOT MAP
```
This ^ and I think it'd be better to rename it to something like "bluespace omnitool" or something along those lines.

LucasTheDrgn: Fair! I used that suffix because that's what the admin-only experimental RCD uses, but this also makes sense. I'll make those changes in a minute.

Mnemotechnician (approved):

Memeji (approved):

PR #530, opened 2025-02-04
Increase CO2 sensor danger threshold to 1%

Eagle0600:
# Description

Increases the default CO2 danger threshold on air sensor components to 1%. The warning threshold is changed to 0.5% (automatically calculated as half of the danger threshold).

This will reduce the currently-common occurance of nuisance alarms, which currently distract atmospherics attention and train people to ignore the air alarms, degrading their purpose. This is an especially large problem with the addition of the atmospheric alert computer.

I have extensively tested the danger of CO2, and determined 1.2% to be the point at which it starts to become potentially dangerous. The danger level of 1% stands off slightly short of this and ensures that firelocks will come down in the event of any dangerous concentration of CO2. The air alarm warnings will also trigger significantly in advance of this (at 0.5%).

In addition, the occurrence of actually dangerous CO2 levels is particularly rare, further reducing the risk that this change will result in uncontrolled CO2 accidents.

Hopefully, this will make it easier for atmospherics engineers to diagnose and fix actual atmospheric issues.

---

<h4>Testing the effects of CO2 exposure</h4>
<p>

Testing the effects of CO2 exposure

</p>
<h4>The new default air sensor settings</h4>
<p>

The new default air sensor settings

</p>
</details>

---

# Changelog

:cl:
- tweak: Increased CO2 alarm thresholds.

### DIFF
--- a/Resources/Prototypes/Atmospherics/thresholds.yml
+++ b/Resources/Prototypes/Atmospherics/thresholds.yml
@@ -36,7 +36,7 @@
 - type: alarmThreshold
   id: stationCO2
   upperBound: !type:AlarmThresholdSetting
-    threshold: 0.0025
+    threshold: 0.01
   upperWarnAround: !type:AlarmThresholdSetting
     threshold: 0.5
 


Eagle0600:
I would also like to change the default station air mix. Currently, it sits at 21%, and results in harpies gasping for air frequently. On at least one occasion, I had to inject dexalin into an SSD harpy patient and have an engineer come and scrub nitrogen from medbay.

The atmospherics distribution mixers on our maps are set to between 21% and 22% air mix, depending on the map. These would need to be changed on individual maps, and I do not want to interfere with existing mapping projects. However, I feel like they should all be changed in line with a unified Floof direction.

However, the distribution mixers have a lesser impact than default station air mix, which is set to 21% oxygen, as defined in two lines of code in Atmospherics.cs. I could (and have, locally) easily change this mix to 23%, which is my preference without interfering with any mapping efforts; but I believe this change would benefit from consensus with direction. Would direction prefer a 22% mix or a 23% mix? I firmly believe an increase of some sort is warranted, and that individual changes to atmospherics distribution mixers should follow in line.

Mnemotechnician:
Increasing oxygen levels won't stop harpies from suffocating. Their lungs are sensitive to carbon dioxide, not the lack of oxygen.

Eagle0600:
> Increasing oxygen levels won't stop harpies from suffocating. Their lungs are sensitive to carbon dioxide, not the lack of oxygen.

I've just tested and re-tested, and harpies don't start gasping until the same CO2 point as everyone else.

Eagle0600:
I've had some more time to think, and I've come to the conclusion that no-one, including me, has performed thorough enough testing on the phenomenon. I was able to put together an easy enough two-gas mixes array for testing, but I only tested one harpy and moved them between cages. Plus, the gas mix changed as the harpy breathed, making the test hard to control.

I think I'm going to have to create a custom testing element that holds the gas at a cell in a specific state, like some kind of custom vent element, and perform more long-term testing with humans and harpies in the same cells.

Mnemotechnician (approved):
Nevertheless, the change *should* be fine.

Eagle0600:
I suppose it's easy enough to fix if it turns out to be a mistake. I'm still going to launch into testing. I'm going to base the testing element on a gas miner modified to clear the air before adding to it.

PR #539, opened 2025-02-07
Quadborgs 3

fenndragon:
# Description

adds 2 new Quadborgs and the Apex Admeme
---

# Changelog

:cl:
- add: Added new Quad borgs
- fix: Fixed the other quad sprites


fenndragon:
um... what do those test fails even mean

Resources/Prototypes/Floof/Entities/Mobs/Cyborgs/quadborg.yml
```diff
@@ -116,4 +119,143 @@
...
+    - Common
+    - Science
+  - type: ActiveRadio
+    channels:
+    - Service
+    - Binary
+    - Common
+    - Science
+  - type: BorgChassis
+    maxModules: 6
```
FoxxoTrystan, line 155:
```suggestion
    maxModules: 4
```
Normal Service Cyborgs has only 4, 6 is only for Generic Type Borgs.

FoxxoTrystan (changes requested):

Resources/Prototypes/Floof/Entities/Mobs/Cyborgs/quadborg.yml
fenndragon: the service is functioning as the generic for quad borgs due to the number of available modules

FoxxoTrystan: Then make it only a generic borg or due to the following changes: Check current [Service Cyborg](https://github.com/Fansana/floofstation1/blob/master/Resources/Prototypes/Entities/Mobs/Cyborgs/borg_chassis.yml) and there you can see the normal ammount of modules required.

fenndragon: negative, Service is the generic of Quadborgs. that's an intended mechanic.

FoxxoTrystan: Then remove BorgModuleService, were not gona boost Quad to be "better" then the original.

fenndragon: it's not better, it's different. the "generic" quad is security which only gets 4 slots 

instead, seeing as I'm sure you don't want security to have 6 slots, I put it on service. 

unless you would rather I give security 6 instead.

FoxxoTrystan: Then make a generic variant.

fenndragon: there is one, it's called security. which would you rather have the 6 slots

M3739:
Just so you know, I am committed to this matter. Once the changes are finalized on fenndragon's end, and everything checks out, I'll revise and update the guidebook entry accordingly, along with any relevant `GuideHelpComponent` along the way.

Happy to oblige.

FoxxoTrystan (approved):

Memeji (approved):

fenndragon:
cool, I guess that means this is ready to merge

PR #545, opened 2025-02-11
Adds Backslots to Sec Armor and Whatever Else I Added It To

VividPups:
Backslot added back

# Description

highly request People who want to get their backshots I mean slots back 

# Changelog

:cl:
- add: add blackslot back to all security armor

### DIFF
--- a/Resources/Prototypes/DeltaV/Entities/Clothing/OuterClothing/armor.yml
+++ b/Resources/Prototypes/DeltaV/Entities/Clothing/OuterClothing/armor.yml
@@ -1,6 +1,6 @@
 #Basic armor vest replacement
 - type: entity
-  parent: ClothingOuterBaseMedium
+  parent: [ClothingOuterBaseMedium, AllowSuitStorageClothing] #floof added by popular demand to allow backslot
   id: ClothingOuterArmorPlateCarrier
   name: plate carrier
   description: A large and bulky carrier featuring steel plates that offer decent protection against gunfire.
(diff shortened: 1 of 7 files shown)


ProphetCrow:
suggest to also include clothingouterwintersec as its an armored coat that is a popular choice on stations like glacier
image

ShadesMars (approved):

Memeji (approved):

Mnemotechnician (changes requested):
This should use entity parenting, but in a different way - adding suit storage slots to base coats/suits instead of editing the parents of each induvidual coat/suit. More coats may be added in the future, and those changes won't be reflected in them if this PR is merged as-is.

VividPups:
> This should use entity parenting, but in a different way - adding suit storage slots to base coats/suits instead of editing the parents of each induvidual coat/suit. More coats may be added in the future, and those changes won't be reflected in them if this PR is merged as-is.

Well from what I saw the way I added was because the blackslot storage itself a parent (which i thought was kinda of odd)

Mnemotechnician:
> Well from what I saw the way I added was because the blackslot storage itself a parent (which i thought was kinda of odd)

That is the correct way, but where you added it is incorrect. You should add the parent to ClothingOuterArmorBasic and other similar base prototypes instead of editing every single armor piece that is a child of it. Likewise, you should add the parent to ClothingOuterWinterCoat instead of editing every single winter coat.

VividPups:
image
this is waht happens when i added to the based

VividPups:
Tested multiple different Winter coats, armors and jackets all of them accepted an weapon. this was tested As a human

Mnemotechnician (approved):
This will do, but you really just added suit storage slots to *all* outer clothing, including jackets and the like. I don't think it will be a major issue, though.

PR #546, opened 2025-02-11
Add Crew Monitor Alerts and Update Surgery Kits

Forzii:
# Description

As requested, cherry-picks the following Delta-V PRs:
https://github.com/DeltaV-Station/Delta-v/pull/2930
https://github.com/DeltaV-Station/Delta-v/pull/2951

---

# TODO

- [x] Cherry-pick https://github.com/DeltaV-Station/Delta-v/pull/2930
- [x] Cherry-pick https://github.com/DeltaV-Station/Delta-v/pull/2951
- [x] Fix isPowered check breaking handheld crew monitor alert

---

<p>

image
image

</p>
</details>

---

# Changelog

:cl:
- tweak: The crew monitor handheld and console now beep when a crew member goes into critical condition or dies!
- tweak: surgical duffels and syndicate surgical duffels now contain a nitrous oxide tank


Content.Server/Medical/CrewMonitoring/CrewMonitoringConsoleSystem.cs
```diff
@@ -43,6 +49,48 @@ private void OnPacketReceived(EntityUid uid, CrewMonitoringConsoleComponent comp
...
+        if (!component.AlertsEnabled)
+            return;
+
+        // station power (for the machine version)
+        if (!this.IsPowered(uid, EntityManager))
+            return;
+
+        // cell power (for the handheld)
+        if (!_cell.HasActivatableCharge(uid))
+            return;
```
Eagle0600, line 63: Replace with
```
        // station power (for the machine version) or cell power (for the handheld)
        if (!this.IsPowered(uid, EntityManager) && !_cell.HasActivatableCharge(uid))
            return;
```

Resources/Prototypes/Entities/Objects/Specific/Medical/handheld_crew_monitor.yml
```diff
@@ -64,6 +64,8 @@
     useRate: 10
   - type: StaticPrice
     price: 750
+  - type: CrewMonitoringConsole # DeltaV - disable crew crit / dead beeping alerts
+    alertsEnabled: false
```
Eagle0600, line 68: Remove this line.

Resources/Prototypes/Entities/Objects/Specific/Medical/handheld_crew_monitor.yml
```diff
@@ -88,6 +90,8 @@
     sprite: Objects/Specific/Medical/syndihandheldcrewmonitor.rsi
   - type: PowerCellDraw
     useRate: 40
+  - type: CrewMonitoringConsole # DeltaV - disable crew crit / dead beeping alerts
+    alertsEnabled: false
```
Eagle0600, line 94: Remove this line.

Resources/Prototypes/Entities/Objects/Specific/Medical/handheld_crew_monitor.yml
Forzii: This disables the alerts for the spy monitor, it doesn't affect the actual handheld crew monitor

Resources/Prototypes/Entities/Objects/Specific/Medical/handheld_crew_monitor.yml
Forzii: This disables the alerts for the spy monitor, it doesn't affect the actual handheld crew monitor

Resources/Prototypes/Entities/Objects/Specific/Medical/handheld_crew_monitor.yml
Eagle0600: You're right, I suppose that was me focusing on the first think that looked like it _might_ have been the problem.

Content.Server/Medical/CrewMonitoring/CrewMonitoringConsoleSystem.cs
Forzii: So, the issue with this is that the console then makes the alert even when unpowered. Arguably a less important bug but still not quite as intended :(

Quanteey: Hey! Original author here. Significant difference between the DeltaV codebase and yours is that at DeltaV the `.IsPowered` returns true for any case where the check is not suitable (No ApcPowerReceiverComponent = not a machine). Same is true for `HasActivatableCharge`, if there is no power cell component then it defaults to true (and this is why the seemingly AND relation between the two checks can work)

For you in `Content.Server/Power/EntitySystems/StaticPowerSystems.cs` the isPowered default is changed to false, while HasActivatableCharge is still defaulting to true. You could get around this by checking for `ApcPowerReceiverComponent `manually before the `IsPowered` check.

Quanteey: Hey! Original author here. A dfference between the two codebases is that over at DeltaV the `.IsPowered` function returns true for any case where the check is not suitable (No `ApcPowerReceiverComponent` = not a machine). Same is true for `HasActivatableCharge`, if there is no power cell component then it defaults to true (and this is why the seemingly AND relation between the two checks can work)

For you in `Content.Server/Power/EntitySystems/StaticPowerSystems.cs` the isPowered default is changed to false, while `HasActivatableCharge` is still defaulting to true. You could get around this by checking for `ApcPowerReceiverComponent` manually before the IsPowered check. The AND gate won't work because `HasActivatableCharge` will always return true in case of a machine.

Forzii: > Hey! Original author here. Significant difference between the DeltaV codebase and yours is that at DeltaV the `.IsPowered` returns true for any case where the check is not suitable (No ApcPowerReceiverComponent = not a machine). Same is true for `HasActivatableCharge`, if there is no power cell component then it defaults to true (and this is why the seemingly AND relation between the two checks can work)
> 
> For you in `Content.Server/Power/EntitySystems/StaticPowerSystems.cs` the isPowered default is changed to false, while HasActivatableCharge is still defaulting to true. You could get around this by checking for `ApcPowerReceiverComponent `manually before the `IsPowered` check.

Heyyy, appreciate you! Thanks for making this, and thanks again for the insight. I'm pretty new to SS14 work. I was thinking of checking for `ActivatableUIRequiresPowerComponent` though I needed to uh, learn how to do that. I think I just got it with some help from one of the Floof maintainers.

Forzii: Alrighty then. I do NOT want to try to catch up to upstreams (or wait for the maintainers to) just to possibly resolve the isPowered differences (or modify isPowered directly) so work-around it is! New commit coming shortly - just tested it.

Memeji (approved):

Mnemotechnician (approved):

PR #548, opened 2025-02-12
Increase Standard Oxygen to 23%

Eagle0600:
# Description

This PR increases standard station-side atmosphere to be 23% oxygen. It also increases standard freezer pressure to one atmosphere from approximately 0.8 atmospheres. These changes will have no effect until a mapper runs `fixgridatmos`, but likewise does not interfere with ongoing mapping efforts.

It also adds the testing tools I used to gather the data that informs this PR.

The data from my test is shown in the media section of the test, but the key conclusion is as follows: Increased oxygen also increases the amount of CO2 that is comfortable for crew by approximately a sixth of a percent per percent oxygen, and the amount of CO2 that is survivable for crew by approximately a third of a percent per percent oxygen.

While I considered increasing the standard oxygen in oxygen canisters and oxygen tanks as well, 21% is actually fine for all crew as is. The advantage gained by increasing standard round-start oxygen content is to increase how _long_ crew can comfortably breathe for, as they inevitably replace the existing O2 with CO2.

This is necessary because the existing atmospheric tools do not do a good job of replacing consumed O2, as every mol of oxygen consumed ends up scrubbed and replaced a mix of O2 and N2 rather than pure O2. Fixing this is outside of the scope of this PR and requires either smarter vents (probably undesirable), smarter scrubbers (my choice), or more active air circulation (arguably more difficult to get stable pressures).

If one of the above fixes for atmospherics is implemented, my suggestion is to revert this O2 concentration change.

---

<h3>Data</h3>
<p>

Thorough CO2 Test

</p>
<h3>Comments and Limitations</h3>
<p>Each cell had both a harpy and a human in it, and the test started simultaneously for all mobs (using map init). Both humans and harpies died at roughly the same rate, but with significant variance in certain cells. The variance followed no identifiable pattern, and which mob died first varied between human and harpy. The first mob to die in the test was a human, but not by much.</p>

<p>Some of the mobs still standing in this image are in the process of slowly dying, some <i>very</i> slowly</p>

<p>It is important to note some of the variables controlled for in this experiment. The harpies and humans shared cells, and any changes they made to the cell atmos were reset every atmos device update. Any actual difference in performance between humans and harpies may have been hidden by these controlled variables. In this case, the differences could only be the way in which the mobs themselves changed their local atmosphere, for example the CO2 they expel.</p>

<p>One possible way to show this difference would be to create two arrays, one for harpies and one for humans, and see if those differences become apparent. Another possible test would be a long-term survival duration test in a cells which are not constantly being reset, but this would require different tools to the ones I made for this test. Possibly something along the lines of an altered fixgridatmos command.</p>

<p>Exactly one of the cells which was lethal (23.5% O2, 2% CO2) stopped being lethal once one of the occupants died (the human), and the other occupant very slowly recovered.</p>

<p>Whether or not an entity is slowly recovering or dying from asphyxiation appears to be dependant partially on randomness at close edge cases, such as the case of the above-mentioned cell. While the dominant trend is unlikely to show a difference, this experiment is likely to have slightly different results if repeated.</p>
</details>

---

# Changelog

:cl:
- tweak: Changed standard station atmos to be 23% Oxygen. This change will only be apparent after remapping.

### DIFF
--- a/Content.Server/Atmos/EntitySystems/AtmosphereSystem.Commands.cs
+++ b/Content.Server/Atmos/EntitySystems/AtmosphereSystem.Commands.cs
@@ -61,8 +61,9 @@ public sealed partial class AtmosphereSystem
        mixtures[5].Temperature = 5000f;
 
        // 6: (Walk-In) Freezer
-       mixtures[6].AdjustMoles(Gas.Oxygen, Atmospherics.OxygenMolesStandard);
-       mixtures[6].AdjustMoles(Gas.Nitrogen, Atmospherics.NitrogenMolesStandard);
+	   // TODO: Adjust pressure? (test first)
+       mixtures[6].AdjustMoles(Gas.Oxygen, Atmospherics.OxygenMolesStandard * Atmospherics.T20C / 235f);
+       mixtures[6].AdjustMoles(Gas.Nitrogen, Atmospherics.NitrogenMolesStandard * Atmospherics.T20C / 235f);
        mixtures[6].Temperature = 235f; // Little colder than an actual freezer but gives a grace period to get e.g. themomachines set up, should keep warm for a few door openings
 
        foreach (var arg in args)
--- a/Content.Server/Atmos/Piping/Other/Components/GasTesterComponent.cs
+++ b/Content.Server/Atmos/Piping/Other/Components/GasTesterComponent.cs
@@ -1,0 +1,24 @@
+using Content.Shared.Atmos;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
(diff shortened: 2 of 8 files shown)


Eagle0600:
The error message appears to be unrelated to any changes made in my PR. Can I request a re-run to check?

Mnemotechnician (approved):
I am... surprised by how much efforts went into this.

Eagle0600:
I decided that no-one actually knew how this worked, and any attempts at a solution were a wild shot in the dark. The solution was to test it, but the existing dev tools aren't good enough for it.

FoxxoTrystan (approved):

PR #552, opened 2025-02-13
Glacier Local 

VividPups:
# Description
As we know, an admin plays this local on a glacier named Katherine. However, it seems they usually have to take a lot of time to set up this character. This character is very well known throughout the community and is always a consent when glacier is on. So I reached ask if they would like an outfit made for them to make it much faster for them,. Since this character always brings a lot of RP when ever they are here 

# Media

N/A

---

# Changelog
:cl:
- add: Added Glacier Local outfit.

### DIFF
--- a/Resources/Prototypes/Floof/Catalog/Fills/Bags/satchel.yml
+++ b/Resources/Prototypes/Floof/Catalog/Fills/Bags/satchel.yml
@@ -1,0 +1,17 @@
+- type: entity
+  id: SatchelGlacierLocalFilled
+  suffix: Glacier Local, Filled
+  parent: ClothingBackpackSatchelLeather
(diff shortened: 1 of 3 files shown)


VividPups:
uumm not sure if this is a real test or not. not understanding how I would fix it. looks like something to do with a harpy?

Mnemotechnician:
I am not a fan of adding such "exclusive" outfits to the game. Admins have access to the loadprototype command which they can use to load those prototypes as needed.

VividPups:
the reason why I made it was because of the time it takes for them to load all of those.

Fansana:
Too specific, instead of doing this loading this as a prototype is much faster.

PR #562, opened 2025-02-17
Ice Metabolizes to Water, Lowers Temperature, and Causes Damage Once Frozen

Eagle0600:
# Description

Ice is now a valid way to satiate thirst, but drinking a bucketful of ice is now a very bad idea.

From testing, a 30u glass is not enough to get an average human to frozen, but it will be more dangerous for smaller characters to drink large quantities. Nonetheless, it will never do more than 1 damage per 1u drank, so even the tiniest characters aren't going to die from drinking a glass.

By request, I also took a look at ice cream, but decided not to do what was requested and instead made both ice cream and cream satiate hunger. This is because fat (the primary component of both) is a known valid survival food. Cream breaks down into nutriment, and ice cream breaks down into nutriment and sugar. Both still satiate thirst.

---

# Changelog

:cl:
- tweak: Ice now satiates thirst, but also makes you cold. Don't drink too much!
- tweak: Cream and ice cream satiate hunger.

### DIFF
--- a/Resources/Prototypes/Reagents/Consumable/Drink/drinks.yml
+++ b/Resources/Prototypes/Reagents/Consumable/Drink/drinks.yml
@@ -53,6 +53,9 @@
       effects:
       - !type:SatiateThirst
         factor: 1
+      - !type:AdjustReagent # Floof - Cream breaks down into fat.
+        reagent: Nutriment # Representing fat. Fat itself is not used because sugar needs nutriment to metabolise.
+        amount: 0.15
 
 - type: reagent
   id: CoconutWater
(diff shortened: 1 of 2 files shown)


Fansana:
hey random idea, can you make icecream metabolize into ice?

Eagle0600:
Since ice cream is mostly fat and suger, it should probably metabolise into those. It already satiates thirst, too, but I can make it metabolise into water instead if you prefer.

Eagle0600:
Okay, I'm having ice cream break down into nutriment (representing fat content) and sugar, while satiating thirst directly. The decision not to break down into ice is motivated to avoid being one reagent away (cream) from making more ice cream inside the body. This is because one of its ingredients (sugar) is also a base nutrition reagent, so I definitely want to break down into that.

Mnemotechnician:
I think 4e55fa798f489c16992671fe8516286184fc984d accidentally reverted the ice part of the PR?

Eagle0600:
Oh what the... thanks for pointing that out, hang on a moment.

Eagle0600:
There, hopefully that fixes it.

Mnemotechnician (approved):

FoxxoTrystan (approved):

PR #568, opened 2025-02-18
Vore fixes and consent menu improvement

LucasTheDrgn:
# Description

- Re-splits the vore consent options, plenty of users want one side but not the other.
- - This is not just a rollback, this includes checking for the proper consent options when deciding if the examine text should include vore info
- Adds a unique popup text for inserting yourself vs being devoured (making it more obvious to both parties when this is happening)
- Adds nuance to the checks on how fast the doafter should be when devouring vs inserting (previously, it would use the downed/stamina status of the prey regardless of who initiated the action
- Fixes bugs involving carrying someone and devouring them/inserting yourself into them
- Sorting added to the consent toggle window so they should appear not only in a consistent order, but in an intentional order.
- - For example, all vore options are grouped together.
- - This system is expandable if proper categories are added in the future

---

# Changelog

:cl:
- add: Unique popup for inserting yourself into a pred
- tweak: Vore consent options are separate again
- tweak: Consent menu has been given consistent sorting
- fix: Insert Yourself verb now checks the stamina/downed status of the pred for doafter speed
- fix: Devouring or inserting yourself into someone you're carrying no longer causes issues

### DIFF
--- a/Content.Client/Consent/UI/Windows/ConsentWindow.xaml.cs
+++ b/Content.Client/Consent/UI/Windows/ConsentWindow.xaml.cs
@@ -157,7 +157,10 @@ public sealed partial class ConsentWindow : FancyWindow
 
         _entries.Clear();
 
-        var consentprototypelist = _protoManager.EnumeratePrototypes<ConsentTogglePrototype>();
+        var consentprototypelist = new List<ConsentTogglePrototype>(_protoManager.EnumeratePrototypes<ConsentTogglePrototype>());
+
+        consentprototypelist.Sort();
+
         foreach (var prototype in consentprototypelist)
             AddConsentEntry(prototype);
 
--- a/Content.Shared/Consent/ConsentTogglePrototype.cs
+++ b/Content.Shared/Consent/ConsentTogglePrototype.cs
@@ -1,13 +1,38 @@
 namespace Content.Shared.Consent;
 
+using System;
 using Robust.Shared.Prototypes;
 
 /// <summary>
 /// TODO
 /// </summary>
 [Prototype("consentToggle")]
-public sealed partial class ConsentTogglePrototype : IPrototype
+public sealed partial class ConsentTogglePrototype : IPrototype, IComparable
 {
     [IdDataField]
     public string ID { get; private set; } = default!;
+
+    [DataField("category")]
+    public string Category { get; private set; } = default!;
+
+    [DataField("priority")]
+    public int priority { get; private set; } = 0;
+
+    public int CompareTo(object? obj) { // Allow for granular sorting to make the menu display consistently and intuitively
+        if (obj == null) return -1;
+
+        ConsentTogglePrototype? other = obj as ConsentTogglePrototype;
+        if (other != null){
+            var cat = this.Category.CompareTo(other.Category);
+            if (cat != 0){
+                return cat; // Categories are different, sort by category
+            }
+            if (this.priority != other.priority) {
+                return this.priority - other.priority; // Priorities are different, sort by priority
+            }
+            return this.ID.CompareTo(other.ID); // Category and priority are the same, sort by ID
+        } else {
+           throw new ArgumentException("Object is not a Consent Toggle");
+        }
+    }
 }
--- a/Content.Server/FloofStation/VoreSystem.cs
+++ b/Content.Server/FloofStation/VoreSystem.cs
@@ -36,6 +36,7 @@ using Content.Server.Power.Components;
 using Content.Server.Nutrition.EntitySystems;
 using Content.Shared.Interaction.Events;
 using Content.Shared.Hands.EntitySystems;
+using Content.Server.Carrying;
 
 namespace Content.Server.FloofStation;
 
@@ -61,6 +62,7 @@ public sealed class VoreSystem : EntitySystem
     [Dependency] private readonly SharedTransformSystem _transform = default!;
     [Dependency] private readonly FoodSystem _food = default!;
     [Dependency] private readonly SharedHandsSystem _hands = default!;
+    [Dependency] private readonly CarryingSystem _carrying = default!;
 
     public override void Initialize()
     {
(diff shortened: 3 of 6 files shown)


Fansana (approved):

Content.Shared/Consent/ConsentTogglePrototype.cs
```diff
@@ -1,13 +1,38 @@
...
     [IdDataField]
     public string ID { get; private set; } = default!;
+
+    [DataField("category")]
+    public string Category { get; private set; } = default!;
+
+    [DataField("priority")]
+    public int priority { get; private set; } = 0;
+
+    public int CompareTo(object? obj) { // Allow for granular sorting to make the menu display consistently and intuitively
```
Mnemotechnician, line 21: Missing override modifier:
```suggestion
    public override int CompareTo(object? obj) { // Allow for granular sorting to make the menu display consistently and intuitively
```

Content.Shared/Consent/ConsentTogglePrototype.cs
```diff
@@ -1,13 +1,38 @@
...
+                return cat; // Categories are different, sort by category
+            }
+            if (this.priority != other.priority) {
+                return this.priority - other.priority; // Priorities are different, sort by priority
+            }
+            return this.ID.CompareTo(other.ID); // Category and priority are the same, sort by ID
+        } else {
+           throw new ArgumentException("Object is not a Consent Toggle");
+        }
+    }
```
Mnemotechnician, line 37: Code style: use guard clauses, prefer brace-less if conditions
```suggestion
        if (obj is not ConsentTogglePrototype other)
            return -1;
        
        var cat = this.Category.CompareTo(other.Category);
        if (cat != 0)
            return cat; // Categories are different, sort by category
        if (this.priority != other.priority)
            return this.priority - other.priority; // Priorities are different, sort by priority
        
        return this.ID.CompareTo(other.ID); // Category and priority are the same, sort by ID
```

Content.Client/Consent/UI/Windows/ConsentWindow.xaml.cs
```diff
@@ -157,7 +157,10 @@ public void UpdateUi()
 
         _entries.Clear();
 
-        var consentprototypelist = _protoManager.EnumeratePrototypes<ConsentTogglePrototype>();
+        var consentprototypelist = new List<ConsentTogglePrototype>(_protoManager.EnumeratePrototypes<ConsentTogglePrototype>());
+
+        consentprototypelist.Sort();
+
```
Mnemotechnician, line 163: Code style: use the .ToList extension method:
```suggestion
        var consentprototypelist = _protoManager.EnumeratePrototypes<ConsentTogglePrototype>().ToList();
        consentprototypelist.Sort();

```

Content.Shared/Consent/ConsentTogglePrototype.cs
```diff
@@ -1,13 +1,38 @@
...
 /// </summary>
 [Prototype("consentToggle")]
-public sealed partial class ConsentTogglePrototype : IPrototype
+public sealed partial class ConsentTogglePrototype : IPrototype, IComparable
 {
     [IdDataField]
     public string ID { get; private set; } = default!;
+
+    [DataField("category")]
+    public string Category { get; private set; } = default!;
```
Mnemotechnician, line 16: `string` is a class (System.String), so the `default` value for it is `null`. This will lead to an NPE if someone forgets to set up a category for their consent prototype. You should either add `required: true` to the data field attribute, or set up a different default value, or make the field nullable.

Mnemotechnician (changes requested):
Mostly minor code style issues.

Content.Shared/Consent/ConsentTogglePrototype.cs
LucasTheDrgn: This is actually giving me an error when compiling, saying "no suitable method found to override". Could it be that the signature is incorrect and a different signature should be used? Or is it more likely that, since I'm pulling the IComparable interface in for this prototype specifically, and prototypes normally aren't comparable like this, that there isn't anything for it to override?

Content.Client/Consent/UI/Windows/ConsentWindow.xaml.cs
LucasTheDrgn: This data type doesn't have a ToList function, but System.Linq.Enumerable has a ToList<T>(IEnumerable<T> source) function if that's acceptable

Content.Shared/Consent/ConsentTogglePrototype.cs
LucasTheDrgn: Defaulting to "" to bring uncategorized toggles to the top of the list (and uncategorizing NoClone as an example)

Content.Client/Consent/UI/Windows/ConsentWindow.xaml.cs
Mnemotechnician: yeah, your IDE is supposed to import that method automatically (if you are using an IDE). We don't typically type out imports ourselves.

Content.Shared/Consent/ConsentTogglePrototype.cs
Mnemotechnician: Uh, I might be wrong and C# might not actually support the override modifier for virtual methods coming from an interface. If that's the case, this suggestion is not relevant.

Content.Client/Consent/UI/Windows/ConsentWindow.xaml.cs
LucasTheDrgn: Yeah, since most of the changes I've made so far were just yml or minor code changes I haven't bothered setting up an IDE, just using Sublime with syntax highlighting, but I am starting to poke around more with the code so that'll probably be worth the effort to set up soon.

Content.Shared/Consent/ConsentTogglePrototype.cs
LucasTheDrgn: Fair enough! I appreciate it being brought up, I'd rather have it intentionally not there then just forgotten or assumed.

LucasTheDrgn:
I think it's mad because I pulled some of the code changes that were suggested using the web editor, but I assure you I have pulled everything to local and tested all the edit that were pulled and everything compiled and ran successfully.

Mnemotechnician (approved):
Approving now, even without that small thing I mentioned on discord.

LucasTheDrgn:
Realized what you meant and had a bout of insomnia so I finally set up the IDE and made those changes, thank you for the Learning Moment

Memeji (approved):

FoxxoTrystan (approved):

PR #571, opened 2025-02-19
Subtle OOC

FoxxoTrystan:
# Description

This PR adds Sublte OOC made by april; requested by the floof team.
And while im at it; porting also filters for admins/players regarding suble.

Following PRs ported:
https://github.com/TheDenSS14/TheDen/pull/159 (Only Chat Changes)
https://github.com/TheDenSS14/TheDen/pull/163

---

# Changelog

:cl:
- add: Ported subtle OOC (Prefix: { )

### DIFF
--- a/Content.Server/Chat/Systems/ChatSystem.cs
+++ b/Content.Server/Chat/Systems/ChatSystem.cs
@@ -187,7 +187,8 @@ public sealed partial class ChatSystem : SharedChatSystem
         string? nameOverride = null,
         bool checkRadioPrefix = true,
         bool ignoreActionBlocker = false,
-        LanguagePrototype? languageOverride = null
+        LanguagePrototype? languageOverride = null,
+        string? color = null
         )
     {
         if (HasComp<GhostComponent>(source))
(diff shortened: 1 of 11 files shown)


Fansana (approved):

Content.Shared/Chat/SharedChatSystem.cs
```diff
@@ -22,6 +21,7 @@ public abstract class SharedChatSystem : EntitySystem
     public const char EmotesPrefix = '@';
     public const char EmotesAltPrefix = '*';
     public const char SubtlePrefix = '-';
+    public const char SubtleOOCPrefix = '.';
```
Mnemotechnician, line 24: Dot is already used for local chat by default, this is bad.

Content.Server/Chat/Systems/ChatSystem.cs
```diff
@@ -187,7 +187,8 @@ public void TrySendInGameICMessage(
         string? nameOverride = null,
         bool checkRadioPrefix = true,
         bool ignoreActionBlocker = false,
-        LanguagePrototype? languageOverride = null
+        LanguagePrototype? languageOverride = null,
+        string? color = null
```
Mnemotechnician, line 191: Why are we introducing a color parameter, that is only used in 2 out of 7 code branches, that is only ever has the value of SubtleOOCColor..? It could be referenced statically or even embedded in the .ftl file...

Mnemotechnician (approved):

Memeji (approved):

PR #572, opened 2025-02-20
Bloodpack Related Additions

Cprn-Sprkl:
# Description

Adds a sterile bag that can be printed from the medical and security techfabs.

Adds recipes for creating bloodpacks from sterile bags and blood, a blood and saline mix, or a perfluorocarbon based blood substitute.

Adds a chemical reaction for the production of blood.

Adds an extra section to the recipe sorter that ensures recipes with greater quantities of reagents are sorted higher than recipes with less. This only gets checked if the recipes use the same amount of solid and unique reagents.

---

# Changelog

:cl: sprkl
- add: Added sterile bags which can be processed into blood packs via a microwave.
- add: Added a recipe to synthesize blood.

### DIFF
--- a/Content.Shared/Kitchen/MicrowaveMealRecipePrototype.cs
+++ b/Content.Shared/Kitchen/MicrowaveMealRecipePrototype.cs
@@ -53,6 +53,7 @@ namespace Content.Shared.Kitchen
             return n;
         }
 
+        //Floofstation specific method - Start
         /// <summary>
         ///     Sums the quantity of reagents in a recipe for sorting the recipe list.
         ///     A fallback check if the IngredientCount is equal when sorting the
@@ -68,5 +69,6 @@ namespace Content.Shared.Kitchen
             }
             return n;
         }
+        //Floofstation specific method - End
     }
 }
--- a/Content.Shared/Kitchen/MicrowaveMealRecipePrototype.cs
+++ b/Content.Shared/Kitchen/MicrowaveMealRecipePrototype.cs
@@ -52,5 +52,21 @@ namespace Content.Shared.Kitchen
             }
             return n;
         }
+
+        /// <summary>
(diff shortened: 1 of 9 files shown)


Content.Shared/Kitchen/MicrowaveMealRecipePrototype.cs
```diff
@@ -52,5 +52,21 @@ public FixedPoint2 IngredientCount()
...
             return n;
         }
+
+        /// <summary>
+        ///     Sums the quantity of reagents in a recipe for sorting the recipe list.
+        ///     A fallback check if the IngredientCount is equal when sorting the
+        ///     recipe list.
+        /// </summary>
+        /// <returns></returns>
+        public FixedPoint2 ReagentQuantity()
```
Mnemotechnician, line 62: Please add a comment indicating that this method is floof-specific (either one comment before the block, or two comments surrounding it (section start/end)

Content.Shared/Kitchen/RecipeManager.cs
```diff
@@ -38,7 +38,15 @@ public override int Compare(FoodRecipePrototype? x, FoodRecipePrototype? y)
 
                 var nx = x.IngredientCount();
                 var ny = y.IngredientCount();
-                return -nx.CompareTo(ny);
+                if (-nx.CompareTo(ny) != 0)
```
Mnemotechnician, line 41: Same as the above. It's also preferable to either keep the original line in the commented-out form, or to describe in a comment how the behavior was changed compared to upstream.

Mnemotechnician (changes requested):
Looks good, just needs a few floofstation edit comments.

Also, I updated your changelog to match [the standards](https://discord.com/channels/1255902263667331193/1255992182880473269/1288518199515811922), feel free to adjust it further if needed.

Mnemotechnician (approved):

Memeji (approved):

PR #580, opened 2025-02-23
Remove Deprecated JobPrototype Whitelist Requirements

M3739:
Round 2153 was arguably a horrible round for both Security and Justice members involved. However, after some discussion with some fellow players and revelations, we found out that the whitelist system for Chief Justice is actually broken. Along with the CJ at the time not being whitelisted according to public sources within the Discord server.

# Description

This PR seeks to fix the aforementioned issue, which if not clear is as entails;

Currently at the moment you need 600 minutes and a "whitelist" to play as Chief Justice. However, due to an oversight in the YAML, this whitelist actually never existed.

For example, the `captain.yaml` utilizes the statement `whitelisted: true` to implement it. But for Chief Justice, this is instead utilizing the `!type;CharacterWhitelistRequirement` to attempt to accomplish this. However, this may likely be reliant on a non-existent component or system from DeltaV, and thus, it does not work as expected.

To correct this, `chief_justice.yml` now utilizes the `whitelisted: true` field instead of the flawed requirement as stated before. This change has also been attempted to be reflected to the Nuclear Operative Commander as it too used the same flawed requirement, but as it turns out, the antags work a bit differently on that regard. (For context, I never got whitelisted for ANY role, and I have access to Nuke Op Commander on live servers.)

HOW

LIVE M3739 REACTION

Edit: Originally, this PR was to fix what was thought to be a botched whitelist, and I was informed that the said whitelist never actually existed to begin with. This PR now removes the deprecated way of whitelisting jobs, so now players should no longer be confused as to Chief Justice needing a whitelist.

The rest of this PR's post is unaltered for posterity.

---

# Changelog

:cl: M3739
- fix: Central Command has conducted a review of it's assignment system. Chief Justice should no longer be erroneously reporting that the job needs a whitelist.

### DIFF
--- a/Resources/Prototypes/DeltaV/Roles/Jobs/Justice/chief_justice.yml
+++ b/Resources/Prototypes/DeltaV/Roles/Jobs/Justice/chief_justice.yml
@@ -3,6 +3,7 @@
   name: job-name-chief-justice
   description: job-description-chief-justice
   playTimeTracker: JobChiefJustice
+  whitelisted: true # Floof - CJ whitelist
   requirements:
     - !type:CharacterPlaytimeRequirement
       tracker: JobClerk
@@ -15,7 +16,7 @@
       min: 36000 # 10 hours
     - !type:CharacterOverallTimeRequirement
       min: 90000 # 25 hours
-    - !type:CharacterWhitelistRequirement # whitelist requirement because I don't want any dingus judges
+    #- !type:CharacterWhitelistRequirement # whitelist requirement because I don't want any dingus judges | Floof - M3739 - How this made it past prod for this long is beyond me.
   weight: 20
   startingGear: CJGear
   icon: "JobIconChiefJustice"
--- a/Resources/Prototypes/DeltaV/Roles/Jobs/Justice/chief_justice.yml
+++ b/Resources/Prototypes/DeltaV/Roles/Jobs/Justice/chief_justice.yml
@@ -3,7 +3,7 @@
   name: job-name-chief-justice
   description: job-description-chief-justice
   playTimeTracker: JobChiefJustice
-  whitelisted: true # Floof - CJ whitelist
+  # whitelisted: true # Floof - CJ whitelist
   requirements:
     - !type:CharacterPlaytimeRequirement
       tracker: JobClerk
@@ -16,7 +16,7 @@
       min: 36000 # 10 hours
     - !type:CharacterOverallTimeRequirement
       min: 90000 # 25 hours
-    #- !type:CharacterWhitelistRequirement # whitelist requirement because I don't want any dingus judges | Floof - M3739 - How this made it past prod for this long is beyond me.
+    #- !type:CharacterWhitelistRequirement # whitelist requirement because I don't want any dingus judges | Floof - M3739 - Does not work properly, do not use.
   weight: 20
   startingGear: CJGear
   icon: "JobIconChiefJustice"
--- a/Resources/Prototypes/DeltaV/Roles/Jobs/Justice/chief_justice.yml
+++ b/Resources/Prototypes/DeltaV/Roles/Jobs/Justice/chief_justice.yml
@@ -3,7 +3,7 @@
   name: job-name-chief-justice
   description: job-description-chief-justice
   playTimeTracker: JobChiefJustice
-  # whitelisted: true # Floof - CJ whitelist
+  # whitelisted: true # whitelist requirement because I don't want any dingus judges | Floof - Chief Justice is not whitelisted.
   requirements:
     - !type:CharacterPlaytimeRequirement
       tracker: JobClerk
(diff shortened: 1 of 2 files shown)


M3739:
As it turns out, antags are a bit special in that they don't accept the `whitelisted: true` statement.

Odd. Change reversed for the Nuclear Operative Commander.

Mnemotechnician:
This whitelist requirement actually came from delta-v. But floof uses that whitelist system to determine whether a player should be able to join, so everyone who can join is whitelisted. What you added is a job whitelist requirement - a different thing. This should probably require admin approval and a prior notice because right now no person is whitelisted for CJ.

M3739:
> The old whitelist requirement actually came from delta-v. But floof uses that whitelist system to determine whether a player should be able to join, so everyone who can join is whitelisted. What you added is a job whitelist requirement - a different thing. This should probably require admin approval and a prior notice because right now no person is whitelisted for CJ.

I see. Then it's unclear as to what the intention is to be exact. The flawed method was no doubt an artifact from DeltaV, but it wasn't addressed until now it would seem. For us, we were under the impression of needing a whitelist due to this message in the job selection;
image

Personally, I feel that Chief Justice should be a whitelisted role, much like the Captain. As a great deal lies on them as the judge when it comes to trials. Be it civil, or criminal. As round 2153 showcases, it can make or break the entirety of sec, or the department. Afterall, should the Captain be charged with Abuse of Power, the Chief Justice will preside over that case.

This also does not create a Catch 22, as the Clerk can also stand in as a judge if need be, allowing for prior experience to rightfully earn the mantle of Chief Justice. This is a role with a heavy amount of gravitas to it. In some cases, more than the Captain.

M3739:
I have been informed that the official decision as of writing this PR is that the **Chief Justice role is not whitelisted**

**Until an official decision declaring otherwise has been made, the PR has been adjusted to now reflect that existing decision.** The Chief Justice role should no longer imply any kind of whitelist, as it doesn't have one.

FoxxoTrystan:
chief justice is not suposed to be whitelisted for the moment.

FoxxoTrystan (approved):
No Whitelist needed, CharacterWhitelistRequirement is SERVER WHITELIST while whitelisted: true is job whitelist.

M3739:
> No Whitelist needed, CharacterWhitelistRequirement is SERVER WHITELIST while whitelisted: true is job whitelist.

Wait, so that "flawed" method of doing it, is actually for the whole server?... Then why does the `jobwhitelistadd` command work in "whitelisting" a player for Chief Justice?

M3739:
> > No Whitelist needed, CharacterWhitelistRequirement is SERVER WHITELIST while whitelisted: true is job whitelist.
> 
> Wait, so that "flawed" method of doing it, is actually for the whole server?... Then why does the `jobwhitelistadd` command work in "whitelisting" a player for Chief Justice?
> 
> Personally, this is very confusing to me.

image
Picture taken on the "Ayin" branch, which is the master branch of my fork. Last synced after yesterday's merges.

Fansana:
There are two whitelist systems, one is the old global whitelist system. Then we importent another that is job specific, which we use for the captain.

M3739:
> There are two whitelist systems, one is the old global whitelist system. Then we importent another that is job specific, which we use for the captain. whitelist: true just checks the old system which every player is in otherwise they wouldn't be able to play.

I.. see...
And it seems to add up. I took a gander at DeltaV's current implementation of the Chief Justice job prototype, and I do see the `whitelisted: true` on the `JobPrototype` in it's current iteration. I also took a look at it's code history, and found out that the `!type:CharacterWhitelistRequirement` was removed and instead replaced with it as result of PR number 1678 on the Delta-V repo, which was when they implemented the JobWhitelists panel. `!type:CharacterWhitelistRequirement` was also deprecated as a result, encouraging the use of `whitelisted: true` instead on the `JobPrototype`.

Funnily enough, it seems we have that JobWhitelists panel, it's just the deprecated way of whitelisting jobs wasn't fully addressed at the time, and thus remains as a vestigial structure in the `chief_justice.yml` job prototype. 

Well, lets change that. `ChiefJustice` will just have the `whitelisted: true` commented out to allow for convenience when Floof decides to get on with that part. And the deprecated job whitelist will be removed from the Nuke Ops commander.

That way, now there should be no further confusion and miscommunication that Chief Justice and Nuke Ops commander is whitelisted.

Wild. Well, that clears up that confusion for me.

M3739:
A bit of tidying up, may as well change the title to something more communicative, since it looks like this will be on the back burner for a long while.

No rush on the matter. This is a little thing compared to the bigger fish we got in the pool at the moment.

Mnemotechnician (approved):
I guess that's fair.

FoxxoTrystan (approved):

PR #581, opened 2025-02-23
Give Cargo Techs Order Access by Default

Mnemotechnician:
# Description
A change seemingly demanded by the playerbase, makes orders a standard access for cargo techs rather than extended (low-pop).

# Changelog
:cl:
- tweak: Cargo technicians now get orders access by default.

### DIFF
--- a/Resources/Prototypes/Roles/Jobs/Cargo/cargo_technician.yml
+++ b/Resources/Prototypes/Roles/Jobs/Cargo/cargo_technician.yml
@@ -10,9 +10,10 @@
   access:
   - Cargo
   - Maintenance
+  - Orders # Floof - made standard
   extendedAccess:
   - Salvage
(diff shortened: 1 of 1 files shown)


Forzii:
Should we also give courier orders access during extended (low-pop) or is that too far? There's been times where I've been a lone courier during low-pop and it's a real pain to get access - but I didn't really want to play a cargo tech.

ShadesMars (approved):

Memeji (approved):

Memeji:
> Should we also give courier orders access during extended (low-pop) or is that too far? There's been times where I've been a lone courier during low-pop and it's a real pain to get access - but I didn't really want to play a cargo tech.

I feel it is the job expectation for couriers to deliver. No real need to give them order access as that is out of their scope, compared to cargo techs.

PR #584, opened 2025-02-25
Quads Update

fenndragon:
# Description

adds movement visuals as well as sprite corrections to all quad borgs 

---

# Changelog

:cl:
- add: Added Quad borg animations


Resources/Prototypes/Floof/Entities/Mobs/Cyborgs/quadborg.yml
```diff
@@ -206,6 +241,23 @@
...
+    - RobotTalk
+    - Sign
+    understands:
+    - TauCetiBasic
+    - SolCommon
+    - Tradeband
+    - Freespeak
+    - Elyran
+    - RobotTalk
+    - Sign
```
Mnemotechnician, line 260: This should be removed. A cyborg should never have access to anything other than robotic and common. PAIs have access to most languages ONLY because they have an intended purpose (or rather, one of their potential purposes) as a personal translator. That is not the case for borgs, and even less so quadborgs.

Mnemotechnician (changes requested):

Resources/Prototypes/Floof/Entities/Mobs/Cyborgs/quadborg.yml
fenndragon: Alright then

fenndragon: so funny story that was even added to the wrong spot

fenndragon: alright fixed.

Mnemotechnician (approved):

PR #597, opened 2025-03-02
Medical Balance Tweaks and Fixes

Mnemotechnician:
# Description
- Theobromine: increased the amount of THB added by drinks like coffee, tea, and their derivatives to 0.1 instead of 0.05, because otherwise drinking coffee and tea would never allow enough THB to accumulate in your body to deal damage (if you have an animal stomach that is). Some drinks already had this value set to 0.1 which allowed them to be poisonous to animals, as intended.

- Medical: EE has badly thrown medical balance off by introducing poorly thought out changes (namely, increasing the amount of healing topicals provide by a factor of 3 instead of nerfing space/fire damage that was the cause of that change). Those changes were introduced to floof in the commit b431d24f86f94433ea0c53982db83e40ef1d3f2c. Most of the changes there were reverted, except for the stack sizes of topicals. Bruise packs and ointments still come in stacks of 15, but heal the pre-surgery amount of damage. The healing of chemicals was also reverted to pre-surgery. The time it takes to apply topicals was reverted back to 3 seconds on others and 9 seconds on self, which that commit changed to 2 and 4 respectively.

- Medical: nerfed passive body part healing from 10 damage/minute to 1 damage/minute. This healing is applied to ALL body parts that are below the "heavily wounded" (now 70 damage) state at the same time and has rendered limb damage practically irrelevant on floof outside of prolongued spacings and fires.

# Changelog
:cl:
- tweak: The healing amounts of topicals (bruise packs, ointments, and so on) were reduced to pre-surgery-update amounts.
- fix: Theobromine gained from coffee and similar drinks will now properly accumulate in your body and deal damage to those with animal stomachs.

### DIFF
--- a/Resources/Prototypes/Reagents/medicine.yml
+++ b/Resources/Prototypes/Reagents/medicine.yml
@@ -43,7 +43,7 @@
       - !type:HealthChange
         damage:
           types:
-            Poison: -1.5 # Was 1, Slight Buff as it should heal for half the amount as Dip or Stelli
+            Poison: -1 # Was 1, Slight Buff as it should heal for half the amount as Dip or Stelli # Floof - reverted - why did you touch this??
       - !type:HealthChange
         conditions:
         - !type:ReagentThreshold
@@ -153,7 +153,7 @@
       - !type:HealthChange
         damage:
           groups:
-            Brute: -2.5 # Was 2, Buffed due to limb damage changes
+            Brute: -2 # Was 2, Buffed due to limb damage changes # Floof - reverted
       - !type:HealthChange
         conditions:
         - !type:ReagentThreshold
@@ -242,9 +242,9 @@
       - !type:HealthChange
         damage:
           types:
-            Heat: -2
-            Shock: -2
-            Cold: -2 # Was 1.5, Buffed due to limb damage changes
+            Heat: -1.5
+            Shock: -1.5
+            Cold: -1.5 # Was 1.5, Buffed due to limb damage changes # Floof - reverted
       - !type:HealthChange
         conditions:
         - !type:ReagentThreshold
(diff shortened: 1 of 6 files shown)


Resources/Prototypes/Reagents/medicine.yml
```diff
@@ -1042,7 +1042,7 @@
       - !type:HealthChange
         damage:
           types:
-            Blunt: -4 # Was 3, Buffed due to limb damage changes(GoobStation)
+            Blunt: -3.5 # Was 3, Buffed due to limb damage changes(GoobStation)  # Floof - reverted
```
Forzii, line 1045: Lace got dropped back to 3 but bruiz only to 3.5?

Mnemotechnician: Yea, I used the git diff of b431d24f86f94433ea0c53982db83e40ef1d3f2c to revert it.

Fansana (approved):

Memeji:
Wanted to bring up some feedback that the gauze shouldn't take 9 seconds to apply to self. Reason being that in those 9 seconds you can lose a lot of blood from bleed. Thoughts on this?

Mnemotechnician:
> Wanted to bring up some feedback that the gauze shouldn't take 9 seconds to apply to self. Reason being that in those 9 seconds you can lose a lot of blood from bleed. Thoughts on this?

That's a good point, I halved the self heal penalty on gauze so it should only take 4.5 seconds now

XenoOutlier:
> > Wanted to bring up some feedback that the gauze shouldn't take 9 seconds to apply to self. Reason being that in those 9 seconds you can lose a lot of blood from bleed. Thoughts on this?
> 
> That's a good point, I halved the self heal penalty on gauze so it should only take 4.5 seconds now

Honestly I think you just shouldn't increase the time back to original. 4 seconds on self apply is plenty Topical heals are already being reduced to base og. values they don't really need 2 nerfs. besides that 9 seconds per self-application on topicals is just tonnes of idle time. For reference to use 15 uses of anyting it would take: 2 minutes and 15 seconds of: Simply standing in place waiting on a do-after. The pre nerf time of 60 seconds per stack was kinda plenty. I understand the idea for balance but also double nerfing isn't always a grand idea. Besides that most people I've seen have never complained about topical applications being too fast.

FoxxoTrystan (approved):

PR #598, opened 2025-03-03
Rebalance Mousetraps

Mnemotechnician:
# Description
Changes the mousetrap damage equation to this (X axis represents mass in kg and Y represents damage): 
image

instead of this:
image

Making them less deadly to min size rodentia, and slightly more harmful to species of larger size (70 kg humans still won't feel it, but it will be quite a pinch to them 12-20 kg folk)

# Changelog
:cl:
- tweak: Adjusted the damage scaling of mousetraps, making them less deadly to minimum size rodentia and felinids, but more dangerous to mid-small species.

### DIFF
--- a/Content.Server/Mousetrap/MousetrapSystem.cs
+++ b/Content.Server/Mousetrap/MousetrapSystem.cs
@@ -53,7 +53,7 @@ public sealed class MousetrapSystem : EntitySystem
         // Small - big damage,
         // Large - small damage
         // Yes, I punched numbers into a calculator until the graph looked right
-        var scaledDamage = -50 * MathF.Atan(physics.Mass - component.MassBalance) + 25 * MathF.PI;
+        var scaledDamage = -30 * MathF.Atan((physics.Mass - component.MassBalance) / 4) + 15 * MathF.PI; // Floof - rebalanced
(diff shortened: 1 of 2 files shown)


Mnemotechnician:
Apparently mousetraps also have a base damage of 2, so the actual damage may be 2 times higher than what I calculated? Need to test by myself or someone else...

Forzii:
> Apparently mousetraps also have a base damage of 2, so the actual damage may be 2 times higher than what I calculated? Need to test by myself or someone else...

Yeah, base damage is 2 plus rodentia take additional blunt damage - that's why some of us rodentia end up in medical with well over 300 blunt damage. Still, this looks much more reasonable - still scary with a much smaller chance of accidental round removal (two traps currently gib tiny rodentia).

ShadesMars (approved):

ghost (approved):

Mnemotechnician:
> Yeah, base damage is 2 plus rodentia take additional blunt damage - that's why some of us rodentia end up in medical with well over 300 blunt damage. Still, this looks much more reasonable - still scary with a much smaller chance of accidental round removal (two traps currently gib tiny rodentia).

Alright, I also changed the base damage to 1 since that's the case.

PR #602, opened 2025-03-06
Weapon Buff

fenndragon:
# Description

Part 1 of a full rework of combat to make stamina less punishing, and double the value of all sources of damage and mitigation as requested by the heads. 

https://github.com/fenndragon/floofstation1.git Refer to this for the expected result of these 3 PR's 

---

# Changelog

:cl:

- tweak: Tweaked Weapon values and stamina cost


VividPups:
Wouldn't structural damage increase. just make it easier for people to just break into things

fenndragon:
> Wouldn't structural damage increase. just make it easier for people to just break into things

It would. It's actually a change I was considering reverting. But for the sake of the test I wanted to start with all values at double to make sure we had a good control group understanding of what was going on

fenndragon:
alright finished doing the changes and fixing up all the weapons so stamina cost isn't a thing really unless we're dealing with a stun weapon. 

the structure values have been returned to normal, this should be ready for testing

ghost:
closing for now

PR #605, opened 2025-03-06
IPC Dead Battery Unique State on Monitor

Forzii:
# Description

IPCs with any suit sensor (jumpsuit or tracking implant) that is on coords or vitals will now display a unique status if their battery is missing or discharged. 

New sprite by Dzordz.

---

# TODO

- [x] Make IPCs have DEAD status on crew monitors when their battery is dead or missing
- [x] Make IPCs instead of a unique status when their battery is dead or missing 

---

<p>

I don't really feel like figuring out a way to record stuff right now x.x

image

image

</p>
</details>

---

# Changelog

:cl: Forzii and Dzordz
- tweak: IPCs now have a unique status on crew monitors if their battery is discharged or missing. Better turn coords or vitals on!


Fansana:
Would it be possible to add an additional status for dead battery? I'm not requiring as a change it but it might be a cool idea. If it#s too hard we can just merge it as it is.

Forzii:
> Would it be possible to add an additional status for dead battery? I'm not requiring as a change it but it might be a cool idea. If it's too hard we can just merge it as it is.

I thought about that, that was actually my original plan. It would be a bit more complicated - I'd have to make changes in the SuitSensorStatus definition to add a new field and then define the logic for displaying the new status in CrewMonitoringWindow... plus we'd need a new sprite for the status message. And... hopefully that's all we'd need to do, not positive. Plus it technically shouldn't work if suit sensors are set to Binary, since by definition binary can only have two states. 

I wanted to avoid dealing with the sprite and minimize changes in base game files, but it's definitely doable. If desired I just need some help with the sprite, cause I don't want to it :(

Content.Client/Medical/CrewMonitoring/CrewMonitoringWindow.xaml.cs
```diff
@@ -209,6 +209,10 @@ private void PopulateDepartmentList(IEnumerable<SuitSensorStatus> departmentSens
             {
                 specifier = new SpriteSpecifier.Rsi(new ResPath("Interface/Alerts/human_crew_monitoring.rsi"), "dead");
             }
+            
+            // Floofstation - unique state for IPC
+            else if (sensor.IsDischarged)
```
Mnemotechnician, line 214: No blank lines between block-less `if` and `else`, please.
```suggestion
            else if (sensor.IsDischarged) // Floofstation - unique state for IPC
```

Content.Server/Medical/SuitSensors/SuitSensorSystem.cs
```diff
@@ -339,6 +340,11 @@ public void SetSensor(EntityUid uid, SuitSensorMode mode, EntityUid? userUid = n
         var isAlive = false;
         if (EntityManager.TryGetComponent(sensor.User.Value, out MobStateComponent? mobState))
             isAlive = !_mobStateSystem.IsDead(sensor.User.Value, mobState);
+        
+        // Floofstation - get IPC battery status
+        var isDischarged = false;
+        if (EntityManager.TryGetComponent(sensor.User.Value, out SiliconComponent? SiliconComp) && (SiliconComp.ChargeState == 0))
+            isDischarged = true;
```
Mnemotechnician, line 347: No need for an if expression, and use EntitySystem.TryComp instead of EntityManager.TryGetComponent:
```suggestion
        var isDischarged = TryComp(sensor.User.Value, out SiliconComponent? SiliconComp) && (SiliconComp.ChargeState == 0);
```

Mnemotechnician (changes requested):
Some minor code style issues, besides that looks good.

Content.Client/Medical/CrewMonitoring/CrewMonitoringWindow.xaml.cs
Forzii: Ah, yeah. That makes sense. I originally was going to use a block statement like the if but it bothered me to do a block for one line of code so I switched away from it.

Content.Server/Medical/SuitSensors/SuitSensorSystem.cs
Forzii: Thanks, that's much cleaner. What's the benefit of using TryComp over TryGetComponent? Is it just that it's native since this is an EntitySystem? I was mostly just referencing the isAlive check above it.

Mnemotechnician: It's just more concise, it still calls EntMan.TryGetComponent under the hood

Mnemotechnician (approved):

FoxxoTrystan (approved):

PR #611, opened 2025-03-10
Add Tiny Trait

Forzii:
# Description

This adds a Tiny trait - basically the micro trait requested. I'd love feedback on the changes and, in particular, any upsides, downsides, or tweaks that should be done for the trait.

Also, I had to tweak the carrying system to allow for people to be picked up with one hand properly.

Currently the trait:
Is restricted from Borg and MedicalBorg jobs (does that actually matter?)
Requires height 158 or lower and width 34 or lower (Oni minimums - and will still reduce to smaller than any race's current smallest size)
Disallows Lightweight and Small
Disallows Oni (component conflicts)
Reduces weight by half (factor 0.5 FixtureDensityModifier min 100kg/m2)
Makes player a PseudoItem of size Normal (Fits in inventories as a 2x2)
Tweaks Carriable to only require 1 hand and reduces base PickupDuration to 2 (default 3)
Reduces visual size by half (scale 0.5, 0.5)
Makes prying doors and such take 40% longer
Reduces melee and stamina damage done by 20%

---

# TODO

- [x] Make a size modification trait function
- [x] Make picking up with one hand work properly
- [x] Create prototype Tiny trait
- [x] Set inverse trait requirements (small, lightweight, martial Artist, Vigor, bionic arm, etc? If I can do proper damage/prying debuffs I probably don't need to lock out too many traits since they'll still be weaker)
- [x] Use the components/systems from Oni refactor to increase time it takes to pry stuff, reduce melee damage? (If I do this, maybe make a different Tiny trait for Oni specifically since their damage numbers are different and I don't want to arbitrarily overwrite them - also they have modified prying too... and probably need to inverse-lock their damage specialization traits)
- [x] Finalize trait (numbers, trait points, size requirement, species requirement, additional changes/downsides, etc)

---

<p>

TBD

</p>
</details>

---

# Changelog

:cl:
- add: Adds a tiny trait! Now you can live out your small dreams.


Mnemotechnician:
Note that you can probably invoke the [height adjust system](https://github.com/Fansana/floofstation1/blob/master/Content.Shared/HeightAdjust/HeightAdjustSystem.cs) instead of messing with visuals and fixtures yourself.

Forzii:
> Note that you can probably invoke the [height adjust system](https://github.com/Fansana/floofstation1/blob/master/Content.Shared/HeightAdjust/HeightAdjustSystem.cs) instead of messing with visuals and fixtures yourself.
> 
> And regarding FixtureDensityModifier, density is expressed in kilograms per square meter.

Good idea, but it looks like the SharedHumanoidAppearanceSystem.SetScale that ultimately gets called does not allow the size to go lower than the minimum for the species (which is directly counter to what I'm trying to do)

Forzii:
Alright, opening this up for review. It functions as-is and I'm planning on getting feedback after it gets released (if it gets approved) and possibly making further tweaks. As always, open to feedback and tweaking this PR in the meantime.

Mnemotechnician (approved):
Looks good

Forzii:
> How's the progress on the last to-do?

I haven't gotten any feedback so, I guess done? I can always tweak it more once we have people starting to use it and get feedback.

Memeji (approved):
Awesome! looks good to go then.

PR #621, opened 2025-03-13
QSI Fixes

LucasTheDrgn:
# Description

The QSI fixes wizden doesn't want you to have

For my original PR:

Adds logic to SwapTeleporterSystem to:

- Cancel if it's clientside (that is, cancel prediction) and either party is outside of the visual range (fixes the Flesh Hallucination bug)
- Stop pulling and being pulled on both sides (fixes Teleporter Accidents allowing for infiltration)
- Cancel the teleport entirely if either side is on a Paused map (Prevents being sent to polymorph/cryo hell)

---

# TODO

- [ ] Fix a few incongruous lines between the wizden version and what needs to happen for the EE version

---

# Changelog

:cl:
- fix: QSI no longer causes nearby players to hallucinate, no longer causes teleporter accidents, and no longer can be used to send people to hell.

### DIFF
--- a/Content.Shared/Teleportation/Systems/SwapTeleporterSystem.cs
+++ b/Content.Shared/Teleportation/Systems/SwapTeleporterSystem.cs
@@ -155,6 +155,13 @@ public sealed class SwapTeleporterSystem : EntitySystem
             return;
         }
 
+        // can't predict if either entity doesn't exist on the client / is outside of PVS
+        if (_netMan.IsClient)
+        {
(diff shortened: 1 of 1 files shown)


LucasTheDrgn:
Oops, found some problems that need to be fixed before this gets pulled.

LucasTheDrgn:
Looking more into it I think I have to redo this whole fix, there's some Stuff going on. Might need to just wait until the migration.

PR #641, opened 2025-03-20
Examine Detail Range Increase

Tk420634:
# Description
Ups the examine details visibility range from 3 to 16.

Would be nice if i knew how to make it not auto-close the window as well, but it's a step in the right direction.  

image

# Changelog

🆑 Fenny
- tweak: Examine range to 16 tiles so you no longer have to walk up to people and oogle them from point blank range to read about their heaving cleavage.

### DIFF
--- a/Content.Shared/Examine/ExamineSystemShared.cs
+++ b/Content.Shared/Examine/ExamineSystemShared.cs
@@ -40,7 +40,7 @@ namespace Content.Shared.Examine
         public const float DeadExamineRange = 0.75f;
 
         public const float ExamineRange = 16f;
-        protected const float ExamineDetailsRange = 3f;
+        protected const float ExamineDetailsRange = 16f; /// Floof Change, TK420634 wuz here
 
         protected const float ExamineBlurrinessMult = 2.5f;
 


Memeji (approved):
From discussion, I like the idea as it means you can see more info regarding consent without being right next to the person.

Tk420634:
> From discussion, I like the idea as it means you can see more info regarding consent without being right next to the person.

It's pretty bog standard.  Traditionally over in SS13 this window also doesn't close when they walk away, neither does what you guys call your consent menu (which is character specific in that space.)  This lets you catch a glimpse of someone, and if you're even remotely quick you can get their Flavor Text (Description) and OOC notes (Consent Info) open while they go on about their day.  Letting you scout out people you want to potentially scene with.

That's not even including stuff like the Character Directory, which I have had bad experiences personally.  It creates a far too picky environment at a certain point and people start only interacting for ERP and that, obviously, can shut the game down at lower pop levels when you pop from person to person and can't find someone that meets EXACTLY your wants.

southbridge-fur:
Your change log is formatted incorrectly, you need it to start with `- tweak:`. Also put your username after the 🆑 emoji

Tk420634:
> Your change log is formatted incorrectly, you need it to start with `- tweak:`. Also put your username after the 🆑 emoji

Sure!  Gimme two seconds lmao

Mnemotechnician (approved):

PR #655, opened 2025-03-22
Floofstation V18 Release

Memeji:
# Description

Floofstation V18 Release

This release contains:
#550 
#575 
#608 
#614 
#617 
#623 
#629 
#630 
#635 
#636 
#637 
#638 
#639 
#641 
#642 
#644 
#645 
#647 
#649 
#650 
#651 
#652 
#653 
#654 
#658 
#659 
#661 
#662 
#663 
#664 
#668 
#669 
#670 
#671 
#673

PR #658, opened 2025-03-22
Tiny Trait Rework

Forzii:
# Description

Tiny trait overhaul! 2.0! Rework! Much improved, probably! I had to make changes to some upstream systems in order to bypass restrictions on size and allow them to work properly when called more than once on the same entity.

- A new way of properly shrinking people is being used! The visual glitches should be gone, weight will be adjusted more accurately, and you'll ACTUALLY shrink relative to your character's size (before it wasn't working as intended, shrinking EVERYONE to 50% of average human/humanoid size).
- Size adjustment has been reduced to 30% (actually putting you pretty close to the size everyone was before this fix)
- Description has been updated, and it's made more clear that this isn't shown on the character creator (like lightweight)
- Tiny people take 10% more brute damage (blunt/slash/pierce), have 20% reduced stamina, and 20% reduced crit/death thresholds (should be 80/180 instead of 100/200 by default - at least for standard races)
- Size maximums have been reduced to 140/29 (8/1 above the minimum size for humans/standard humanoids)
- It now takes 60% longer to struggle out of bags for tiny folks (please don't abuse the bugs/exploits that let you get out instantly)
- Now shows a unique examine message (really on the fence with this one. It adds visual clutter to the examine screen, but I think it's a fun message that can help clarify when someone can't tell if someone is small - minimum size small race - or actually tiny)
- No longer locks out lightweight, as there is no conflict anymore
- Changed point cost to 0. Unlike most others, this trait is built to be somewhat self-balancing.

Note: TECHNICALLY some of the modifiers in the localized description, and the changes above, aren't entirely accurate. The characters have 20 reduced stamina (20% for standard races), 20 reduced crit threshold (20% for standard races) and 20 reduced death threshold (10% for standard races since they die at 200 damage, 20% if you only take the difference from that and the crit threshold - but effectively giving them the same amount of health after crit, 100 for standard races, as normal)

---

# TODO

- [x] Fix the visual bugs, scaling mistakes, etc
- [x] Make it take longer for tiny folks to struggle out of inventories (request from Trystan)
- [x] Reduce size scale to 70% (30% smaller) instead of 50%
- [x] Make tiny folks more fragile
- [x] Reduce size maximums (related to the reduction in size scale)
- [x] Add examine message to tiny peoples
- [x] Remove lightweight/tiny cross-deny
- [x] Re-review changes to look for typos, code simplification, missed cleanup, missing comments, etc
- [x] Get final feedback!

---

<p>

image
image

</p>
</details>

---

# Changelog

:cl:
- tweak: Tiny trait overhaul! Lowered size maximums, changed size scaling, increased fragility, added examine text, and much more!
- fix: Tiny now scales properly and the visual bugs should be squashed!


Memeji (approved):

Resources/Prototypes/Floof/Traits/physical.yml
```diff
@@ -138,25 +133,36 @@
...
\ No newline at end of file
+      pickupDuration: 2 # Default is 3, plus they will already be lighter. Get scooped.
+    - type: CanEscapeInventory
+      baseResistTime: 8 # Default is 5, takes 60% longer when tiny TODO: fix players able to drag themselves out of bags instantly!
+    - type: CritModifier
+      critThresholdModifier: -20 # reduce health to crit by 20 (20% for most standard races)
+    - type: DeadModifier
+      deadThresholdModifier: -20 # reduce health to dead by 20 to align with the crit modifier. This effectively means they have the same health after being crit as someone without the trait.
+    - type: StaminaCritModifier 
+      critThresholdModifier: -20 # reduce stamina by 20 (20% for standard races)
```
Mnemotechnician, line 160: I **think** those 3 will conflict with threshold modifier traits (we have like 2-4 of them and will have one more in the future) and stamina threshold modifier traits (we have 2 from EE and one from floof).

I'm not actually sure if it will conflict, because if TraitReplaceComponent is applied on MapInit, and XModifier systems operate on MapInit as well, then applying the same component several times **could** result in multiplicative effect (since TraitAddComponent would effectively remove and re-add these component) - needs testing.

Content.Server/FloofStation/Traits/TraitSystem.Functions.cs
```diff
@@ -1,25 +1,20 @@
...
-using Robust.Server.GameObjects;
-using System.Numerics;
+using Content.Shared.HeightAdjust;
 
 namespace Content.Server.FloofStation.Traits;
 
-// Scales/modifies the visual size of the character. DOES NOT MODIFY DENSITY/WEIGHT. Use in conjunction with FixtureDensityModifier component.
-public sealed partial class TraitScaleVisualSize : TraitFunction 
+// Scales/modifies the size of the character using the Floofstation modified heightAdjustSystem function SetScale
+public sealed partial class TraitSetScale : TraitFunction
```
Mnemotechnician, line 8: It would be preferable to leave the original function intact and implement the new one as a separate trait function, but I guess it's not a big deal so long as the change's properly commented.

Mnemotechnician (commented):
Non-critical issues. We can still merge it since V18 is nigh, but would be preferrable to resolve those at some point.

Mnemotechnician (approved):
Gonna approve it and leave those issues to be fixed later.

Resources/Prototypes/Floof/Traits/physical.yml
Forzii: Nope! I tested these and they work fine! The way those systems work is by making a one-time modification to a different component on component startup - so there's no conflict. I was actually building a new system/component and trashed it when I realized there was no conflict.

Content.Server/FloofStation/Traits/TraitSystem.Functions.cs
Forzii: Ah - you want me to keep the old function? I'm the one who made the old one and it's uh, a bit buggy. You're the one who told me I wasn't supposed to use the scalevisualsystem!

Resources/Prototypes/Floof/Traits/physical.yml
Forzii: Actually, I will double-check that all the traits that use those are using replacecomponent... otherwise we might have a race condition

Edit for reference: https://github.com/Fansana/floofstation1/blob/master/Content.Shared/Traits/Assorted/Systems/TraitStatModifierSystem.cs

Edit2: checked all the current traits and the use replacecomponent so we should be good!

Mnemotechnician:
Resolved the aforementioned issues already, so going to go ahead and merge for #655

PR #677, opened 2025-03-27
Tiny Trait Hotfix 2

Forzii:
# Description

Height/width was swapped ALSO apparently I made a bad assumption about the default height/width of the HumanoidAppearanceComponent (or something else I dunno this is a HOTFIX).

Fixed the width/height swap, tweaked the size maximums (they were somewhat based off bad scaling apparently), and reverted to exactly old behavior for HeightAdjustSystem.SetScale when the new restricted flag isn't set. 

THIS SHOULD WORK HOPEFULLY... I'm tired and need to sleep soon and want to make sure the fix is ready asap... it seems to be working properly now in testing...

---

# TODO

- [x] Fix things!!
- [x] Re-validate to make sure I didn't screw anything else up, maybe see if there's a more elegant solution to the fix

---

<p>

image

</p>
</details>

---

# Changelog

:cl:
- fix: Scaling definitely wasn't messed up - it was all a dream

### DIFF
--- a/Content.Shared/HeightAdjust/HeightAdjustSystem.cs
+++ b/Content.Shared/HeightAdjust/HeightAdjustSystem.cs
@@ -58,12 +58,8 @@ public sealed class HeightAdjustSystem : EntitySystem
         else
             succeeded = false;
         
-        if (restricted && EntityManager.HasComponent<HumanoidAppearanceComponent>(uid)) // Floofstation - if restricted is true assume unmodified behavior
-            _appearance.SetScale(uid, scale);
-        else if (EntityManager.HasComponent<HumanoidAppearanceComponent>(uid)) // Floofstation - if restricted is false use adjusted scale and passthrough restricted flag
-            _appearance.SetScale(uid, adjScale, restricted: restricted); 
-        else
-            succeeded = false;
+        // Floofstation - removed EntityManager.HasComponent<HumanoidAppearanceComponent>(uid) gate due to addition of EnsureComponent earlier
+        _appearance.SetScale(uid, restricted ? scale : adjScale, restricted: restricted); // Floofstation - if restricted is true fallback to default scaling, otherwise use adjusted scaling based on current size
 
         RaiseLocalEvent(uid, new HeightAdjustedEvent { NewScale = scale });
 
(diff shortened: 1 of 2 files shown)


Content.Shared/HeightAdjust/HeightAdjustSystem.cs
```diff
@@ -58,8 +58,10 @@ public sealed class HeightAdjustSystem : EntitySystem
         else
             succeeded = false;
         
-        if (EntityManager.HasComponent<HumanoidAppearanceComponent>(uid))
-            _appearance.SetScale(uid, adjScale, restricted: restricted); // Floofstation - added restricted flag and switched to using adjusted scale
+        if (restricted && EntityManager.HasComponent<HumanoidAppearanceComponent>(uid)) // Floofstation - if restricted is true assume unmodified behavior
+            _appearance.SetScale(uid, scale);
+        else if (EntityManager.HasComponent<HumanoidAppearanceComponent>(uid)) // Floofstation - if restricted is false use adjusted scale and passthrough restricted flag
+            _appearance.SetScale(uid, adjScale, restricted: restricted);
```
Mnemotechnician:
```suggestion
        if (HasComp<HumanoidAppearanceComponent>(uid)) // Floofstation - if restricted is true assume unmodified behavior, otherwise use initial scale
            _appearance.SetScale(uid, restricted ? scale : adjScale, restricted: restricted);
```

Mnemotechnician (approved):
Should be good given the change itself works. Here's a cleaner syntax though:

Content.Shared/HeightAdjust/HeightAdjustSystem.cs
Forzii: The more I look at the code the more confused I am that my original solution didn't work. appearance.Height and appearance.Width should both default to 1f, so setting the scale via multiplication should have worked (if I hadn't swapped height/width). Something else must also potentially modify those values before it makes it to the heightadjustsystem for the first time. Oh well.

Anyway I realized the HasComponent check isn't actually needed anymore since the code I added does an EnsureComponent soooo - I'ma take your suggestion and fold it into a bit of simplification.

PR #685, opened 2025-03-28
Underwear and Undershirt Expansion

Tk420634:
What if I just ported almost all the undershirts and underwear from Coyote Bayou in one pass?

Do note, included only one set of Shapely Undershirts (there are like 5?), which I suggest players use with a Nudity Permit.  Long term they should layer OVER your uniform.  The idea is that the player then color matches them to the outfit and BAM.  Big Honking Gazongas on whatever outfit you want to have big honking gazongas on.

I know that alternatively we can do something with maps or something, Trystan keeps saying the word like I know what it means.  We could always pull them at that point.

Share and Enjoy.

https://github.com/user-attachments/assets/7b228170-2cf0-41a3-b419-f5135e55268f

:cl: Fenny
- add: Over 100 Undershirts & Underwear (total) from SS13, including Shapely Shirts from Coyote Bayou


Memeji (approved):
Love to see it

Forzii:
Looks like a whole lot of png/RSI without any copyright/license changes in the meta file - are any of these items covered under different licenses or need to be properly credited to artists?

Tk420634:
> Looks like a whole lot of png/RSI without any metadata - are any of these items covered under different licenses or need to be properly credited to artists?

All of them should be the standard /tg/ license from CB.  We had very few sprites that required crediting, mostly just stuff from Mojave Sun or a few outfits from Eris, none of which is in this.  All of these came from the code before we put those in, with the exception being the shapelies.  Which I added as the standard license.  Generally speaking with SS13 stuff you can USUALLY assume it's free to use/remix.  There are very few codebases that use other licenses, mostly because no one wants to deal with them, and the way they usually do is adding a text file for crediting wholesale.  They are still very open source wild west over there.

I didn't think to check what the meta datas license was set to though, so I may need to poke it with a stick.

Tk420634:
> > Looks like a whole lot of png/RSI without any metadata - are any of these items covered under different licenses or need to be properly credited to artists?
> 
> All of them should be the standard /tg/ license from CB. We had very few sprites that required crediting, mostly just stuff from Mojave Sun or a few outfits from Eris, none of which is in this. All of these came from the code before we put those in, with the exception being the shapelies. Which I added as the standard license. Generally speaking with SS13 stuff you can USUALLY assume it's free to use/remix. There are very few codebases that use other licenses, mostly because no one wants to deal with them, and the way they usually do is adding a text file for crediting wholesale. They are still very open source wild west over there.
> 
> I didn't think to check what the meta datas license was set to though, so I may need to poke it with a stick.

Yeah, I need to hit the meta with a stick, will in a bit.  Wanna play for another round lol.  Thank you Forzii!

Tk420634:
> Love to see it

There ya go, beastie.  Ready when you can find a second to tag team it.

ShadesMars (approved):

PR #699, opened 2025-03-30
Examine Tooltip for Players Persists Thru Walls

Superlagg:
# Description

When you examine someone, the tooltip should still show up even if they hide behind a wall. Still obeys max distance though, so it'll disappear if they run too far away.

Whole point of this is to make it a bit easier to read peoples' flavortexts.

---

# Changelog

:cl:
- tweak: Player examine tooltips will persist even if they go out of line of sight.

### DIFF
--- a/Content.Shared/Examine/ExamineSystemShared.cs
+++ b/Content.Shared/Examine/ExamineSystemShared.cs
@@ -112,7 +112,10 @@ namespace Content.Shared.Examine
 
             // Floofstation edit - if the examined thing is a player, just return true
             if (examined != null)
-                if (EntityManager.TryGetComponent<ActorComponent>(examined, out var _))
+                if (EntityManager.TryGetComponent<ActorComponent>(examined, out var _)
+                    && examined.HasValue
+                    && Transform(examiner).Coordinates.TryDistance(EntityManager, _transform, Transform(examined.Value).Coordinates, out var distance)
+                    && distance <= GetExaminerRange(examiner))
                     return true;
 
             return InRangeUnOccluded(
--- a/Content.Shared/Examine/ExamineSystemShared.cs
+++ b/Content.Shared/Examine/ExamineSystemShared.cs
@@ -326,6 +326,9 @@ namespace Content.Shared.Examine
             _hasDescription = hasDescription;
         }
 
+        /// <summary>
+        ///     Returns <see cref="Message"/> with all <see cref="Parts"/> appended according to their priority.
+        /// </summary>
         public FormattedMessage GetTotalMessage()
         {
             int Comparison(ExamineMessagePart a, ExamineMessagePart b)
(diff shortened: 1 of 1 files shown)


Content.Shared/Examine/ExamineSystemShared.cs
```diff
@@ -109,6 +110,11 @@ public virtual bool CanExamine(EntityUid examiner, MapCoordinates target, Ignore
             if (EntityManager.GetComponent<TransformComponent>(examiner).MapID != target.MapId)
                 return false;
 
+            // if the examined thing is a player, just return true
```
Mnemotechnician:
```suggestion
            // Floofstation edit - if the examined thing is a player, just return true
```

Content.Shared/Examine/ExamineSystemShared.cs
```diff
@@ -320,9 +326,6 @@ public ExaminedEvent(FormattedMessage message, EntityUid examined, EntityUid exa
             _hasDescription = hasDescription;
         }
 
-        /// <summary>
-        ///     Returns <see cref="Message"/> with all <see cref="Parts"/> appended according to their priority.
-        /// </summary>
```
Mnemotechnician: Why is this getting removed?

Mnemotechnician (changes requested):

Mnemotechnician:
Also, I do not see a range check here. You are skipping the entire InRangeUnoccluded call, but not providing a fallback range check.

Content.Shared/Examine/ExamineSystemShared.cs
Superlagg: Was a mistake, putting it back!

Superlagg:
> Also, I do not see a range check here. You are skipping the entire InRangeUnoccluded call, but not providing a fallback range check.

That's what I thought, but in testing, the tooltip still disappeared after moving a screen away from the dummy I gave an actor component. Not sure why that is, though.

Mnemotechnician:
I would prefer not to rely on undefined behavior and add an explicit range check here. You can achieve it by adding a simple `&& Tranform(examiner).TryDistance(EntityManager, _tranform, Tranform(target).Coordinates, out var distance) <= GetExaminerRange(examiner)` clause to your `if` statement's condition, or otherwise.

Mnemotechnician (approved):

Content.Shared/Examine/ExamineSystemShared.cs
```diff
@@ -109,6 +110,14 @@ public virtual bool CanExamine(EntityUid examiner, MapCoordinates target, Ignore
             if (EntityManager.GetComponent<TransformComponent>(examiner).MapID != target.MapId)
                 return false;
 
+            // Floofstation edit - if the examined thing is a player, just return true
+            if (examined != null)
+                if (EntityManager.TryGetComponent<ActorComponent>(examined, out var _)
```
Mnemotechnician:
```suggestion
                if (TryComp<ActorComponent>(examined, out var _)
```

Superlagg: Sure thing! I assume TryComp just checks if the component exists, while TryGetComponent also fetches the actual component?

Mnemotechnician: It is the same thing, but in the more preferable way

Memeji (approved):

PR #703, opened 2025-03-31
Add Heavyweight Trait

Forzii:
# Description

Adds a heavyweight trait as requested by players. It's effectively the opposite of lightweight. I made them mutually exclusive and also locked out harpies, since birds are explicitly lightweight and less dense. It's like, their whole thing.

As always, feedback welcomed.

---

# TODO

- [x] Add heavyweight trait

---

<p>

image

</p>
</details>

---

# Changelog

:cl:
- add: Added a heavyweight trait. Some of ya'll be dense.


Tk420634:
BIG AND ROUND!

Mnemotechnician (approved):

Mnemotechnician (approved):

Mnemotechnician:
...oh, didn't realize I was the one approving it before. Bleh.

Resources/Prototypes/Floof/Traits/physical.yml
```diff
@@ -141,4 +145,26 @@
...
+  requirements:
+  - !type:CharacterJobRequirement
+    inverted: true
+    jobs:
+    - Borg
+    - MedicalBorg
+  - !type:CharacterSpeciesRequirement 
+    inverted: true
+    species:
+    - Harpy # Doesn't really make sense with their wings and flight ability
```
M3739, line 225:
```suggestion
    - Harpy # Doesn't really make sense with their wings and flight ability
    - Diona # Trees are heavy as it is.
    - Chitinid # Ants got weight.
```
You may want to consider adding `Diona` and `Chitinid`, since they are both species that feature higher fixture density as part of their unique traits.

For reference, the `BaseMobSpeciesOrganic` has the fixture density set to `185`. Chitinids have this set to `220` instead, while Diona have a whopping `300` as their density.

Recall that both Rodentia and Felinids are currently restricted from the lightweight trait due to their already less than normal density. IPCs have the standard density. 

**Whether or not you want to restrict IPCs as well is up to your discretion.**

Forzii: Makes sense to me! An IPC made out of denser materials seems reasonable to me, so I'm not going to add them currently. I'm actually a little surprised IPCs are standard density.

ShadesMars (approved):

PR #705, opened 2025-04-01
Implant Trackers Rework

VividPups:
# Description

Trackers are 1k for like 5, and sooo many times EVERY round, people would buy trackers at the round start to the point it feels like it is Min-maxing for security and command, to which people that do over time cause the HOS and security to become less in quality.  
This PR changes the Trackers to be a T2 research item for medical. This would help push cooperation between  Medical and Epi with this change. This will also allow Medical to hand out and give out Implants to crew to help with RR 
# Medical 
image
image
image

# Changelog

:cl:
- tweak: Tracker implant are now a T2 researchable under civil


Forzii:
1,900% increase is wild.

VividPups:
> 1,900% increase is wild.
 
 even a shitty cargo can get that amount within an hour

LucasTheDrgn:
Honestly they're kinda cooking with this though, at 1k they might as well be roundstart for sec and command (if it's that cheap and a one-off purchase, why isn't NT outfitting command with them from home base?), but since they're not roundstart they should take _some_ effort to get. Maybe also make them researchable as an alternative acquisition method?

Mnemotechnician:
> even a shitty cargo can get that amount within an hour

20k for something so basic is a bit *too* much. Cargo shouldn't be working for an hour just to buy security a single crate.

VividPups:
> > even a shitty cargo can get that amount within an hour
> 
> 20k for something so basic is a bit _too_ much. Cargo shouldn't be working for an hour just to buy security a single crate.

Except it isn't basic, and even bad cargo techs can get enough to buy it. the problem is, Security and command will just ask for it within 10 minutes because they are too cheap. Even 10k is extremely cheap. after which it makes doing some of the obj's not worth it because of the tracker

Mnemotechnician:
We shouldn't force cargo to powergame even more than they already do in order to buy what is required from them.

VividPups:
It isn't powergaming to raise the price. It is however Highly abused that security and command just order and buy trackers within ten mins of the round because the price is so cheap. Thus making the game extremely more difficult for any sort of kill obj. Raising the price would make it where it would take long for them to buy it to help even the curve between round start buying and the 30 mins grace period.

VividPups:
> We shouldn't force cargo to powergame even more than they already do in order to buy what is required from them.

would the maintainers be more open in Removing the trackers from cargo and instead put it as an researchable, that is a t2 or t1 item with epi? then it would be a bit RNG based that can help either Sec or Antag depending on the Round?

VividPups:
oops

VividPups:
would a better Alt for this be that it is a researchable instead of a buyable? and a possible loot from an expedition?

Memeji:
I'm fine with either solution. Maybe not as loot?
But making it researchable could work.

VividPups:
> I'm fine with either solution. Maybe not as loot? But making it researchable could work.

I wouldn't mind changing it to that but I would like to Get the feedback from the other Maintainers first

ShadesMars (approved):

VividPups:
Hi it's been two weeks since an maintainer was comment on this PR. Could I get an Update on the trackers

Memeji:
As said, fine with it being researchable.
@Mnemotechnician @FoxxoTrystan @Fansana 
Thought from yall?

ghost (approved):

ShadesMars (approved):

VividPups:
looks like the last person that needs to check is @Mnemotechnician

Resources/Prototypes/Entities/Objects/Misc/subdermal_implants.yml
```diff
@@ -82,6 +82,9 @@
   description: This implant has a tracking device attached to the suit sensor network, as well as a condition monitor for the Security radio channel.
   categories: [ HideSpawnMenu ]
   components:
+    - type: Sprite  #Floof change
+      sprite: Objects/Specific/Medical/implanter.rsi
+      state: implanter1
```
Mnemotechnician: Why is this here? Subdermal implants should have no sprites unless you know what you are doing.

Resources/Prototypes/Entities/Objects/Misc/subdermal_implants.yml
```diff
@@ -101,7 +104,7 @@
       mobState:
       - Critical
     - type: Rattle
-      radioChannel: "Security"
+      radioChannel: "Security" # Floof change medical should know when a mass cass happen
```
Mnemotechnician: I'd agree with this change, *if* it was here... Makes tracking implants more useful, too.

Resources/Prototypes/_Floof/Research/civilianservices.yml
```diff
@@ -14,6 +14,20 @@
   technologyPrerequisites:
   - Hydroponics
 
+- type: technology
+  id: AdvancedTracking
+  name: research-technology-mini-hypo
```
Mnemotechnician: Advanced tracking technology is named "advanced medical injector"? Probably missing an ftl entry.

Resources/Prototypes/_Floof/Research/civilianservices.yml
```diff
@@ -14,6 +14,20 @@
...
+  icon:
+   sprite: Objects/Specific/Medical/implanter.rsi
+   state: implanter0
+  discipline: CivilianServices
+  tier: 2
+  cost: 15000
```
Mnemotechnician: I think 15k for a technology that hardly has any use for anyone is a bit overkill. Such technologies are typically made to cost 5k or 7.5k.

Mnemotechnician (changes requested):
Other than that should be good.

Resources/Prototypes/Entities/Objects/Misc/subdermal_implants.yml
VividPups: it was because when I did it on the Parent it wasnt working when i printed

Resources/Prototypes/Entities/Objects/Misc/subdermal_implants.yml
VividPups: Sadly, it doesn't support multiple channels. My other idea was to make a Medical tracker and a security tracker, I would rather have support for MULTIPLE channels, but I would need help with some changes in the code itself

Resources/Prototypes/_Floof/Research/civilianservices.yml
VividPups: it was yea Ill fix it soon

Resources/Prototypes/Entities/Objects/Misc/subdermal_implants.yml
Mnemotechnician: You aren't supposed to add a sprite here. This entity is invisible without admin interference (or some obscure bugs)

Resources/Prototypes/Entities/Objects/Misc/subdermal_implants.yml
Mnemotechnician: I mean it'd make more sense for it to alert medical is med is the ones who can produce trackers.

VividPups: was chanage to medical

Resources/Prototypes/Entities/Objects/Misc/subdermal_implants.yml
VividPups: fix

Mnemotechnician (approved):
Okay. you still don't need either of those sprite components on the implanter and on the implant, but I'm just going to go ahead and approve it since I don't want to bother. Everything else seems fine.

Memeji:
> Okay. you still don't need either of those sprite components on the implanter and on the implant, but I'm just going to go ahead and approve it since I don't want to bother. Everything else seems fine.

What would need to be changed? Just to remove the components?
I'd rather resolve this here just in case, if that is ok.
If it is a nothing-burger, we can skip it. I just wanna be sure.

PR #707, opened 2025-04-01
Adds a Couple Traits that Let Dogs (and Such) Eat Chocolate / Onions

Superlagg:
# Description

Say you want to be a pretty dog, but oh no, Vulpkanin can't eat chocolate! That'd good and all, if your pretty dog character was actually canonically a Vulpkanin. Maybe you're some kind of space wolf that can eat chocolate all day, or you feel its more in character that chocolate doesn't poison your pretty dog. Either way, its a trait that lets you do that.

Also adds in a framework to add in other traits that'll change dietary stuff. Should be easy to add traits that make humans be poisoned by chocolate!

---

# Changelog

:cl:
- add: Added two traits that'll let your dog (or mouse, or cat, or fish) eat theobromine and/or allicin!


Content.Server/Traits/TraitSystem.Functions.cs
```diff
@@ -400,3 +406,75 @@ public override void OnPlayerSpawn(EntityUid uid,
         staminaComponent.Cooldown += CooldownModifier;
     }
 }
+
+// FLOOF START
+[UsedImplicitly]
+public sealed partial class TraitModifyMetabolism : TraitFunction
```
Mnemotechnician: This should be moved to floofstation namespace.

LucasTheDrgn:
Reptilians, Arachnids (potentially Arachne? They seem to have human stomachs but they also have the blood sucker component so I'm not sure they have have animal metabolisms or not), and Resomi should also have these traits available.

Mnemotechnician (changes requested):
Interesting way to implement this, I originally expected this to simply remove animal metabolism from the entity. However, changes to upstream files need to be commented as belonging to floofstation.

Also, this introduces a breaking change by swapping the metabolism type requirement for theobromine/allicin poisoning. While this is not necessarily an issue right now, this will cause problems if upstream (or even someone unaware of this change) decides to add new animal poison metabolizers... I think perhaps we should start writing those down so future contributors don't have to figure it out themselves.

Content.Server/Traits/TraitSystem.Functions.cs
Superlagg: How would I do that?

Superlagg:
> Interesting way to implement this, I originally expected this to simply remove animal metabolism from the entity. However, changes to upstream files need to be commented as belonging to floofstation.
> 
> Also, this introduces a breaking change by swapping the metabolism type requirement for theobromine/allicin poisoning. While this is not necessarily an issue right now, this will cause problems if upstream (or even someone unaware of this change) decides to add new animal poison metabolizers... I think perhaps we should start writing those down so future contributors don't have to figure it out themselves.

Sure, where would I write those down in?

Superlagg:
> Reptilians, Arachnids (potentially Arachne? They seem to have human stomachs but they also have the blood sucker component so I'm not sure they have have animal metabolisms or not), and Resomi should also have these traits available.

Turns out, these chems check against the mob's *heart*, since that's what metabolizes poisons for some reason. Ya know, looking at it, *most* species have an animal heart, which makes them affected by these chems, so I'm just gonna remove the species requirement. Its a positive trait that does nothing for species that are unaffected by these chems, after all!

Content.Server/Traits/TraitSystem.Functions.cs
Mnemotechnician: Just move it to Content.Server/Floofstation/.../... .cs, create the file/namespace if it doesn't exist

Mnemotechnician:
> Sure, where would I write those down in?

RIght now you don't have to, it's just a remark; something I've recently realized we've been lacking.

Superlagg:
> > Sure, where would I write those down in?
> 
> RIght now you don't have to, it's just a remark; something I've recently realized we've been lacking.

Aight, imma make a note to include this whenever that starts to become a thing =3

Mnemotechnician (approved):

Memeji (approved):

PR #752, opened 2025-04-13
Fix Jug Names

ghost:
:cl:
- fix: Fixed jug labels being doubled up.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Specific/chemical-containers.yml
+++ b/Resources/Prototypes/Entities/Objects/Specific/chemical-containers.yml
@@ -85,7 +85,7 @@
 
 - type: entity
   parent: Jug
-  name: jug (carbon)
+  name: jug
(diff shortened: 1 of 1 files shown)


Mnemotechnician:
While this will "fix" the labeling issue, we also need a way to distinguish them in the spawn menu... I also hope the chem dispenser will automatically display the chemicals inside.

ghost:
funny thing actually, those jugs can't be spawned using the spawn menu because of the hideInSpawnmenu thingy. And yes the ui does display the labels.

Mnemotechnician:
> funny thing actually, those jugs can't be spawned using the spawn menu because of the hideInSpawnmenu thingy. And yes the ui does display the labels.

Oh, yea, then it's fine. I just thought of adding suffixes or something, but if they're NoSpawn, no need.

PR #767, opened 2025-04-14
Batch Fixes

Mnemotechnician:
# Description
1. Turns out something got fucked up during the upstream merge. Initially (on wizden impl) MeleeWeaponComponent.AttackRate was expressed in hertz. That was fine. Then EE decided to randomly change it to seconds. We inherited that change. During the last upstream merge the said change got reverted, but not all melee weapons had their attack rates updated. I also mistakingly updated a few weapons in the wrong way - this should now be fixed.

2. The grid preloader was made active in non-debug builds regardless of whether the relevant CVar is enabled. This is because neither the grid preloader system, nor the LoadMap gamerule system (which loneops and unknown shuttle events depend on) handle the case where this cvar is disabled, leading to events being non-functional. Also adds a statement to log an error when the grid preloader fails to fetch a loaded grid.

3. This PR also FINALLY converts tail thumps and crew monitor beeps to mono and lowers their volume (by 5 and 11 db respectively). 

<p>

Melee weapons I tested:

https://github.com/user-attachments/assets/2e09f1b3-6bf1-4d74-8083-cb50c4c0de6b

Loneop:

Uploading simplescreenrecorder-2025-04-14_23.16.51.mp4…

</p>
</details>

# Changelog
:cl:
- fix: Most melee weapons should no longer have their attack rates inverted.
- fix: Loneop and unknown shuttle events should now work correctly.
- fix: Tail thumps and crew monitor alerts now have correct sound positioning and lower volume.


Mnemotechnician:
Okay, seems like (2) didn't fully fix the issue, I'll look further...

Mnemotechnician:
Okay, fixed the loneop part by adding the AntagLoadProfileRule component to it. Yes, all that fuss because of a missing component.

Memeji (approved):

PR #769, opened 2025-04-14
Paintbrush

Memeji:
# Description

Adds a paintbrush that can be used for making art on station.

2x2 items that contains a lot of charges and can be used like the rainbow crayon.

---

# Changelog

:cl:
- add: Added a paintbrush to loadouts and the artistic toolbox and crate.


Mnemotechnician:
I think 1000 charges is a bit too overkill. Players creating too many decals in one place may cause other player's clients to lag.

Memeji:
> I think 1000 charges is a bit too overkill. Players creating too many decals in one place may cause other player's clients to lag.

Was a fear but I didn't see too much of an issue by myself when spamming it.
Would around 500 be good?

PR #772, opened 2025-04-15
Un-Butcher the Translator System

Mnemotechnician:
# Description
Upstream decided to butcher it during the upstream merge, because apparently PowerCellDrawSystem now requires entities not only to have the PowerCellDrawComponent enabled, but also to have an ItemToggleComponent and have it enabled. So dumb and redundant.

In doing so, upstream broke automatic language switching and made the power cell usage not work at all (or at least in some cases) because the translator system handles the toggling of translators on its own.

This comments out changes made by upstream and makes it so that the translator system simply synchronizes ItemToggle.Activated with Translator.Enabled and PowerCellDraw.Enabled, and also fixes a bug in PowerCellSystem that would cause power cell updates to accumulate over time while the item is disabled and come down all at once once it's enabled, resulting in rapid charge loss proportional to the time the item spent inactive.

# Changelog
:cl:
- fix: Translators should now correctly use power when activated.


Mnemotechnician:
https://github.com/user-attachments/assets/b85ab377-fde2-488e-bb37-1828cc0ab5ff

Memeji (approved):

PR #780, opened 2025-04-17
Fix Suits Prices and Re-Add Juggernaut Suit

ghost:
:cl:
- tweak: Brought back the syndicate jug suit with tweaked prices.


Mnemotechnician (approved):

Mnemotechnician:
It would appear there are now two copies of syndie and jug suits now though, gotta comment the old ones first...

PR #781, opened 2025-04-17
Makin Mining Drill Great Again V 4

VividPups:
Description.
Adds the Wideswing again to be able to hit 2 rocks at the same time 

---
# Changelog

:cl:
- tweak: Mining Drill wide swing is now at a 45 degree angle

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Melee/mining.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Melee/mining.yml
@@ -52,7 +52,7 @@
     state: handdrill
   - type: MeleeWeapon
     autoAttack: true
-    angle: 27
+    angle: 27 #Floof
     wideAnimationRotation: -90
     soundHit:
       path: "/Audio/Items/drill_hit.ogg"
@@ -77,7 +77,7 @@
     state: handdrill
   - type: MeleeWeapon
     autoAttack: true
-    angle: 27
+    angle: 27 #floof
     wideAnimationRotation: -90
     soundHit:
       path: "/Audio/Items/drill_hit.ogg"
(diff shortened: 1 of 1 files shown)


Mnemotechnician:
Last time I checked, the mining drill has an attack distance of like... 3-4 tiles? This will make it insanely overpowered

VividPups:
> Last time I checked, the mining drill has an attack distance of like... 3-4 tiles? This will make it insanely overpowered

Never seen it had that far of range

Mnemotechnician:
> Never seen it had that far of range

It did have that much, even if visuals told you otherwise. I personally have discovered and demonstrated it on the floof server, too. But I always considered it a bug, so if you can fix it or prove that upstream has somehow sneakily fixed it, we can allow your change.

(Although I personally still think 45° is a bit too much for a drill - that makes the pickaxe lose its upside. Perhaps something smaller like 22.5° would be better?)

VividPups:
I mean I feel like since the Upstream Salvagers' job is extremely to relacks. I mean after last night me salvage team and did 2 expeditions and the station was completely stocked with all material and the Department its too the point where cargo techs can really fuck off because of how Easy it is to get diamonds. Like the mining drill is one thing...but being able to get like 5 stacks of diamonds in one shift?

Mnemotechnician:
I do agree salvage is too easy, and know people who agree with that, but the community doesn't exactly seem to agree with that, so there isn't much we can do. I'd recommend you to post in #server-issues-and-concerns.

VividPups:
22 is a bit too small, making it difficult to get both rocks in 1 swing. So it seems that 27 degrees is a good sweet spot. I haven't been able to get it 3 rocks in one hit, which is what I am aiming for.  However, about the pickaxe, I feel like the pickaxe itself needs to be reworked to be better in mining or buffed to where it isn't fully unbearable

Memeji (approved):
Seems good from past discussion with the reduced angle. I think being able to hit two can be fine. It can be tweaked with feedback.

Mnemotechnician (approved):

PR #718, opened 2025-04-02
Upstream Merge (1351)

ghost:

PR #804, opened 2025-04-19
Return Sec Watch to Security PDA

WhiteHusky:
PR #718 removed it for some reason

# Changelog

:cl:
- fix: Security PDA has SecWatch back


Mnemotechnician (approved):

Mnemotechnician:
Probably because NanoChat was added and upstream didn't increase the cartridge limit

PR #814, opened 2025-04-19
Upstream Merge #1382

ghost:
Upstream PR: https://github.com/simple-station/einstein-engines/pull/1382
# Description

This PR does *some* refactoring of MobState in preparation for adding a more in-depth SoftCritical system, mostly adding code support for the new threshold existing in the first place. Additionally, the MobStateComponent now also includes several DataFields that allow more precise fine tuning of what all mobs are allowed to do in each state, as opposed to hardcoding a majority of these interactions.

The actual SoftCritical state isn't currently used by anything, and so is a largely vestigial feature. It would have to be added individually to mobs, such as the BaseMobSpeciesOrganic when the time comes to actually add this feature.

# Changelog

:cl: EinsteinEngines
- add: Lots of refactoring of the Mob State Systerm in preparation for implementing a much more detailed Soft Critical state.

### DIFF
--- a/Content.Server/Traits/TraitSystem.Functions.cs
+++ b/Content.Server/Traits/TraitSystem.Functions.cs
@@ -348,6 +348,9 @@ public sealed partial class TraitModifyMobThresholds : TraitFunction
     public int CritThresholdModifier;
 
     [DataField, AlwaysPushInheritance]
+    public int SoftCritThresholdModifier;
+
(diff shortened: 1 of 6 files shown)


Mnemotechnician (approved):
This code is so bad... "Allow interacting while dead". But it's upstream, so not much we can do.

ShadesMars (approved):

PR #839, opened 2025-04-21
ERP Seeking Toggle

Tk420634:
A relatively simple toggle to show that a player is specifically seeking out an ERP partner.

Dan Code in a Fen VSC.

image

---

# Changelog

:cl: FennyAndDan
- add: ERP Seeking Consent Menu Toggle to Alert others that you are IN DA MOOD.

### DIFF
--- a/Content.Server/FloofStation/HornyComponent.cs
+++ b/Content.Server/FloofStation/HornyComponent.cs
@@ -1,0 +1,4 @@
+namespace Content.Server.FloofStation.Horny;
+
+[RegisterComponent]
+public sealed partial class HornyComponent : Component;
--- a/Content.Server/FloofStation/HornySystem.cs
+++ b/Content.Server/FloofStation/HornySystem.cs
@@ -1,0 +1,40 @@
+using Content.Shared.Examine;
+using Content.Shared.IdentityManagement;
+using Content.Server.Consent;
+
+namespace Content.Server.FloofStation.Horny;
+
+public sealed class MindbrokenSystem : EntitySystem
+{
+    [Dependency] private readonly ConsentSystem _consent = default!;
+
+    public override void Initialize()
+    {
+        base.Initialize();
+        SubscribeLocalEvent<HornyComponent, ExaminedEvent>(OnExamined);
+    }
+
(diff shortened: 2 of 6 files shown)


Cprn-Sprkl:
Would it be possible to make the text more like the 'subtle' action colour? It being purple could make it easily missed on characters with the anti-psionic trait.

The other concern would be this making people less likely to approach each other. Maybe something to make clear that this is a green flag, but not a stop sign when toggled off? A second toggle with an "I'm absolutely not looking" kind of statement?

Tk420634:
> Would it be possible to make the text more like the 'subtle' action colour? It being purple could make it easily missed on characters with the anti-psionic trait.
> 
> The other concern would be this making people less likely to approach each other. Maybe something to make clear that this is a green flag, but not a stop sign when toggled off? A second toggle with an "I'm absolutely not looking" kind of statement?

It should be pink in game lol, that was a fairly early screenshot because I'm dumb as hell.

image

I'm not personally sold on the text for this either, if it makes you feel any better.  Very open to better wording!

As for the second switch, we could.  Just want it to be like 'i'm literally only here to play the game?'

Cprn-Sprkl:
The wording seems fine except maybe "Go talk to them!" being a little unnecessary.

If you want to get really in the weeds you could drop any pronouns unless there's a variable that picks up on the pronouns the character chose. Rewording it to "NAME seems to be looking for a good time!" would accomplish that while keeping a decent flow.

The other toggle could maybe show something like "NAME seems to be focused on other things right now."

Tk420634:
> The wording seems fine except maybe "Go talk to them!" being a little unnecessary.
> 
> If you want to get really in the weeds you could drop any pronouns unless there's a variable that picks up on the pronouns the character chose. Rewording it to "NAME seems to be looking for a good time!" would accomplish that while keeping a decent flow.
> 
> The other toggle could maybe show something like "NAME seems to be focused on other things right now."

Yeah, it feels a bit wordy for what I want it to accomplish.  I'll poke at it in just a bit.

Tk420634:
REHASHING INTO SOMETHING ENTIRELY DIFFERENT.  HARD TO EXPLAIN, BUT TRUST.

PR #843, opened 2025-04-23
Fix Saving Prefered Quick Store Item Locations

### LINKED ISSUE #758: Save Item Location bind no longer works
There is a `Save Item Location` bind that when used over an item in inventory saves its location so when it is dropped into that storage later (item clicked upon bag or via keybind) will automatically be placed in the same location. Seems like the recent update broke it.

WhiteHusky:
Upstream cherrypick: space-wizards/space-station-14#32480

Fixes #758

# Changelog

:cl:
- fix: Saving item locations are working again


WhiteHusky:
image
image

Memeji (approved):
Looks good to me.

PR #844, opened 2025-04-23
Arrivals Outpost Tweaks

Mnemotechnician:
# Description
- Unblocks the passageways to the rest of the arrivals outpost
- Removes pretty much all valuable loot (all departments except sec get 1 filled departmental locker, sec gets none)
- Changes all the stuff in the museum to less-valueable stuff (dull sword, cc collars, etc) and replaces the windows with plastitanium ones
- Adds a stub disposal room and reroutes the disposal pipes to it instead of the impenetrable power room.
- Replaces the booze-o-mat with an all-access version because there isn't ever a bartender to manage the arrivals bar.
- Adds a spray paint crate to the dressing room for convenience.

<p>

terminal-0

</p>
</details>

# Changelog
:cl:
- tweak: Updated the arrivals outpost to allow access to the previously blocked-off rooms.

### DIFF
--- a/Resources/Maps/Misc/terminal.yml
+++ b/Resources/Maps/Misc/terminal.yml
@@ -31,25 +31,12 @@ tilemap:
 entities:
 - proto: ""
   entities:
-  - uid: 1
-    components:
-    - type: MetaData
-      name: Map Entity
-    - type: Transform
-    - type: Map
-      mapPaused: True
-    - type: PhysicsMap
-    - type: GridTree
-    - type: MovedGrids
-    - type: Broadphase
-    - type: OccluderTree
-    - type: LoadedMap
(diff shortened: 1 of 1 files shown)


Mnemotechnician:
It would appear I saved it as a map rather than a grid, dunno if it has any implications other than slightly worse performance. Loaded fine on my instance.

southbridge-fur (approved):

southbridge-fur (changes requested):
Please save this as a grid, it can mess up the terrain generation otherwise

Mnemotechnician:
> Please save this as a grid, it can mess up the terrain generation otherwise

Makes sense, didn't notice that since the outpost is in space now

PR #867, opened 2025-04-25
Upstream Merge #1446

ghost:
Upstream PR: https://github.com/simple-station/einstein-engines/pull/1446
# Description

When https://github.com/Simple-Station/Einstein-Engines/pull/720 (PR to raise trait limit from 5 to 10) was merged, we only had **53** traits. Now, we have a total of **109** traits. I was conservative with the trait limit back then because we didn't have as much traits, but now that we have so many we should raise the default trait limit from **10** to **14**. I believe 14 should be as high as it gets from now on.

For reference, the quirk limit in /tg/ station is just a maximum of 6 positive quirks. So a typical build that reaches the 14 trait limit would be something like 6 positive traits, 6 negative traits to afford all of those positive traits, 1 language trait and 1 accent trait. 

# Changelog

:cl: EinsteinEngines Skubman
- tweak: The maximum trait limit has increased from 10 traits to 14 traits.

### DIFF
--- a/Content.Shared/CCVar/CCVars.cs
+++ b/Content.Shared/CCVar/CCVars.cs
@@ -422,7 +422,7 @@ namespace Content.Shared.CCVar
         ///     How many traits a character can have at most.
         /// </summary>
         public static readonly CVarDef<int> GameTraitsMax =
-            CVarDef.Create("game.traits_max", 10, CVar.REPLICATED);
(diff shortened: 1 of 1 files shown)


Mnemotechnician:
Don't think this one's relevant.

Forzii:
> Don't think this one's relevant.

It's not, since we are using the CVAR to overwrite it, but it doesn't hurt to merge the updated default.

PR #883, opened 2025-04-25
Re Adding Time Requirement to the Jobs

VividPups:
# Description

I have seen multiple times now that Command members and even basic department members make Mistakes that cause not only Admins' action but also ruin someone's round. I HAVE ALSO seen multiple times that people in the VC and in the Discord complain that a certain person doesn't know a certain part of their job that is a major part. Ignorance shouldn't be an excuse, especially when you are a command member and don't know how to do certain things. there is a difference between forgetting and not knowing.

For example  CMO should know bacis chems, the CE should know atmos or at the basics of TEG,  LO should know the struggles of salvage. and security should know the law and sops. They shouldn't go senior and then go to dorms early in the round and not teach the cadets. and so one and so forth.  How can you be in charge of a department without the know-how of the sub-department.

This re-add the Time requirement that would require 1-3 shifts to unlock the roles,
even if this server is more RP-focused than job-focused. this is still an MRP server and should have some sort of basic understanding 

# Changelog

:cl:
- tweak: Re-added time requirements for jobs.


VividPups:
YML is probably going to fail

VividPups:
I would like if another person would Test this It seem that there might be problems I am getting but not sure how to fix do to limited Knowledge

Memeji:
If I get time and no one else has, I'll try it out

PR #891, opened 2025-04-26
Makes the PDA Beep when Messaged, Even if Its Open

Superlagg:
# Description

So, if you have the PDA open, with a conversation selected, and youve been chatting for a while, it doesnt always scroll to the bottom, which tends to make it hard to know if you got a message. That, and even when it does, it can easily get overlooked if you're doing anything other than reading the PDA.

This just makes the PDA always beep when you get a message, regardless if you have the selected thread opened. You can still mute it, dont worry!

---

# Changelog

:cl:
- tweak: The PDA always beeps when you get a message, even if the conversation is already open. Dont worry, you can still mute it!


Memeji (approved):
I myself am fine with this.

Mnemotechnician (approved):

PR #898, opened 2025-04-29
Venteventredux

fenndragon:
# Description

alright, This PR adds back some of the variety we lost from the vent events at a much MUCH reduced rate with much stricter spawning

it also adds back the various entities that had been disabled with anything dangerous being EXTREMELY rare 

in total, the majority of these vent creatures have been reduced in frequency by a factor of 10 or more, with some of the dangerous things showing up 30X less often 

the idea here is to give more variety to rounds on what can and will pop out of the vents, and freshen up the gameplay monotony a little bit. 

(I should also mention, the infamous Reagentslime vent spawn is set to only occur at a minimum of 80 players online) 
---

# Changelog

:cl:
- add: Added vent variety back into the game 
- tweak: Tweaked spawn values and event frequency


Resources/Prototypes/DeltaV/Entities/Mobs/Species/chitinid.yml
```diff
@@ -153,7 +136,7 @@
         sprite: "Effects/creampie.rsi"
         state: "creampie_moth"
         visible: false
-  - type: Chitinid
+  #- type: Chitinid #Floof: apparently there's a performance issue related to this comp.
```
Mnemotechnician: Why does this randomly modify the chitinid prototype?

Resources/Prototypes/Entities/Mobs/NPCs/Rslimes.yml
```diff
@@ -117,189 +117,188 @@
...
+    - id: ReagentSlimeUnstableMutagen
+      prob: 0.0001
+    - id: ReagentSlimeLead
+      prob: 0.00001
+    - id: ReagentSlimechlorinetriflouride
+      prob: 0.0002
+    - id: ReagentSlimePotassium
+      prob: 0.0002
+    - id: ReagentSlimeLotophagoiOil
+      prob: 0.0001
```
Mnemotechnician: This prototype should be moved to the respective station events file.

Mnemotechnician (changes requested):
Other than that, this will need some discussion

Resources/Prototypes/Entities/Mobs/NPCs/Rslimes.yml
fenndragon: I just re-enabled it where it was. didn't bother moving it cause that's not what this PR does

Resources/Prototypes/DeltaV/Entities/Mobs/Species/chitinid.yml
fenndragon: I literally list why in the comment

M3739: So, me being curious as to what exactly the `Chitinid` component is, I found out that this is pertaining to their ability to purge radiation out in the form of chitzite. Our lovely glowy green stone.

However, I too find this change to be **grossly out of scope** of what the PR is aiming to achieve.

Resources/Prototypes/Entities/Mobs/NPCs/Rslimes.yml
Mnemotechnician: The point stands, this is not the correct place for it. It is way harder to find like this.

Resources/Prototypes/DeltaV/Entities/Mobs/Species/chitinid.yml
Mnemotechnician: Two things to mention:
1. This change is not relevant to the PR and not mentioned in the PR body.
2. The statement "apparently there's a performance issue" does not justify the complete removal of one of the unique features of this species. Given that there are no apparent issues with the ChitinidSystem, if the update rate is too high, we can simply reduce it. Without profiling data or testing, I find it hard to believe the performance impact of this system comes in any correlation with other systems, such as device networking, NPC AI, and physics.

fenndragon:
Soooo the vote was yes, And this PR is fixed now, so what's the holdup

Memeji:
I'm still not sure about the Reagent Slimes being a vent Critter event.
It feels like they should be special to the Reagent Anomaly, imo.

fenndragon:
> I'm still not sure about the Reagent Slimes being a vent Critter event. It feels like they should be special to the Reagent Anomaly, imo.

but it got voted on...? didn't we already discuss this?

Memeji:
Ah sorry, can you send me where the vote is? I don't remember where it was at.

Memeji (approved):
Approving as vote is a yes for it.

southbridge-fur:
I asked Mnemotechnician if I could merge this and they said yes so I'm just gonna do that

PR #900, opened 2025-04-30
Upstream Merge #1989 | Power Cell Draw Fix

Raikyr0:
# Description

https://github.com/Simple-Station/Einstein-Engines/pull/1989

Power cell draw wasn't tick invariant, so batteries on this codebase were being drained anywhere from 30 to 60x faster than people expect, depending on server performance.

# Changelog

- fix: Batteries no longer drain power 30 to 60 times faster than physics states. This is particularly a large buff to IPCs and MODSuits, who both no longer have to deal with batteries lasting an absurdly short amount of time.)

### DIFF
--- a/Content.Server/PowerCell/PowerCellSystem.Draw.cs
+++ b/Content.Server/PowerCell/PowerCellSystem.Draw.cs
@@ -30,7 +30,8 @@ public sealed partial class PowerCellSystem
             if (!TryGetBatteryFromSlot(uid, out var batteryEnt, out var battery, slot))
                 continue;
 
-            if (_battery.TryUseCharge(batteryEnt.Value, comp.DrawRate, battery))
+            // TCJ: "Multiplying by frameTime to make this tick-invariant. Otherwise it'll draw 30x to 60x faster than you expect."
+            if (_battery.TryUseCharge(batteryEnt.Value, comp.DrawRate * frameTime, battery))
                 continue;
 
             Toggle.TryDeactivate((uid, toggle));


Mnemotechnician:
That is an extremely weird change, although it does makes sense. But it comes with severe balance implications which upstream obviously never considered. Batteries in general last 5-20 minutes depending on the battery type. Decreasing the rate of power cell draw of all battery-powered devices 30- and 60-fold will make it so that flashlights, translators, ninja suits, anomaly locators, and many other things will be able to last an entire shift on one medium power cell without the need to ever recharge (save for EMP accidents of course). This needs to be considered before this change can be merged.

Memeji (approved):

Memeji:
> That is an extremely weird change, although it does makes sense. But it comes with severe balance implications which upstream obviously never considered. Batteries in general last 5-20 minutes depending on the battery type. Decreasing the rate of power cell draw of all battery-powered devices 30- and 60-fold will make it so that flashlights, translators, ninja suits, anomaly locators, and many other things will be able to last an entire shift on one medium power cell without the need to ever recharge (save for EMP accidents of course). This needs to be considered before this change can be merged.

Fair, we should change the batteries' storage to compensate for this buff.

Mnemotechnician (changes requested):
Yeah. A simple solution would be to simply multiply frameTime by 30/60 (whatever the default update rate is). But this is assuming that upstream is right and frameTime is actually expressed in seconds rather than the fraction of speed the game is running at (I'm not certain myself, and different game engines do it differently)

Raikyr0:
> Yeah. A simple solution would be to simply multiply frameTime by 30/60 (whatever the default update rate is). But this is assuming that upstream is right and frameTime is actually expressed in seconds rather than the fraction of speed the game is running at (I'm not certain myself, and different game engines do it differently)

C# PRs are basically magic to me

From a RP perspective, this gives IPC/Borg players more freedom to RP without relying on Epi researching micro-reactors every round.

Raikyr0:
For whatever reason, I thought this merged and closed the branch by accident.

Memeji:
Any update/comment on Mnemo's requested change above?

Raikyr0:
> Any update/comment on Mnemo's requested change above?

I have many comments on the matter. However none of them involve me fiddling with C# magic that Solidius coded.

But yeah, if nothing else let the IPCs E/RP without having to rush Tier 3 Industrial every round.

Memeji:
> > Any update/comment on Mnemo's requested change above?
> 
> I have many comments on the matter. However none of them involve me fiddling with C# magic that Solidius coded.
> 
> But yeah, if nothing else let the IPCs E/RP without having to rush Tier 3 Industrial every round.

From my understanding, the change may help with that (and mind you I'm not opposed at all for allowing more time to RP), but it will also potentially make batteries and power not much of an issue.

> "But it comes with severe balance implications which upstream obviously never considered. Batteries in general last 5-20 minutes depending on the battery type. Decreasing the rate of power cell draw of all battery-powered devices 30- and 60-fold will make it so that flashlights, translators, ninja suits, anomaly locators, and many other things will be able to last an entire shift on one medium power cell without the need to ever recharge (save for EMP accidents of course). This needs to be considered before this change can be merged."

So I would like to test a solution to that, or see just how different of a timeframe these batteries last with this change (and see if it is as bad as thought), before this is pushed to stable.

@Mnemotechnician Anything to add on this?

Raikyr0:
> From my understanding, the change may help with that (and mind you I'm not opposed at all for allowing more time to RP), but it will also potentially make batteries and power not much of an issue.

I really don't believe batteries and power should be an "issue" rather then a minor inconvenience, people with translators already carry around spare batteries. 

I have more issue with the fact that flashlights and flashlight adjacent permanently becoming unusable at low power. 

As for the Ninja, I think you can treat that as a rare buff to an antagonist which they (antags) direly need.

Mnemotechnician:
If the concern is IPCs, you should reduce *their* power cell draw rates, not completely break the system that's been in place for years.

Raikyr0:
> If the concern is IPCs, you should reduce _their_ power cell draw rates, not completely break the system that's been in place for years.

The concern is IPCs for us but it's also eventually going to be Modsuits and Mechs too. The system is not being broken, rather an issue was found and is getting fixed. Not to mention the 30 to 60 times number is greatly exaggerated from what I've seen on Goob.

Mnemotechnician:
> The concern is IPCs for us but it's also eventually going to be Modsuits and Mechs too. The system is not being broken, rather an issue was found and is getting fixed. Not to mention the 30 to 60 times number is greatly exaggerated from what I've seen on Goob.

Testing is required to see if it's 30-60x or just 2x. That depends on how frameTime is expressed.

Memeji:
Will run this PR locally today and test the power draw change just to confirm!

> > The concern is IPCs for us but it's also eventually going to be Modsuits and Mechs too. The system is not being broken, rather an issue was found and is getting fixed. Not to mention the 30 to 60 times number is greatly exaggerated from what I've seen on Goob.
> 
> Testing is required to see if it's 30-60x or just 2x. That depends on how frameTime is expressed.

Memeji (approved):
Alright, I tested this and tbh I didn't see much of an issue.
Stuff drained at a good speed imo so it should be good to go.

Memeji:
Just make sure to merge master and such, and we should be fine!

PR #901, opened 2025-04-30
Fix Certain Species Fixtures

Forzii:
# Description

So, it turns out there's two separate long-standing bugs with certain species fixture scaling. This appears to be contributing to certain issues, including: incorrect weights, incorrect fixture hitbox, and... I dunno, probably other stuff.

The two issues are 1) originating fork species scaling left in place despite EE's heightAdjustSystem and 2) defaultHeight/defaultWidth being applied as scaling to all characters on top of character creation scaling. Because of how EE's heightAdjustSystem works, the height/width scaling is masked/dropped but the cumulative fixture adjustments stay. This is also the source of one of the bugs caused by my original Tiny trait rework - I unintentionally exposed these repeated fixture adjustments as visual scaling adjustments as well.

So far I have been unable to determine why defaultWidth and defaultHeight are being applied in this way, despite them only being used in two places in the code. The issue almost certainly lies in the SharedHumanoidAppearanceSystem LoadProfile. The same issue also causes the affected species to have a different weight when they are spawned in verses what is displayed in the character creator.

My search found that one or both of these bugs applies to, at minimum, the following races: dwarf (ignored because disabled), rodentia, harpy, shadowkin, felinid, oni, harpy, unathi (reptilian), and resomi. I have provided mitigations to both issues in the form of yaml changes.

I'll give an example of rodentia, since they are what led to me discovering this:

The rodentia yaml contains DeltaV's original scaling that was never removed/changed, the systems just overwrite the sprite scaling. That makes the fixtures scaled down to 0.8 (0.28) despite the code assuming they are at 1.0 (0.35). Then we get to "defaultHeight" and "defaultWidth". It seems that these are taken into account, fixtures adjusted AGAIN down to 0.8 (which is what those are set to for rodentia) before this too is forgotten and assumed to be at 1.0 (0.35). If you then make a rodentia actually be it's default size (0.8, 0.8) via the character creator, this fixture adjustment is made a THIRD time. Your fixtures are then, instead of 0.8 of standard, actually 0.8x0.8x0.8 of standard. 

As a secondary example, Skittish Flora (my primary character) is a 117cm (3'10) rodentia. She weighs 8.9kg (~19lb?). This is absurd (though I love it). With these fixes, she instead weighs 21.6kg (~46lb). This is much more reasonable. These fixes make me a tiny bit sad, but make my logical brain happy. I was baffled by how light some of the characters could get.

---

# TODO

- [x] Finally look into why rodentia and felinid appear to be SO MUCH LIGHTER than they should be. 
- [x] Mitigate both bugs via yaml changes
- [ ] (optional) Fix bug two (2) properly instead of masking it via yaml changes
- [ ] (optional) Consider if these fixes are enough to make lightweight eligible for some of these races

---

<p>

</p>
</details>

---

# Changelog

:cl:
- fix: Mitigated two underlying issues that caused many races to have incorrect mass, unintended hitbox sizes, etc. This impacts rodentia, harpy, shadowkin, felinid, oni, harpy, unathi (reptilian), and resomi.

### DIFF
--- a/Resources/Prototypes/DeltaV/Entities/Mobs/Species/rodentia.yml
+++ b/Resources/Prototypes/DeltaV/Entities/Mobs/Species/rodentia.yml
@@ -25,7 +25,7 @@
       fix1:
         shape:
           !type:PhysShapeCircle
-          radius: 0.28
+          radius: 0.35 # Floofstation - changed from 0.28. Default shape scaling is both not needed and causes issues due to EE's heightAdjust system. 0.35 is standard, DeltaV used 0.28 to shrink their size to 80%
         density: 140
         restitution: 0.0
         mask:
@@ -33,7 +33,7 @@
         layer:
         - MobLayer
   - type: Sprite
-    scale: 0.8, 0.8
+    # scale: 0.8, 0.8 Floofstation - while this doesn't technically cause any issues, this creates potential instability if changes to the heightAdjust system are made. As someone who made changes to the heightAdjust system, this can be very confusing.
     layers:
       - map: [ "enum.HumanoidVisualLayers.TailBehind" ]
       - map: [ "enum.HumanoidVisualLayers.Chest" ]
--- a/Resources/Prototypes/DeltaV/Species/rodentia.yml
+++ b/Resources/Prototypes/DeltaV/Species/rodentia.yml
@@ -13,10 +13,10 @@
   lastNames: names_rodentia_last
   naming: LastFirst
   minHeight: 0.65
-  defaultHeight: 0.8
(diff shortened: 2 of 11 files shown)


Forzii:
Opening for review - I'll leave it up to ya'll if you want to merge now with the YAML mitigations or wait until we can find a real fix for the default Height/Width issue. I can also remove the default Height/Width mitigation and just do the fix for the origin fork scaling being left in.

Mnemotechnician (approved):

Memeji (approved):
Lets try and go with this in its current state and further update it with feedback as it comes in.

PR #922, opened 2025-05-02
Upstream Merge #1505

ghost:
Upstream PR: https://github.com/simple-station/einstein-engines/pull/1505


ShadesMars (approved):

Memeji (changes requested):
Test Fail seems valid.

I presume this PR being added is adding to the test and catching Pebble?

Mnemotechnician:
Yeah, the test itself is built in a dumb way. It should do a "contains" check, not "equals".

The new added filter clause makes it so that the borg job doesn't get counted even if there are spawn points for it...

Memeji:
Yeah, not sure on this one, besides modifying at with what you described. If not, it may need to be skipped to avoid test fail issues moving forward?

PR #935, opened 2025-05-02
Add Safety Harness to Lathes

WhiteHusky:
Also bumps the amount of harnesses in the EngiDrobe.

I want the flock of harnessed engineers to be real.

image

# Changelog

:cl:
- add: Safety harnesses can now be researched and crafted in the Engineering TechFab.
- tweak: A few more safety harnesses were added to the EngiDrobe

### DIFF
--- a/Content.Shared/Lathe/EmagLatheComponent.cs
+++ b/Content.Shared/Lathe/EmagLatheComponent.cs
@@ -8,16 +8,20 @@ namespace Content.Shared.Lathe
     [RegisterComponent, NetworkedComponent, AutoGenerateComponentState]
     public sealed partial class EmagLatheRecipesComponent : Component
     {
+        // FloofStation Modified
+        // AlwaysPushInheritance done to remain consistent with LatheComponent
         /// <summary>
         /// All of the dynamic recipes that the lathe is capable to get using EMAG
         /// </summary>
-        [DataField, AutoNetworkedField]
+        [DataField, AlwaysPushInheritance, AutoNetworkedField]
         public List<ProtoId<LatheRecipePrototype>> EmagDynamicRecipes = new();
 
+        // FloofStation Modified
+        // Ditto from EmagDynamicRecipes
         /// <summary>
         /// All of the static recipes that the lathe is capable to get using EMAG
         /// </summary>
-        [DataField, AutoNetworkedField]
+        [DataField, AlwaysPushInheritance, AutoNetworkedField]
         public List<ProtoId<LatheRecipePrototype>> EmagStaticRecipes = new();
     }
 }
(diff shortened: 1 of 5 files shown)


VividPups:
so, what is this post to be or do?

ShadesMars (approved):
I have no issues with this.

Mnemotechnician (approved):
Would prefer it to be in the engineering techfab instead, but this is fine

Memeji:
> Would prefer it to be in the engineering techfab instead, but this is fine

Agreed here, would fit better in the engi techfab.

WhiteHusky:
@Mnemotechnician @Memeji Done, but since I don't see a way to extend the static recipes short of copy-pasting from the autolathe, it's a dynamic recipe behind salvage research so it's cheap to get (and _passably_ makes sense.)

Mnemotechnician:
> @Mnemotechnician @Memeji Done, but since I don't see a way to extend the static recipes short of copy-pasting from the autolathe, it's a dynamic recipe behind salvage research so it's cheap to get (and _passably_ makes sense.)

Uh-huh. That is indeed a problem. We could mark the StaticRecipes field with the `[AlwaysPushInheritance]` attribute to allow to simply extend the list of recipes (tested this strategy in the ID locks PR), but I don't know how it would affect other lathes. However, I don't mind the current solution either.

WhiteHusky:
> Uh-huh. That is indeed a problem. We could mark the StaticRecipes field with the `[AlwaysPushInheritance]` attribute to allow to simply extend the list of recipes (tested this strategy in the ID locks PR), but I don't know how it would affect other lathes. However, I don't mind the current solution either.

Giving the attribute a glance, it might work and all that would be need to be done is having a separate `Autolathe` component with _no_ recipes.

Loosely, the inheritance would end up like this.
```
BaseAutolathe:
    Autolathe:
        CargoTechFab
        EngineeringTechFab
        […]
                        
    AutolatheNoRecipes:
        Protolathe
        CircuitImprinter
        ExosuitFabricator
        […]
```

Mnemotechnician:
Yea, might be a bit problematic if the protolathe and others inherit from the autolathe...

WhiteHusky:
> Yea, might be a bit problematic if the protolathe and others inherit from the autolathe...

They do, hence having to create a `BaseAutolathe` that `Autolathe` and a new `AutholatheNoRecipes` (or inherit from `BaseAutolathe`, but I like being descriptive) for the Protolathe and others.

WhiteHusky:
Reviewing `lathe.yml` it looks like the attribute can be added without problem since the chain of parents shouldn't cause any unexpected behavior, though on second thought I feel like if you're parenting from another lathe you're probably wanting to extend it's recipes anyhow. And testing it, the lathes don't have anything weird added to them as a result.

So with that change, Autolathe verses EngiFab
image

ghost:
test fails unrelated

PR #941, opened 2025-05-03
Floof/Enable Nightvision Trait

ghost:
Media:
image

:cl:
- add: Re-added the Light Amplification (night vision) trait.


ShadesMars (approved):

Resources/Prototypes/Traits/physical.yml
```diff
@@ -806,24 +806,32 @@
...
-#  functions:
-#    - !type:TraitAddComponent
-#      components:
-#        - type: NightVision
-#
-#- type: trait
+- type: trait
+  id: LightAmplification
+  category: Physical
+  points: -4
```
FoxxoTrystan, line 812:
```suggestion
  points: -5
```

In all honest il say even -5 due to how OP its is.

Raikyr0 (changes requested):
Missing Line at 837

requirements:

Mnemotechnician (approved):
Should be good now (maybe needs testing)

Memeji (approved):
Will need to mostly see how it works in practice for balance reasons, ideally done so on a test server.

Raikyr0:
I feel like test wise, it's better to run it through stable to get a wider demographic and poll it to see if it's too powerful or not,

Personally I think it's fine. It's been nerfed compared to normal.

Raikyr0 (approved):

PR #951, opened 2025-05-04
Friendlier Animals! (And a Non-Exclusive Excommunicant)

XenoOutlier:
Makes animal friend now WORK on more animals and also makes a new faction for hostile clown-mobs also fixes the behonker being affected by animal friend. Plus a niitpick bugfix for the excommunicant still being exclusive and overriding backpacks.

# Description

Have you ever wished animal friend actually worked on anything because it seemed to never work? Congrats now it do, also no mantis' will pull out their hair or lack thereof because they had their backpack replaced by their sawn-off.

---

# TODO

- [x] Everything

---

# Changelog

:cl:
- add: New separate faction for Hostile clown mobs called "ClownRage" do with this as you will.
- tweak: Animal friend now works on: Tarantulas (But not Clown Spiders), Space adders and Gorillas and does not work on the Behonker anymore.
- fix: The excommunicant like many other loadout weapons is no longer exclusive. You may now have a backpack, mantis.


XenoOutlier:
It worked this time thanks again to Carlen

Resources/Prototypes/ai_factions.yml
```diff
@@ -169,3 +171,12 @@
     - GladiabotGreen
     - GladiabotBlue
 
+- type: npcFaction
```
Mnemotechnician: This should be moved to the floofstation namespace

Mnemotechnician (changes requested):
Other than that all should be good, although I'm not sure if allowing salvage to just ignore half of the space critters is a good idea.

Also, you only need to mention details you consider important to other players in the changelog, no need to explain technical details.

Resources/Prototypes/ai_factions.yml
XenoOutlier: What you mean Clown faction?

XenoOutlier:
> Other than that all should be good, although I'm not sure if allowing salvage to just ignore half of the space critters is a good idea.
> 
> Also, you only need to mention details you consider important to other players in the changelog, no need to explain technical details.

The issue is if you aren't playing salvage it only affects slimes basically

Resources/Prototypes/ai_factions.yml
XenoOutlier: Either way, moved it to Floof > Entities > Mobs under "clowns.yml"

Mnemotechnician: That's not the correct namespace for it either (supposed to be /_Floof/ai_factions.yml), but that'll do

Mnemotechnician (approved):

Memeji (approved):

PR #956, opened 2025-05-05
Change Dragon Briefing and Fix Uncommented Dragon System Changes

Cprn-Sprkl:
# Description

Adds to the dragon briefing to specify that the player does not have to worry about rift timers on this server.

Adds back previously deleted code and comments it out with an explanation for why this block of code was removed. Also properly tags this as a floofstation change.

---

# Changelog

:cl: sprkl
- tweak: Space dragons are made more aware they won't vanish without rifts.

### DIFF
--- a/Resources/Locale/en-US/dragon/dragon.ftl
+++ b/Resources/Locale/en-US/dragon/dragon.ftl
@@ -2,4 +2,4 @@ dragon-round-end-agent-name = dragon
 
 objective-issuer-dragon = [color=#7567b6]Space Dragon[/color]
 
-dragon-role-briefing = dragon-role-briefing = Summon 3 carp rifts and take over this quadrant; or don't, and spend some time seeing the sights, meeting the locals, or building an interesting horde. It's not like you'll disappear without those rifts.
+dragon-role-briefing = Summon 3 carp rifts and take over this quadrant; or don't, and spend some time seeing the sights, meeting the locals, or building an interesting horde. It's not like you'll disappear without those rifts.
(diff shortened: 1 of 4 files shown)


Cprn-Sprkl:
Test fail seems to simply say it took too long to run and quit.

Forzii:
Could you also modify Resources/Locale/en-US/objectives/conditions/carp-rifts.ftl? The only reason I never pushed my PR is because I couldn't get dragons to spawn with their objectives in my local environment. Never figured out why, and I don't want to push untested changes.

`objective-carp-rifts-description = Use the rift action to open {$count} rifts and ensure they do not get destroyed. If you don't open a rift after 5 minutes, you get killed.`
to something like
`objective-carp-rifts-description = Use the rift action to open {$count} rifts and ensure they do not get destroyed. Rifts are NOT necessary for your survival.`

Either that or make a floof-specific description and change the localization string of the objective itself instead of the ftl. Either way, if you're willing. S'up to you.

Mnemotechnician (approved):

southbridge-fur (approved):
good change

Resources/Locale/en-US/dragon/dragon.ftl
```diff
@@ -2,4 +2,4 @@ dragon-round-end-agent-name = dragon
 
 objective-issuer-dragon = [color=#7567b6]Space Dragon[/color]
 
-dragon-role-briefing = Summon 3 carp rifts and take over this quadrant!
+dragon-role-briefing = dragon-role-briefing = Summon 3 carp rifts and take over this quadrant; or don't, and spend some time seeing the sights, meeting the locals, or building an interesting horde. It's not like you'll disappear without those rifts.
```
Forzii: Might want to remove the extra `dragon-role-briefing = `

Cprn-Sprkl: Good catch!

ShadesMars (approved):

PR #960, opened 2025-05-07
Remove Default Hotbar Loadouts Keybinds

Mnemotechnician:
# Description
People are not used to them, some people use shift to sprint (why?), and they don't want to rebind them, something like that. Should still be possible to rebind them in the settings menu.

# Changelog
:cl:
- tweak: Removed default keybinds for switching between hotbar loadouts (shift+num). You can still bind them yourself using the settings menu.

### DIFF
--- a/Resources/keybinds.yml
+++ b/Resources/keybinds.yml
@@ -557,46 +557,47 @@ binds:
 - function: Hotbar9
   type: State
   key: Num9
-- function: Loadout0
-  type: State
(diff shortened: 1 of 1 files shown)


Mnemotechnician:
Will need to check to make sure it works as expected, but I'm certain it will work just fine

ShadesMars (approved):

Memeji (approved):

PR #968, opened 2025-05-09
Fix ID Locks and ID Consoles

Mnemotechnician:
# Description
1. Fixes the bug where ID consoles wouldn't update the target's station records unless their job was changed and the changer had permission to change ALL of their accesses
2. Fixes inheritance issues with id-lockable entities that have lead to all departmental closets inheriting the id lock component from ClosetBase and not from their respective departmental versions, and not revealing the owner's info.
3. Fixes the bug with the ID lock system checking for non-localized job name (which is always null for some reason, all systems use LocalizedJobTitle instead)

<p>

image
image
image
image
image

</p>
</details>

# Changelog
:cl:
- fix: The ID card console will now correctly update your station records even if your job and accesses stay the same.
- fix: ID-lockable closets and other things should now correctly display the owner's info (unless the locker is anonymous)


ghost:
unrelated test fail

PR #969, opened 2025-05-09
Weaponized Dildos

Raikyr0:
# Description

Dildos and Vibrators can now be swung like weapons with a stun value

Also changed Vibrators to continue making the noise till turned off

Thanks @ItsDzordz for testing, fixing the rotation (on the wide swing) and the good sound loop (for the vibrator).

# Changelog

:cl:

- tweak: Dildos and Vibrators are now melee weapons
- fix: Vibrator noise now continues till turned off


southbridge-fur:
You have no idea how much joy it brings me to see a PR like this show up in my notifications. 

I don't think that test fail has anything to do with you. It seems to be an issue with what I imagine was an upstream merge

Raikyr0:
> I don't think that test fail has anything to do with you. It seems to be an issue with what I imagine was an upstream merge

Yeah nah, the fail is just an exit code 1 which means it just did the random crash it usually does, completely ignorable.

Mnemotechnician (approved):

Mnemotechnician:
> Yeah nah, the fail is just an exit code 1 which means it just did the random crash it usually does, completely ignorable.

No. Any exit code other than 0 is an error. Error code 1 usually indicates a generic/uncategorized error that the application will print more info about to STDERR or a log file. If you look at the top of the log, you will find the actual error message: JobCyborg having two primary departments whereas only one is allowed. It is however not related to your PR and is caused by the (already reverted) AI merge.

southbridge-fur:
(that means you should git pull master)

Memeji (changes requested):
Make sure to pull master to get the fix for the checks.

Memeji (approved):
checks passed, good to go

PR #971, opened 2025-05-11
Removes the HarpyHideWing Tag

VividPups:
# Description
Multiple, if not all, Vulp mains want their tail showing when wearing a Hardsuit or an outersuit. However, when looking into it. The problem at first stems from the Singer trait, going deeper into it it actually is from a tag called HarpyHideWing.  Now 
  
After removing the tags, and checking MULTIPLE different tails, wings on both Vulps and Harpies I haven't seen any reason to have it.  If we discover something in the near or far future, it would be better to change the Component vs reading the tags 

# Changelog
:cl:
- fix: Vulp and Harpys tails now show in outerwear


VividPups:
I would also recommend someone looking into the Singer trait. Might be able to separate the tag from the component However, at that point I would not be able to help

Mnemotechnician:
Interesting choice. Did you make sure harpy wing covers still work correctly? (some hardsuits feature custom wing covers, such as the paramedic hardsuit)

VividPups:
It seems that when I remove the take it removed the cover wings but when I added the Tag back to the hardsuit instead of the based. I still does not show the covers. Just hide the tail again

VividPups:
for example

image
image

VividPups:
image
image
image


this is me adding the Tag back to the based hardsuit also and it doesn't seem to work

VividPups:
image
image
image

This is from my updated master with NO changes to basic wings.

it seems that the harpy's cover wings might be broken.

VividPups:
wow thats really dumb

PR #976, opened 2025-05-13
New Wedding Rings

VividPups:
# Description

NEW wedding rings to pick

# Media 
image

# Changelog

:cl:
- add: 5 new wedding rings in the autoDrobe and Loadouts


Raikyr0:
Is the top right one a silk ring and not a silver one?

Mnemotechnician (approved):
I think it looks good. Would prefer the new entities to share the same parent, but it's not a requirement since they don't have shared components (yet).

Memeji (approved):
Beautiful sprites, also will parrot what Mnemo said. But that can be changed if it ever has functionality added

VividPups:
> Is the top right one a silk ring and not a silver one?

Yes, the top ones are silk not sliver. Silk is cooler

VividPups:
> I think it looks good. Would prefer the new entities to share the same parent, but it's not a requirement since they don't have shared components (yet).

Yea, I didn't expect them to be just a Sprite change, which is why I kept it with the same parent instead of parenting off of the Gold ring

PR #981, opened 2025-05-15
Kitsune Fixes + EE 1709

M3739:
# Description

This PR aims to implement @SolStar2's fixes, namely DV 3724,  3725, and 3726.

3725, and 3726 was applied successfully. 3724, however, was found to be a solution to be an issue we do not possess: an issue found to be unrelated to the Kitsune existed in our codebase revolving around the cloning system, which would later be resolved by implementing EE 1709 by 'VMSolidus', which fixed an issue resulting in the clone not properly copying the appearance of the original person.

Upon testing after implementation of 1709 showed that 3724 was redundant as it properly transferred colors over to the cloned kitsune's fox form and fox fire. It should be worth noting that 3724 was DV's solution to ensure the cloned kitsune will retain their fox form and fox fire's colors. And therefore is noted down as such for future reference, in the event where an upstream merge causes the issue to finally appear on our end in a similar nature.

Another change is purely backend, Wizden 35017 by 'poklj' would move `CloneAppearance` from `HumanoidAppearanceSystem` to `SharedHumanoidAppearanceSystem`. This is because the Changelings added back in Wizden's 34002 had their `SharedSystem` requiring the ability to modify someone's appearance.

While we do not have Changelings in our own build, this was originally a prerequisite for 3724, but as 3724 was found to be redundant, it has been kept for the possibility of compatibility with future additions.

Wizden 35017 is able to be removed upon the request of maintainers.

Addendum:
After further discussion with SolStar, the kitsune's density has been changed to be the same as the Felinids.

<p>

image
image

</p>
</details>

---

# Changelog

:cl: DeltaV SolStar, Einstein Engines VMSolidus, M3739
- fix: Cloning should now properly replicate the original body's appearance.
- fix: Female kitsune should now no longer laugh like a man when in fox form.
- fix: Kitsune foxfire will now emit light, regardless of how dark the color is.
- fix: Kitsune should now no longer weigh as much as a feather.

### DIFF
--- a/Content.Server/Cloning/CloningSystem.Utility.cs
+++ b/Content.Server/Cloning/CloningSystem.Utility.cs
@@ -327,7 +327,9 @@ public sealed partial class CloningSystem
                 pref = pref.WithFlavorText(flavorText);
 
             _humanoidSystem.LoadProfile(mob, pref);
+            return;
         }
+        _humanoidSystem.LoadProfile(mob, pref);
     }
 
     /// <summary>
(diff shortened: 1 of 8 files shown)


Memeji (approved):
Looks good

PR #986, opened 2025-05-16
Fix Changelog for #935

WhiteHusky:
It was previously behind a research, but later in the PR became another recipe in the engineering techfab.

Plus, uh, lets see what happens if we try to fix history?


ShadesMars (approved):

Memeji (changes requested):
May need to merge master for the test fail fix.

PR #1003, opened 2025-05-20
Upstream Merge #1528 | AddTraitSpecial

ghost:
Upstream PR: https://github.com/simple-station/einstein-engines/pull/1528
Adds a way to specify traits in the jobs special area.

Hashset so you can specify multiple traits. Tested on Nuclear14 to give the tribals a specific language in a friendly way.
```
  special:
  - !type:AddTraitSpecial
    traits: [ LanguageTribal, LanguageChinese ]
```

### DIFF
--- a/Content.Server/Jobs/AddTraitSpecial.cs
+++ b/Content.Server/Jobs/AddTraitSpecial.cs
@@ -1,0 +1,30 @@
+using Content.Server.Traits;
+using Content.Shared.Roles;
+using Content.Shared.Traits;
+using JetBrains.Annotations;
+using Robust.Shared.Prototypes;
+
+namespace Content.Server.Jobs;
+
+
+[UsedImplicitly]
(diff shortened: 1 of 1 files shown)


ShadesMars (approved):

Mnemotechnician (approved):
This is unsafe because it doesn't check if the mob already has the trait, and some traits are not duplicate-safe... But it's fine, I guess.

Memeji:
> This is unsafe because it doesn't check if the mob already has the trait, and some traits are not duplicate-safe... But it's fine, I guess.

What are the possible issues coming from duplicate? We should probably take that into account.

Mnemotechnician:
> What are the possible issues coming from duplicate? We should probably take that into account.

Foreigner trait gives you a translator every time the relevant component is added, psi-power traits can add multiple instances of the same power I imagine, and more. Some minor inconveniences.

PR #1007, opened 2025-05-20
Remove Deathgasp Emote From the Emote Menu

TheRealMasterChief117:
# Description
The deathgasp emote is prohibited via the rules, yet it's still able to be used as an emote in the menu?
This PR removes the emote from the emote menu. It will still work when a mob dies, however.

image
image
Doesn't appear in the emote menu.

:cl:
- remove: Removed being able to use the deathgasp emote from the emote menu.

### DIFF
--- a/Resources/Prototypes/Voice/speech_emotes.yml
+++ b/Resources/Prototypes/Voice/speech_emotes.yml
@@ -251,13 +251,13 @@
 - type: emote
   id: DefaultDeathgasp
   name: chat-emote-name-deathgasp
-  icon: Interface/Emotes/deathgasp.png
-  whitelist:
-    components:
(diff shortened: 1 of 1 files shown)


M3739:
> The deathgasp emote is prohibited via the rules,

If I may ask, which specific rule (or rules) does this violate?

TheRealMasterChief117:
> > The deathgasp emote is prohibited via the rules,
> 
> If I may ask, which specific rule (or rules) does this violate?

The rules might've changed last time I checked. It used to be prohibited, though. I'll try to find it.

TheRealMasterChief117:
ok so i cant find it so this pr is useless, if an admin can confirm tho ill open

PR #1009, opened 2025-05-20
Change to Hypno Trait

Raikyr0:
# Description

Changes Hypnotic Gaze to use 0 points, and allows it to be used on mind broken individuals (Black-eye Shadowkin)
(Also changed a "His" to a "Their")

Why/Reason
Roleplay only Trait, also no reason why you can't dominate the mind of someone who has no soul. 

# Changelog

:cl:
- tweak: Hypnotic Gaze now costs 0 and can be used on mind broken individuals (X-waveform and Black-eyes)


ShadesMars (approved):
While this is fine, the doafter is broken, this should be fixed.

And in the shadekin rework they are not mindbroken anymore.

Memeji:
> While this is fine, the doafter is broken, this should be fixed.
> 
> And in the shadekin rework they are not mindbroken anymore.

What would need to be fixed in the doafter?

Mnemotechnician (approved):
I don't think it should cost 0 as it is a buff, but since it's a mostly-ERP trait, it's whatever.

Raikyr0:
> I don't think it should cost 0 as it is a buff, but since it's a mostly-ERP trait, it's whatever.

I like to say, it still costs a trait slot. While some may not consider a trait slot as resource, it's still limited till someone writes a function that allows certain traits to take no slot.

PR #1022, opened 2025-05-24
Upstream Merge #1828/#2008/#2011/#2382 | 10 Point Trait Buff + Trait Touch Ups

Raikyr0:
# Description

Upstream Merge which changes the big 10 pointer traits to be more impactful rather then newbie traps.

This changes Platelet Factories to the one EE is currently using, which provides 5 times the healing (4 times airloss) [#1828](https://github.com/Simple-Station/Einstein-Engines/pull/1828) and greatly improved natural blood regeneration. [#2382](https://github.com/Simple-Station/Einstein-Engines/pull/2382)

Also changes Power Overwhelming to no longer use 2 spell slots, which made people unable to get new spells.  [#2011](https://github.com/Simple-Station/Einstein-Engines/pull/2011) Does not include the full #2011 pull, as I'm pretty sure we'll spend all week arguing over healing word being a round start trait.

Tweaks to Psionics cost, effectively useless/niche round start powers are slightly cheaper, 

Striking Calluses no longer has a bazillion restrictions, +1 damage and is cheaper. (changed from the original to use the wide punch we have here)

Spinarette works a little faster now.

Fixes the Oni Traits numbers too [#2008](https://github.com/Simple-Station/Einstein-Engines/pull/2008)

Added Oni Mystic Trait, which is high potential but you give up most of your melee bonuses (30% down to 5%) 

I started this as a fix for the 10 cost traits, and ended up having to merge the rest of the stuff along with it so it doesn't cause issues.

# Changelog

:cl:

- tweak: Platelet Factories buffed from 0.07 to 0.35 (0.25 airloss) and improves natural blood regeneration.
- tweak: Power Overwhelming no longer uses 2 spell slots. (For those who don't understand the esoteric psionics system, makes it so you don't have to drink 3 times as many drugs for your first spell)
- tweak: Striking Calluses doesn't have 3 different requirements anymore, does a little more damage and costs less.
- tweak: Spinarette spins slightly faster.
- tweak: Roundstart psionic powers are slightly cheaper (because they're niche/useless)
- tweak: Low/High Potential are now lower and higher.
- tweak: Oni traits aren't a debuff anymore.
- added: Oni Mystic trait for the Shrine Maidens, nerfs Oni melee bonus in exchange for potentia gain.

### DIFF
--- a/Resources/Prototypes/Traits/physical.yml
+++ b/Resources/Prototypes/Traits/physical.yml
@@ -451,30 +451,17 @@
 - type: trait
   id: StrikingCalluses
   category: Physical
-  points: -4
+  points: -2
   requirements:
     - !type:CharacterJobRequirement
       inverted: true
       jobs:
         - Prisoner # Bionics should be "Confiscated" from long term prisoners.
-        - Gladiator # Floof
-    - !type:CharacterSpeciesRequirement
-      species:
-        - Human # Entirely arbitrary, I've decided I want a trait unique to humans. Since they don't normally get anything exciting.
-                # When we get the Character Records system in, I also want to make this require certain Backgrounds.
     - !type:CharacterTraitRequirement
       inverted: true
       traits:
         - Claws
         - Talons
-    - !type:CharacterLogicOrRequirement
-      requirements:
-        - !type:CharacterTraitRequirement
-          traits:
-            - MartialArtist
-        - !type:CharacterJobRequirement
-          jobs:
-            - Boxer
   functions:
     - !type:TraitReplaceComponent
       components:
@@ -485,7 +472,7 @@
           animation: WeaponArcFist
           damage:
             types:
-              Blunt: 6
+              Blunt: 7
 
 - type: trait
   id: Spinarette
@@ -508,7 +495,7 @@
       components:
         - type: Sericulture
           action: ActionSericulture
-          productionLength: 2
+          productionLength: 1
           entityProduced: MaterialWebSilk1
           hungerCost: 1
     - !type:TraitAddTag
@@ -718,6 +705,7 @@
     - !type:CharacterDepartmentRequirement
       departments:
         - Security
+        - Command
     - !type:CharacterTraitRequirement
       traits:
         - CyberEyes
--- a/Resources/Prototypes/Traits/physical.yml
+++ b/Resources/Prototypes/Traits/physical.yml
@@ -554,11 +554,11 @@
           damageCap: 400
           damage:
             groups:
-              Brute: -0.35
-              Burn: -0.35
+              Brute: -0.30
+              Burn: -0.30
               Airloss: -0.25
-              Toxin: -0.35
-              Genetic: -0.35
+              Toxin: -0.30
+              Genetic: -0.2
         - type: BloodDeficiency # Incredibly actually works
           bloodLossPercentage: -0.025
 
--- a/Resources/Prototypes/Traits/physical.yml
+++ b/Resources/Prototypes/Traits/physical.yml
@@ -553,6 +553,10 @@
       inverted: true
       species:
         - IPC
+    - !type:CharacterTraitRequirement
+      inverted: true
+      traits:
+        - BloodDeficiency
   functions: # TODO: Code Platelet factories as an actual obtainable implant, and replace this with TraitAddImplant
     - !type:TraitReplaceComponent
       components:
@@ -560,14 +564,16 @@
           allowedStates:
           - Alive
           - Critical
-          damageCap: 200
+          damageCap: 400
           damage:
             groups:
-              Brute: -0.07
-              Burn: -0.07
-              Airloss: -0.07
-              Toxin: -0.07
-              Genetic: -0.07
+              Brute: -0.35
+              Burn: -0.35
+              Airloss: -0.25
+              Toxin: -0.35
+              Genetic: -0.35
+        - type: BloodDeficiency # Incredibly actually works
+          bloodLossPercentage: -0.025
 
 - type: trait
   id: DermalArmor
--- a/Resources/Locale/en-US/traits/traits.ftl
+++ b/Resources/Locale/en-US/traits/traits.ftl
@@ -260,17 +260,22 @@ trait-description-LightStep =
 trait-name-Swashbuckler = Swashbuckler
 trait-description-Swashbuckler =
     You are an expert in swordsmanship, wielding swords, knives, and other blades with unrivaled finesse.
-    Your melee Slash bonus is increased to 35%, but your melee Blunt bonus is reduced to 20%.
+    Your melee Slash bonus is increased to 35%, but your melee Blunt bonus is reduced to 25%.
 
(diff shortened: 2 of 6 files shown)


Resources/Prototypes/Traits/physical.yml
```diff
@@ -553,21 +540,27 @@
...
-              Brute: -0.07
-              Burn: -0.07
-              Airloss: -0.07
-              Toxin: -0.07
-              Genetic: -0.07
+              Brute: -0.35
+              Burn: -0.35
+              Airloss: -0.25
+              Toxin: -0.35
+              Genetic: -0.35
```
Mnemotechnician: I am heavily against this. This will get a person from crit to full health in 5 minutes. Platelet factories users are already able to completely ignore the need to go to medical, this would only worsen the problem.

Mnemotechnician (changes requested):

Resources/Prototypes/Traits/physical.yml
Mnemotechnician: If we want to accept this change, I propose making damage regeneration consume hunger and thirst similar to blood regeneration.

Resources/Prototypes/Traits/physical.yml
```diff
@@ -553,21 +540,27 @@
...
+        - type: BloodDeficiency # Incredibly actually works
+          bloodLossPercentage: -0.025
```
Mnemotechnician, line 563: This is... also no. You should modify (notice: modify, not replace!) the Bloodstream component instead, as this also provides free blood regeneration

Resources/Prototypes/Traits/physical.yml
Raikyr0: This is a discussion I had with the EE balance team before and no you can't go from crit to full unless under an extremely niche scenario. The single reason why Airloss is -0.25 instead of -0.35.

As for the medical issue, current users heal 4 damage every minute. There is no way current Platelet Factory users can avoid going to medical if they crit, not to mention, according to an user, current Platelet Factories is actually slower compared to natural regeneration and is a nerf since it replaces the natural regen.

Ultimately, this is a trait that is a 10 point commitment. The current Platelet Factories is not something anyone would take if they knew the actual numbers.

Resources/Prototypes/Traits/physical.yml
Raikyr0: Also a discussion I had with the EE balance team, the blood regeneration doesn't occur unless you have full food and water. (Which rapidly deplete when it has to regenerate blood) Also has a long delay between losing blood and starting to regenerate blood.

Which gives the user a reason to visit the kitchen and stay hydrated.

Mnemotechnician: Oh yea, blood deficiency modifies the bloodstream component. This is acceptable, although very hacky.

Resources/Prototypes/Traits/physical.yml
Mnemotechnician: You are missing the point. The point is that this is free healing.

Raikyr0: Which is the whole point of this trait. 

It was free healing before, but horrifically bad for it's value. This merely makes it so, we're not false advertising users into thinking this 10 cost trait is good because it costs 10.

Mnemotechnician: it was never horrific. In my experience, every salvage powergamer would take it to avoid going to medical even in case of injuries they cannot treat. No matter the trait cost, such advantage cannot be justified on a server where roleplay and interactions between players take priority over raw gameplay.

Raikyr0: Every Salvage power gamer already avoids going to medical and simply asks for a medical scanner and enough meds to last them the round, hell some rounds I see a dedicated medic working with Salvage, either from Salvage or Medical. 

They were never going to voluntarily go to medical regardless of trait, this is an issue with the player and the job. Salvage is a job that spends most of it's shift off station, and any good salvage doesn't end up in medical. Salvage is a role that inherently avoids role play unless they're actively not doing their job.

Not to mention there's hardly any roleplay with medical, they come in, lie down on a bed, get scanned, injected with chems, leave without saying anything but thanks, sometimes not even.

Mnemotechnician: Exactly, and we shouldn't be making the situation worse.

Archeonan:
God please yes. So many of my characters are implied to have a sort of natural regeneration but I just never take the natural regeneration perk because of how ass it is

Resources/Prototypes/Traits/physical.yml
Memeji: So I can understand having a discussion to buff it from 0.07, that is fair.

But 0.35 is indeed WAY too much, I tested it out. The healing is insanely good at keeping you alive. Which makes sense but might be too much.

When added with Liquor Lifeline, you can go down, and if your enemy is one traitor and too busy dealing with other sec, you can get back up pretty reliably.

This could turn everyone you fight into a guess on if they'll get up, or if you have to confirm their death. Which would be hard in a situation as a single traitor or a Lone operative.

You'd may need to resort to explosives in case of a sec using something like Liquor + Platelet.

From my testing, Liquor can probably give you a few seconds. Taking you out of crit and you have that one chance.

Adding Platelet onto it gives you _many_ more chances.
You go down, if not targeted you can get up again.
Go down again? Same thing, you can get up if the other player doesn't target you. Which can be really hard to do when they're being shot at by several of your allies.

I have a feeling that with this kind of buff, you could have the sec department all have just platelets, and it'd already reduce the amount of meds needed to get themselves back into the fight.
That's not accounting for adding Liquor Lifeline onto it.

In turn, you'd see more explosives being used because it would be the only reliable way to kill fast enough.


I believe 0.35 is too much considering what you can do with it.
Perhaps 0.1 or 0.15?
_Maybe_ 0.2, at most imo. I may do some testing tweaking the values to see where it can fit.

Raikyr0: The number was accepted upstream, most likely due to the increased number actually hostile antags.
We can use the Foxtrot numbers which is 0.3 if that's desired.

As for Liquor Lifeline, we need to consider how many people out there, will run this 16 point setup and inject pure ethanol into their system (which might honestly be LRP powergaming) to achieve this effect. I can expect this from a LRP server, but people will still take their character traits based on RP on our MRP server.

Also, if you go down the first time, and somehow get back up, you will likely still be suffering from severe bloodloss and likely go back down from the bloodloss.

But yeah, TLDR; taking Liquor Lifeline and Platelets together is a niche scenario which would require a perfect setup to get back up.

From what I've seen, sec isn't the department that takes Platelets the most, it's more likely to be passengers, salvage and epistemics.

Memeji (approved):
Many interesting changes, I'm going to approve this PR for testing on Stable for V26.
Then give it a week or two to see how it goes.

Memeji:
Alright, I will be merging this PR for testing next time Master is merged to Stable.

PR #1034, opened 2025-05-27
Fix CPR Sound Looping

Mnemotechnician:
# Description
Fixes infinite CPR sound looping when the performer of the CPR attempts to start a second CPR while the first one is ongoing

# Changelog
:cl:
- fix: CPR sound will no longer loop indefinitely. Hopefully.


Mnemotechnician:
Should actually be fixed now that I've figured out the cause of it.

Memeji (approved):

PR #1036, opened 2025-05-27
Revert "Update PowerCellSystem.Draw.cs (#900)"

Mnemotechnician:
# Description
Just as I predicted, the power cell draw fix made entities with PowerCellDraw (translators, anomaly locators, etc) draw power 60 times slower than normal.

Translator's battery after 2 minutes of usage:
image

# Changelog

:cl:
- fix: Reverted the power cell fix you may find mentioned below (above?) in the changelog.


Memeji (approved):
Tested with flashlights, turns out they use a different method than other items which is why it seemed fine.

PR #1055, opened 2025-06-04
Bite Interaction and some Interaction Verb Fixes

Mnemotechnician:
# Description
Adds a new bite interaction. A relatively fast and close-range interaction that deals between 2.5 and 7.5 pierce and has a cooldown of 5 seconds (on average will deal at most 1 damage per second if the user perfectly spams it every 5 seconds, in reality will be lower). Requires your mouth to not be obstructed (a mask or a muzzle would stop you from using it), but does not require free hands.

I think it serves as a fun way to retaliate if you are cuffed and your captor is standing too close. Also could be used in ERP context and whatnot.

Also:
- Adds the same behavior to lick and kiss interactions. They no longer require your hands to be free (you can use them while cuffed), but now require your mouth to be unobstructed.
- Likewise, makes it so that looking at things does not require free hands or being able to interact otherwise.
- Allows you to lick windows and anomalies, for the funny; and kiss/lick/bite food for dramatic effect. Neither of those interactions do anything other than show popups.
- Adds two new field to InteractionVerbPrototype that define whether the verb should check the user's consciousness in order to be performable and whether CanInteract should be checked (previously the latter was never checked and the latter was checked awlays).
- Fixes petting, hugging, and other no-op interactions having failure effects ("X fails to do Y"), but having no defined locales for them - by removing the said effects.

---

# TODO
- [X] Add fun
- [X] Sleep
- [X] Test fun
- [X] Fix fun
- [X] Test fun again
- [X] Realize the ~~third~~ fourth step didn't help and fix again

---

<p>

Basics:

https://github.com/user-attachments/assets/73f1e160-158e-4df2-8e26-9939e69db305

Interactions that don't require hands:

https://github.com/user-attachments/assets/b0914620-2f96-4866-965a-eed1c44eb1b5

Addendum:

https://github.com/user-attachments/assets/00d1ed01-ecbe-4a79-ab37-543ee2aa5b97

</p>
</details>

# Changelog
- add: Added a new interaction - bite. Deals minor piercing damage and can only be used in very close range and with a long cooldown.
- tweak: Bite, Lick, Kiss, and LookAt interactions can now be used even while you are cuffed or otherwise cannot use hands. The first three of them instead require your mouth to be free, however (e.g. a muzzle or a helmet would stop you).
- tweak: You can now lick windows and anomalies. Yum.

### DIFF
--- a/Content.Server/InteractionVerbs/Actions/ModifyHealthAction.cs
+++ b/Content.Server/InteractionVerbs/Actions/ModifyHealthAction.cs
@@ -10,6 +10,11 @@ public sealed partial class ModifyHealthAction : InteractionAction
     [DataField(required: true)] public DamageSpecifier Damage = default!;
     [DataField] public bool IgnoreResistance = false;
 
+    /// <summary>
+    ///     Floofstation - a random factor applied to the damage.
+    /// </summary>
+    [DataField] public InteractionVerbPrototype.RangeSpecifier RandomFactor = new() { Min = 0.75f, Max = 1.25f };
+
     public override bool IsAllowed(InteractionArgs args, InteractionVerbPrototype proto, VerbDependencies deps)
     {
         return deps.EntMan.HasComponent<DamageableComponent>(args.Target);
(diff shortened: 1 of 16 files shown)


Archeonan:
I dunno if its possible but if you have the masochist trait, getting bit should give you a mood boost

Mnemotechnician:
> I dunno if its possible but if you have the masochist trait, getting bit should give you a mood boost

It's possible, but not in the scope of this PR. I'll add some kinda template for that in case someone decides to implement it in the future though.

Mnemotechnician:
It is done

Memeji (approved):
LGTM

PR #1059, opened 2025-06-05
Holymelon Halo to Loadout

Raikyr0:
# Description

Originally I was gonna add butcher-able to the holymelon to get the halo from Goobstation but turns out the code is exactly the same, just no one has figured out that you need to butcher interact it with a knife instead of cutting it.

So, this just adds it to the loadout, which allows it to be renamed and stuff.

# Changelog
:cl:
- add: Holymelon Halo to Loadout


ShadesMars (approved):

Memeji (approved):
Seems fine! Cool Watermelon Halo time.

PR #1060, opened 2025-06-05
Upstream Merge #2170 | Loadout And Trait Slot Cost

Raikyr0:
# Description

https://github.com/Simple-Station/Einstein-Engines/pull/2170

This PR gently reworks the traits and loadouts system to where both
systems have a new Datafield for "Slots", which declares how many
selections the trait/loadout occupies. The actual implementation of
these differs between the two, since Traits has an actual selection
limit, while Loadouts do not(but instead respect CharacterItemGroups).
In both cases, Slots are defaulted to 1, and can be omitted in most
cases.

Where this shines particularly is in Traits, where it serves as a new
second line for trait balancing. Traits such as Accents, which are
completely worthless, otherwise would never be taken at all since they
compete with other traits for Opportunity Cost. Conversely, there's only
so many points we can increase a powerful trait to, (literally, they
cannot cost more than the current trait maximum). So traits can be made
to occupy >1 slot selections.

Slots can be defined for Item Groups as well, which is what this
datafield does on Loadouts, which both of these share. But since Traits
have two different places that their slots are used, I've opted to
create a split slot datafield for Traits. To simplify things, their
ItemGroupSlots defaults to their Slots datafield, but can be
independently changed as desired.

Minor Difference from the Original PR, this doesn't include changes to Accents yet, I'll add em in after this.

# Changelog

:cl:
- There's no player facing changes, just framework for future ones

### DIFF
--- a/Content.Client/Lobby/UI/HumanoidProfileEditor.xaml.cs
+++ b/Content.Client/Lobby/UI/HumanoidProfileEditor.xaml.cs
@@ -1659,7 +1659,7 @@ namespace Content.Client.Lobby.UI
                     continue;
 
                 points += preferenceSelector.Trait.Points;
-                _traitCount += 1;
+                _traitCount += preferenceSelector.Trait.Slots;
             }
 
             TraitPointsBar.Value = points;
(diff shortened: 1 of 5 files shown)


Mnemotechnician:
Please, use git to cherry-pick the changes. Do not copy them by-hand.

Mnemotechnician:
Did some testing, seems to work.

Mnemotechnician (approved):

PR #1065, opened 2025-06-05
Slot Cost Removal (Uses #1060)

Raikyr0:
# Description

Uses PR #1060 to add slot cost removal to accents, languages, vision and sex traits

Discuss if changes need to be adjusted, this will likely be in maintainer discussion for a while. (altho I could just remove those and get the slot removal pushed in first)

# Changelog

:cl:
- add: Added slot cost removal to accents, languages, vision and sex traits


Mnemotechnician:
This needs testing. We can't merge webedit-ed code-altering PRs without manual testing.

Mnemotechnician:
Please update the PR so we can test & merge it. I apparently cannot push an update to this branch.

Resources/Prototypes/_Floof/Traits/lewd.yml
```diff
@@ -59,10 +62,11 @@
     - type: SolutionContainerManager
       solutions:
         vagina:
-          maxVol: 25
+          maxVol: 30
```
Mnemotechnician: Why does this PR change this? Same with the other trait.

Resources/Prototypes/Traits/skills.yml
```diff
@@ -155,7 +155,7 @@
 - type: trait
   id: LightStep
   category: Auditory
-  points: -2
+  points: -1 # Only Works when you have no shoes on
```
Mnemotechnician: Not true, works even with shoes. As someone who uses this trait all the time, I think the cost is justified.

Resources/Prototypes/Traits/physical.yml
```diff
@@ -742,7 +746,7 @@
 - type: trait
   id: Redshirt
   category: Physical
-  points: 8
+  points: 10
```
Mnemotechnician: This not justified, 8 points is already a lot for simply not having a crit state.

Resources/Prototypes/Traits/physical.yml
```diff
@@ -598,7 +601,8 @@
 - type: trait
   id: CyberEyes
   category: Physical
-  points: -4
+  points: -2 # Exists to enable other traits and does nothing on it's own
```
Mnemotechnician: This is just... no

Resources/Prototypes/Traits/neutral.yml
```diff
@@ -116,6 +120,7 @@
 - type: trait
   id: Liar
   category: Mental
+  points: 1
```
Mnemotechnician, line 123: Why does an accent give points?

Resources/Prototypes/Traits/disabilities.yml
```diff
@@ -314,7 +316,7 @@
 - type: trait
   id: Clumsy
   category: Physical
-  points: 1
+  points: 5 # Clumsy makes you unable to use guns, can't climb tables and medpen others
```
Mnemotechnician: While I agree with this reasoning, 5 points is way too much. Security and command cannot be clumsy, so it's not that important.

Resources/Prototypes/Traits/disabilities.yml
```diff
@@ -222,7 +224,7 @@
 - type: trait
   id: SnailPaced
   category: Physical
-  points: 5
+  points: 8
```
Mnemotechnician: 5 and 8 points for these is iffy. It's true that speed matters a lot in combat, but outside of combat those traits barely affect anything.

Resources/Prototypes/Traits/disabilities.yml
```diff
@@ -176,7 +178,7 @@
 - type: trait
   id: BadKnees
   category: Physical
-  points: 3
+  points: 5 # Opposite of Parkour Training and a horrific debuff for its cost
```
Mnemotechnician: This is iffy too.

Mnemotechnician (changes requested):
The PR still needs to be updated since #1060 is merged. Nonetheless, I have a lot of concerns about the new trait point values.

Mnemotechnician:
>  (altho I could just remove those and get the slot removal pushed in first)

I think this would be the best decision. While some traits could use some balacing, such radical changes would require a lot of discussion.

Resources/Prototypes/Traits/skills.yml
Raikyr0: must be an upstream thing that it doesn't work with shoes on then

Resources/Prototypes/_Floof/Traits/lewd.yml
Raikyr0: so it fills perfectly into 1 glass.

Resources/Prototypes/Traits/neutral.yml
Raikyr0: this isn't an accent, this one is actually a debuff.

Resources/Prototypes/Traits/disabilities.yml
Raikyr0: trust me, i tried to medpen someone else as a clown to save them and injected myself. this applies for hyposprays and hypopens too. this is an insane detriment to someone who i was watching play a clumsy paramed.

M3739: I'd imagine one simply has to play a different way when rocking this trait.

After taking a peek at it's code, it only makes a chance for hyposprays and medical pens to, yes, inject yourself. However, syringes are fool-proof. 

The only 3 other things it seems to affect is using a defibrillator (you can shock yourself, yes), climbing, and shooting a gun. Climbing is negligible given the prone system we possess.

In my opinion, this trait is not as detrimental as you think. You simply have to work a bit differently.

Resources/Prototypes/_Floof/Traits/lewd.yml
```diff
@@ -59,10 +62,11 @@
...
           reagents:
           - ReagentId: NaturalLubricant
-            Quantity: 25
+            Quantity: 30
```
M3739: The `maxVol` and `Quantity` changes in lines 19, 22, 65, and 68 are out of scope.

M3739 (changes requested):
The natural languages seem to be exempt from this PR due to unknown reasons.

As a language trait, it should be subject the same slot cost with the rest of it's peers.

And I agree with Mnemo that right now as it stands, the trait cost adjustments are too drastic to accept without discussion. I second the suggestion that you should just focus on pushing the slot costs.

Raikyr0:
Alright I removed the trait cost changes for now.

Resources/Prototypes/Traits/neutral.yml
M3739: image
lol

M3739 (approved):
LTGM.

PR #1066, opened 2025-06-05
Revenant Buff

Raikyr0:
# Description

Buffs to Revenant, increases their max HP and regen speed, increased number of things that they can emag, added a voice mask, added night vision, added harpy singing, allowed it to pilot mechs (they can die real easy like this), and do emotes.

Why?
They felt like the short straw on impact and RP compared to a Space Dragon. This should spice it up slightly, without making them too dangerous. (still drastically less so then a dragon)

# Changelog
:cl:
- add: Added Voice Mask, Night Vision, Singing, Mech Piloting, Emoting to Revenant.
- tweak: Buffed Revenant HP and Regen, and things they can emag.

### DIFF
--- a/Content.Shared/Revenant/Components/RevenantComponent.cs
+++ b/Content.Shared/Revenant/Components/RevenantComponent.cs
@@ -19,7 +19,7 @@ public sealed partial class RevenantComponent : Component
     /// </summary>
     [DataField, ViewVariables(VVAccess.ReadWrite)]
     [AutoNetworkedField]
-    public FixedPoint2 Essence = 75;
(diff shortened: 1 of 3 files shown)


Mnemotechnician (approved):
I don't mind this. Changing essence stats is iffy, but the voicemask and extended emag whitelist are great.

Memeji (approved):
I like the noises, voicemask, and emotes as they'll add more potential for playing out as a spooky revenant.

Essence values seem fine I think.
It'll be interesting with the expanded whitelist too.

PR #1073, opened 2025-06-09
Logistics Addition - Boards

Raikyr0:
# Description

Adds a more reliable way for Logistics to obtain a R&D board and an AmmoFab board.

Why?
The former is so Epistemics can continue the work easier after they finish their initial server. Still reliant on the fact it costs 100k.
The latter is quality of life for Salvage (and traitors), it requires Armory access to open, so you should ask your local Warden. (or just PKA it open), price is set at 50k, since like the RD board they can find one on an expedition.

Prices are placeholder.

# Changelog

:cl:
- add: R&D and Ammo Fab to Cargo's Catalog

### DIFF
--- a/Resources/Prototypes/Catalog/Cargo/cargo_circuitboards.yml
+++ b/Resources/Prototypes/Catalog/Cargo/cargo_circuitboards.yml
@@ -7,3 +7,23 @@
   cost: 2000
   category: cargoproduct-category-name-circuitboards
   group: market
+
+﻿- type: cargoProduct
+  id: RDServerBoard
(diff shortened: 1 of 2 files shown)


Memeji (approved):
I'm ok with this change, to add a way to guarantee a R&D board in case Salv luck is rolling bad.

The pricing and Sec Board could be discussed, I will leave it open for others to chime in regarding what they think there.

Mnemotechnician (approved):

ShadesMars (approved):

PR #1075, opened 2025-06-09
Admeme Tool - Dildos

Raikyr0:
# Description

Adds an Admin Dildo (for admin bits) and replaces the Immovable rod texture with a Red Rocket Dildo.

# Changelog

This does not go in the changelog, there are no player facing changes

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Fun/immovable_rod.yml
+++ b/Resources/Prototypes/Entities/Objects/Fun/immovable_rod.yml
@@ -7,7 +7,7 @@
   - type: InteractionOutline
   - type: MovementIgnoreGravity
   - type: Sprite
-    sprite: Objects/Fun/immovable_rod.rsi
+    sprite: _Floof/Objects/Fun/Lewd/Toys/Dildos/dildo_canine_red.rsi
     state: icon
     noRot: false
   - type: ImmovableRod
--- a/Resources/Prototypes/Entities/Objects/Fun/immovable_rod.yml
+++ b/Resources/Prototypes/Entities/Objects/Fun/immovable_rod.yml
@@ -7,7 +7,7 @@
   - type: InteractionOutline
   - type: MovementIgnoreGravity
   - type: Sprite
-    sprite: _Floof/Objects/Fun/Lewd/Toys/Dildos/dildo_canine_red.rsi
+    sprite: _Floof/Objects/Fun/Lewd/Toys/Dildos/dildo_canine_red.rsi # Very Obviously Floofstation
     state: icon
     noRot: false
   - type: ImmovableRod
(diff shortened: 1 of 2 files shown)


Mnemotechnician:
The dildo immovable rod sounds very LRP, in my opinion. We definitely need more opinions on that one.

Raikyr0:
> The dildo immovable rod sounds very LRP, in my opinion. We definitely need more opinions on that one.

So, it's closer to an easter egg because you get VERY little time to react and see the immovable rod.

Memeji:
While funny, and while hard to really see what it is when it tears through the station, it is going to be noticed.

Which will have odd lore ramifications and people shouting about how a dildo just ripped a hole through the station.
Or how a dildo just turned someone into organs. (Which tbh freaks me out thinking about it).
Mostly brings into question how serious or an RP are we going for when we have a dildo destroying an expensive station every shift, I suppose.

Raikyr0:
> While funny, and while hard to really see what it is when it tears through the station, it is going to be noticed.
> 
> Which will have odd lore ramifications and people shouting about how a dildo just ripped a hole through the station. Or how a dildo just turned someone into organs. (Which tbh freaks me out thinking about it). Mostly brings into question how serious or an RP are we going for when we have a dildo destroying an expensive station every shift, I suppose.

The main point of it is it's funny, but, the lore ramifications of a immovable rod is already weird. The rod is supposed to not be moving and we are, but we have an anchor. (it's an unstoppable rod at that point). Rods don't happen that often either, but yeah it's an easter egg. Perhaps it'll be funny the first time someone spots it, but it'll end up going back to the common rod reaction.

M3739:
Ubuntu tests seem to have folded in half... Again. Seems unrelated to the PR.

As for the direction... I am not too keen on the red rocket immovable rod myself.  It's a neat easter egg... Until we flip the coin and find out whether or not someone is going to take well the fact they got expediently disassembled by what appears to be the same toy they had in their bag moving at hyper velocity speeds.

One can even make the argument that getting obliterated by the regular rod is less offending than getting obliterated by a warp-speed dildo rearranging your insides across the whole station.

Mnemotechnician:
> The main point of it is it's funny, but, the lore ramifications of a immovable rod is already weird. The rod is supposed to not be moving and we are, but we have an anchor. (it's an unstoppable rod at that point).

Just because the station has an anchor doesn't mean it's "not moving". If it was not, any asteroid impact would be equivalent to a nuclear blast due to the insane relative speeds bodies around it would be moving at. The station anchor simply adjusts the station's velocity to match that of nearby celestial bodies (perhaps the mythical gas giant?).

Memeji:
Will have to close this PR as it doesn't match the direction we wanna take at the moment.

PR #1077, opened 2025-06-09
Mopping Up 

Raikyr0:
# Description

Fixed Mop swing rotation, slightly increased footprint clean radius on mops and advanced mops, added a small footprint clean radius on Cleanbots.

Why?
Janitor QoL, most of the time it's not even janitors cleaning it and just Cleanades, hell I don't even use a Mop I take a spray bottle and 4 buckets of space cleaner.

# Changelog

:cl:
- tweak: increased footprint clean radius on mops, added clean radius to cleanbots

### DIFF
--- a/Resources/Prototypes/Entities/Mobs/NPCs/silicon.yml
+++ b/Resources/Prototypes/Entities/Mobs/NPCs/silicon.yml
@@ -228,6 +228,8 @@
     flavorKind: station-event-random-sentience-flavor-mechanical
   - type: Absorbent
     pickupAmount: 10
+    footprintCleaningRange: 0.5 # Floof - to make janitors lives easier
+    maxCleanedFootprints: 5 # Floof
   - type: UseDelay
     delay: 2
   - type: SolutionRegeneration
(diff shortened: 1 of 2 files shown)


Mnemotechnician:
Two things:
- Cleanbots already have default footprint clean radius, and this change makes their cleaning radius about 1.5 higher and therefore their cleaning area 2.25 times greater.
- Increasing the area of effect of mops 4-fold is a way too insane buff.

Raikyr0:
> Two things:
> 
>     * Cleanbots already have default footprint clean radius, and this change makes their cleaning radius about 1.5 higher and therefore their cleaning area 2.25 times greater.
> 
>     * Increasing the area of effect of mops 4-fold is a way too insane buff.

I can lower it, this is more so just to remove footprints faster.

Memeji (approved):
Numbers feel ok now to me, will let others chime in for their thoughts

Mnemotechnician (changes requested):
The existing numbers were balanced around the idea of "one mop hit should clean roughly one tile worth of footprints". This change completely breaks that balance and makes mops magically clean footprints way outside of that 1 tile radius

Raikyr0:
> The existing numbers were balanced around the idea of "one mop hit should clean roughly one tile worth of footprints". This change completely breaks that balance and makes mops magically clean footprints way outside of that 1 tile radius

This is intended to be a quality of life improvement for when every single hallway is soaked in blue acid blood and people walking through it creating hundreds of footsteps and there's only 1 HoP doing janitor work. 

You could say hire more Janitors but that becomes a player issue when no one wants to step up. Or even use cleanades (since we can buy dedicated boxes of em), at that point the buff seems perfectly fine.

PR #1078, opened 2025-06-09
Bootleg Surgery+

Raikyr0:
# Description

Adds Surgery Tool lines to advanced tools since they were lacking those over normal tools, they're as fast as normal surgery tools due to their powered nature, but still worse then advanced surgery tools.

Also added welders use speed for experimental welder because advanced industrial had a speed bonus but the experimental didn't. 

# Changelog

:cl:
- add: Added Surgery Tool tags to advanced tools
- add: Added use speed bonus to Experimental Welder

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Tools/welders.yml
+++ b/Resources/Prototypes/Entities/Objects/Tools/welders.yml
@@ -216,7 +216,7 @@
       - ReagentId: WeldingFuel
         Quantity: 1
   - type: Tool
-    speedModifier: 1.3 # it being faster is fine
+    speedModifier: 1.0 # it being faster is fine
     # Floof Shitmed - Improvised Tools
   - type: Cautery
     speed: 1.0
--- a/Resources/Prototypes/Entities/Objects/Tools/welders.yml
+++ b/Resources/Prototypes/Entities/Objects/Tools/welders.yml
@@ -216,7 +216,7 @@
       - ReagentId: WeldingFuel
         Quantity: 1
   - type: Tool
-    speedModifier: 1.5 # For whatever reason this item didn't have this when an advanced industrial welding tool did
+    speedModifier: 1.3 # it being faster is fine
     # Floof Shitmed - Improvised Tools
   - type: Cautery
     speed: 1.0
--- a/Resources/Prototypes/Entities/Objects/Tools/welders.yml
+++ b/Resources/Prototypes/Entities/Objects/Tools/welders.yml
@@ -219,7 +219,7 @@
     speedModifier: 1.5 # For whatever reason this item didn't have this when an advanced industrial welding tool did
     # Floof Shitmed - Improvised Tools
   - type: Cautery
-    speed: 1.3
+    speed: 1.0
   - type: SurgeryTool
     startSound:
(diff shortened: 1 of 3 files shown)


Resources/Prototypes/Entities/Objects/Tools/welders.yml
```diff
@@ -199,6 +215,16 @@
       reagents:
       - ReagentId: WeldingFuel
         Quantity: 1
+  - type: Tool
+    speedModifier: 1.5 # For whatever reason this item didn't have this when an advanced industrial welding tool did
```
M3739: > "# For whatever reason this item didn't have this when an advanced industrial welding tool did"

The reason in question is because the experimental welder does not mention having a hotter flame anywhere in it's description.
image
image
The resulting change will also make it attractive to IPCs, as they can patch themselves up more efficiently without consulting any robotics or engi personnel. Which ideally, we want to keep such as a matter of necessity more than convenience.

M3739: For design reference, the experimental welder predates the advanced industrial in the Wizden repo. And the advanced industrial welder was designed to be unique in that it is faster than other welders.

You can look up Wizden PR No. 2453 for the experimental welder, and Wizden PR No. 7855 for the advanced industrial welder for reference.

Resources/Prototypes/Entities/Objects/Tools/welders.yml
```diff
@@ -199,6 +215,16 @@
...
+    # Floof Shitmed - Improvised Tools
+  - type: Cautery
+    speed: 1.3
```
M3739: Refer to the comment on the `speedModifier`.
```suggestion
    speed: 1.0
```

M3739 (changes requested):

Resources/Prototypes/Entities/Objects/Tools/welders.yml
```diff
@@ -199,6 +215,16 @@
       reagents:
       - ReagentId: WeldingFuel
         Quantity: 1
+  - type: Tool
+    speedModifier: 1.3 # it being faster is fine
```
M3739: **The direction has been decided that we want to keep the advanced industrial welder to be the only welder with the faster welding speed.**

The experimental welder already has self-recharging, and does not need a faster welding speed. The microreactor, for example, self-charges but only has the capacity of a medium power cell.

Experimental does not mean it is a direct upgrade. It is an experiment, an idea to try out. And that idea, is, self-refueling. Nothing more, and nothing less at this moment. We intend to to preserve the choice between either wanting faster welding speed, or nigh-infinite fuel. Not both in the same tool.

**Remove the welding speed increase.** Other than that, everything else checks out in this PR.

M3739 (changes requested):

M3739 (approved):
At this point, the `Tool` component on the experimental welder is now redundant as it will inherit the same component and it's values from it's parent `Welder.`

Functionality wise, everything now checks out.

Memeji (approved):
Looks good on my end as well.

ShadesMars (approved):

PR #1079, opened 2025-06-09
Slide Papers and Mails Through Doors

Superlagg:
# Description

Papers (and anything descending of paper), envelopes, playing cards, and non-package mail no longer collide with doors. Basically if a mouse can go through it, bureaucracy can too.

Now you can slide notes under doors, and deliver mail to people banging in dorms! Send people secret admirer notes! Hand out surveys about their current dorms experiences! Have all your important paperwork fly out the airlock!

About that last part, folders and such *dont* go through doors, so if you dont want those Vital Secret Documents just flying out into space in that one in a billion situation where something launches it into a door leading out into space, just put it in a folder or something!

---

# Changelog

:cl: Superlagg
- tweak: Papers, Envelopes, Playing Cards, and Small Mail can now be slid under doors! You can do this by just tossing it through the door, kinda like mice!

### DIFF
--- a/Content.Shared/Physics/CollisionGroup.cs
+++ b/Content.Shared/Physics/CollisionGroup.cs
@@ -70,6 +70,8 @@ public enum CollisionGroup
     // Soap, spills
     SlipLayer = MidImpassable | LowImpassable,
     ItemMask = Impassable | HighImpassable,
+
+    ItemSlideUnderDoorMask = Impassable | LowImpassable, // FLOOF ADD: Allows some items to slide under doors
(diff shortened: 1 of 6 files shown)


M3739 (approved):
Apologies for the wait.

After testing, appears to work as intended.

LGTM

Mnemotechnician (approved):

Mnemotechnician:
For future reference, a more clean way to do this would be to implement an abstract parent containing just the Fixtures component and add it as the first parent to all affected entities. But this works too.

PR #1080, opened 2025-06-10
Pipe Bomb in Your Mail

ItsDzordz:
Zrzut ekranu 2025-06-10 023933

# Description
This pr allows small explosives, like grenades, c4 and most importantly, pipe bombs to be put in the envelopes.
Adds a new tag for items that can fit in the envelopes, called "EnvelopeFiller"

---

# TODO

- [x] Enjoy

---

<p>

image 

</p>
</details>

---

# Changelog

:cl:

- tweak: Allowed small explosives to be put inside of the envelopes.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Misc/paper.yml
+++ b/Resources/Prototypes/Entities/Objects/Misc/paper.yml
@@ -806,6 +806,7 @@
         whitelist:
           tags:
             - Paper
+            - EnvelopeFiller
   - type: ActivatableUI
     key: enum.PaperUiKey.Key
     requiresComplex: false
--- a/Resources/Prototypes/Entities/Objects/Weapons/Bombs/plastic.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Bombs/plastic.yml
@@ -34,6 +34,9 @@
         base:
           Primed: { state: primed }
           Unprimed: { state: complete }
+  - type: Tag
+    tags:
+    - EnvelopeFiller
 
 - type: entity
   name: composition C-4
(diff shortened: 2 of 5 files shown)


Cprn-Sprkl:
There is another PR that is making it so envelopes have no door collision: https://github.com/Fansana/floofstation1/pull/1079

Unless there is some tag preventing it; these two being merged to the server would make it so grenades, C4, and pipe bombs can also be tossed through locked doors by simply sticking them in an envelope first.

Memeji:
> There is another PR that is making it so envelopes have no door collision: #1079
> 
> Unless there is some tag preventing it; these two being merged to the server would make it so grenades, C4, and pipe bombs can also be tossed through locked doors by simply sticking them in an envelope first.

Good catch, should think on this and maybe make it so if it is filled with an explosive it can't go under a door.

M3739:
> > There is another PR that is making it so envelopes have no door collision: #1079
> > Unless there is some tag preventing it; these two being merged to the server would make it so grenades, C4, and pipe bombs can also be tossed through locked doors by simply sticking them in an envelope first.
> 
> Good catch, should think on this and maybe make it so if it is filled with an explosive it can't go under a door.

I second this. Having explosives be wrapped in a envelope before vibe-checking whoever or whatever is on the other side of the door is less than ideal.

ItsDzordz:
> > There is another PR that is making it so envelopes have no door collision: #1079
> > Unless there is some tag preventing it; these two being merged to the server would make it so grenades, C4, and pipe bombs can also be tossed through locked doors by simply sticking them in an envelope first.
> 
> Good catch, should think on this and maybe make it so if it is filled with an explosive it can't go under a door.

Well i dont know how to code that in, if you want it to be this way you will have to code it in yourself.

Superlagg:
the way my PR works is that it just changes the fixture on those specific items. If theres a way to modify the collision flags on the fly, that'd be perfect for this, just have the envelope check if the thing could explode, then modify the flags

ghost:
Gonna deny this for now as it has issues mentioned previously, a more realistic option would be adding packages that can not be shoved under doors. Adding all these checks just makes it to complicated.

PR #1083, opened 2025-06-11
Succubus Trait

Raikyr0:
# Description

Adds the Succubus Trait. It uses reverse-engineered Liquor Lifeline code.
In practice, similar to the Vampire metabolism of blood, but heals less, without the food restriction (in case you don't want to be caught chugging cum in public when someone invites you to eat).

Alright, i've redone alot of shit, instead of using old vampire ideas like converting to a chemical, it simply skips the converting process and does it.

# TODO

- [ ] The Balancing Part
- [x] The Coding Part

# Changelog

:cl:
- add: Added Succubus Trait


Raikyr0:
Just running the checks before putting it back into draft for balancing and bug fixing. (Surely the build and test debug will show me the actual issues instead of failing 62 times to unrelated issues)

M3739:
Alright, so I tested it out a bit, and it appears that it only seems to result in a blood level increase, but no healing anywhere to be found.

The saline part is coming through, and so is the vitamins, but the omnizine portion is not coming up.

Raikyr0:
> Alright, so I tested it out a bit, and it appears that it only seems to result in a blood level increase, but no healing anywhere to be found.
> 
> The saline part is coming through, and so is the vitamins, but the omnizine portion is not coming up.

that's really weird, maybe there's too many chemicals being processed?

Raikyr0:
Ok, I added a chemical called Succubium, (tad on the nose I know) which combines the effects of the 3 chemicals so it doesn't clog the bloodstream.

M3739:
Gonna test this again later on.

M3739 (changes requested):
After some testing,  it now doesn't seem to work at all. No longer raises blood levels.
image

Raikyr0:
> After some testing, it now doesn't seem to work at all. No longer raises blood levels. image

Not sure what's wrong, could you inject the succubus chemical instead of cum? It's either the chemical that's not working. Or the conversion process.

M3739:
> Not sure what's wrong, could you inject the succubus chemical instead of cum? It's either the chemical that's not working. Or the conversion process.

Ill test that next.

M3739:
> Ill test that next.

The Succubium works as intended. Pretty good at raising blood levels, actually. It only heals ever so slightly.
image

Raikyr0:
> > Ill test that next.
> 
> The Succubium works as intended. Pretty good at raising blood levels, actually. It only heals ever so slightly. 

Yeah it's not supposed to heal a lot.

M3739 (changes requested):
Still does not work, unfortunately...
image
Here's a bit of a pointer. the `!type:OrganType` condition looks for a `MetabolizerType`  in a metabolizing organ.

That being said, we have an in-house trait modify type called `!type:TraitModifyMetabolism`, courtesy of @Superlagg and his #707 that may be useful to achieve your ends. You can read more about it in this file located [here](https://github.com/Fansana/floofstation1/blob/0e86f73bda61ae9a2a3ab7177b4b882babb7550c/Content.Server/_Floof/Traits/TraitSystem.Functions.cs#L16-L37).

Raikyr0:
> Still does not work, unfortunately... image Here's a bit of a pointer. the `!type:OrganType` condition looks for a `MetabolizerType` in a metabolizing organ.
> 
> That being said, we have an in-house trait modify type called `!type:TraitModifyMetabolism`, courtesy of @Superlagg and his #707 that may be useful to achieve your ends. You can read more about it in this file located [here](https://github.com/Fansana/floofstation1/blob/0e86f73bda61ae9a2a3ab7177b4b882babb7550c/Content.Server/_Floof/Traits/TraitSystem.Functions.cs#L16-L37).

I'll try the  - !type:TraitModifyMetabolism if it works I might have to clean up the excess files

M3739 (changes requested):
Still doesn't work. Perhaps try adding the `add: true` field to the `TraitModifyMetabolism`?

M3739 (commented):
Happy to report it works as intended following the latest commit.

Just do some clean up, then we can discuss balancing.

M3739 (approved):
Functionality is confirmed, and it's working as intended.

With how scarce the actual thing is in comparison to blood, I'd reckon it's okay in it's current state. While allowing for some interesting scenarios.

Ill invite discourse from the other maintainers to ensure the balancing part, as many pairs  of eyes are better than one.

Mnemotechnician (approved):
i don't mind this so long as it actually works. However, one thing to look out for is the possibility of people powergaming with this trait by e.g. constantly carrying a bottle-of-this-thing and chugging from it to regenerate, but I guess that also applies to vampires.

M3739:
Bruhg

M3739:
Alright, this has been sitting for a long while. Merging.

ashkitten:
> one thing to look out for is the possibility of people powergaming with this trait by e.g. constantly carrying a bottle-of-this-thing and chugging from it to regenerate, but I guess that also applies to vampires.

Something that could be considered for the future is adding DNA to more bodily fluids than just blood, and then having the trait only work if there's DNA present. However, I suspect that would not be appreciated by some people who want more "vegan" characters.

PR #1092, opened 2025-06-12
Upstream/1605: Replace Tech Lockout With Softcap

### LINKED ISSUE #1091: Incremental points instead of tier locking
#Description
This issue aims to address the boring way research unlocks currently work—specifically, the fact that EPI can only ever research one tier 3 technology without acquiring another server.

We want to remove the restriction on research, allowing EPI to potentially unlock all technologies using a single server. However, each additional technology would cost more than the last. This means EPI will need to think strategically about which technologies are most valuable early on and which ones can be delayed or skipped.

An additional feature I just came up with is the creation of a button that allows players to reroll available technologies. This would let players pay a smaller cost to potentially get better research options, without risking the increased point cost associated with unlocking a new technology.

Mnemotechnician:
# Description
See https://github.com/Simple-Station/Einstein-Engines/pull/1605

Resolves #1091

Also adds softcaps to floofstation things that were missing them, and adjusts softcap multipliers on some low-utility things.

image

# Changelog
:cl: Upstream
- remove: Removed the "Tier 3 Tech Lockout" mechanic. You are no longer limited to 1 discipline worth of t3 research.
- add: Research is now "Softcapped" by Tier 3 unlocks. Each unlocked Tier 3 technology multiplicatively increases the cost of all other research. Rushing a capstone can be quite expensive, as will be getting multiple capstones. I hope you're glad that you can build more than one prober now without guaranteeing the station will explode.


Raikyr0:
This is a good change to see, but please also consider https://github.com/Simple-Station/Einstein-Engines/pull/1603 and https://github.com/Simple-Station/Einstein-Engines/pull/1602 as well.

PR #1096, opened 2025-06-13
LPO Inherits Character Appearance

Raikyr0:
# Description

Adds a line that should allow Listening Post Operatives to inherit their character profile appearance

# Changelog

:cl:
- tweak: Tweaked it so LPOs spawn as your character


M3739:
The build test failing after 30 minutes is wild. Also a test crashing on itself.

Build fail is unrelated to this PR.

Memeji:
Re-running failed jobs

M3739:
After a quick test, it would appear that it would not work as expected.

image

Raikyr0:
> After a quick test, it would appear that it would not work as expected.
> 
> image

I love figuring out why it doesn't work.

PR #1098, opened 2025-06-14
100% Orange Juice - Commission PR

Raikyr0:
# Description

PR Commission created for Karin, for the cost of one half eaten sandwich, one eternal love, and a possible art commission.
Completely unlikely to be merged unless she bribes all the maintainers.

Adds a slime only trait, that changes the bloodstream to orange juice, breast milk to orange juice, and adds the reagent slime on hit effect which injects the bloodstream (orange juice).

Only setting to open to see if the Milk Producer component works with reagents other then milk.


CloneTrooperSimp:
Hello! Karin's player here, and I just wanted to say some things.
I dunno if this PR is being worked on or will be worked on eventually, but that second part, about the reagent slime on hit effect, is entirely not needed, and if it needs to be removed for the PR to be approved, I do not mind it at all. It would be majorly fun to have and very accurate to the character, but it's not the primary focus of why I asked for this PR to be made.
Second of all, I still owe that sandwich and that art, the eternal love is already there, but the items will be delivered, I promise!

PR #1099, opened 2025-06-14
An Eggcellent Idea

Raikyr0:
# Description

This (idea) came to me in a (honestly fever) dream, now you can take a trait to lay eggs.

# Changelog

:cl:
- add: Added Egg Laying trait


Resources/Prototypes/_Floof/Traits/lewd.yml
```diff
@@ -63,3 +63,22 @@
           reagents:
           - ReagentId: NaturalLubricant
             Quantity: 25
+
+- type: trait
+  id: EggLayer
+  category: Physical
+  points: 0
```
M3739, line 70: As it turns out, you can feed yourself with this, especially if you have an animal stomach. If anything, this is worth way more than 0 points.

Resources/Prototypes/_Floof/Traits/lewd.yml
```diff
@@ -63,3 +63,22 @@
...
+  requirements:
+    - !type:CharacterJobRequirement # Weird if a borg could lay an egg
+      inverted: true
+      jobs:
+        - Borg
+        - MedicalBorg
```
M3739, line 76:
```suggestion
    - !type:CharacterJobRequirement # Weird if a borg could lay an egg
      inverted: true
      jobs:
        - Borg
        - MedicalBorg
    - !type:CharacterSpeciesRequirement
      inverted: true
      species:
        - IPC
```

Resources/Prototypes/_Floof/Traits/lewd.yml
```diff
@@ -63,3 +63,22 @@
           reagents:
           - ReagentId: NaturalLubricant
             Quantity: 25
+
+- type: trait
+  id: EggLayer
```
M3739, line 68: You are missing the FTLs for the trait itself: 
image

Resources/Prototypes/_Floof/Actions/egg.yml
```diff
@@ -0,0 +1,10 @@
+- type: entity
+  id: ActionLayEgg
+  name: Lay egg
+  description: Lays an unfertilized egg for use in cooking.
+  categories: [ HideSpawnMenu ]
+  components:
+  - type: InstantAction
+    icon: _Floof/Interface/Actions/egg.png
+    event: !type:SericultureActionEvent
```
M3739, line 9: Seems like this still uses the same message for creating webs. A unique message should be implemented.
image

Resources/Prototypes/_Floof/Traits/lewd.yml
```diff
@@ -63,3 +63,22 @@
...
+        - Borg
+        - MedicalBorg
+  functions:
+    - !type:TraitAddComponent
+      components:
+        - type: Sericulture
+          action: ActionLayEgg
+          productionLength: 1
+          entityProduced: FoodEgg
+          hungerCost: 2
```
M3739: As hilarious that immediately laying 30+ eggs in a single sitting is without effort, egg-laying is orders of magnitude more demanding on the body and longer to do than spiders making webs.

Adjust accordingly. Afterall, your basically pushing unfertilized zygotes contained in an organic vessel. Such is an egg.

Resources/Prototypes/_Floof/Traits/lewd.yml
M3739: When writing the FTL, know that animals that lay eggs are known to be **oviparous**. You may also want to mention the egg having a rigid shell to reduce confusion with other types of animal eggs.

M3739 (changes requested):
We love ideas that come to us in a dream. It functions, and works, but some things need to be fixed and revised before it could set sail.

Resources/Prototypes/_Floof/Traits/lewd.yml
Raikyr0: You can technically feed yourself with the breast trait too, and to some extent, the cock trait.

Resources/Prototypes/_Floof/Traits/lewd.yml
Raikyr0: I added an FTL, 

Your body through either mutation, oviposition, cybernetic or noospheric influence has the ability to lay eggs.

Resources/Prototypes/_Floof/Traits/lewd.yml
Raikyr0: FTL covers this,

Resources/Prototypes/_Floof/Actions/egg.yml
Raikyr0: Fixed this by making the popup text more generic and not mention silk, to cover my bases later when i use it again.

Resources/Prototypes/_Floof/Traits/lewd.yml
M3739: > FTL covers this,

No. As in **IPCs function similarly to borgs.** It isn't weird, it's **functionally broken.**

The sericulture code was originally intended and designed to allow you to make webs at the cost of your own current hunger. However, the code lacks a check to ensure you actually have the hunger component to begin with. And so in a test run, I noticed I could make an unlimited number of eggs, as long as I have power, I could lay eggs endlessly.
I count 614 eggs until my battery died.

Which we don't want because making entities is expensive resource wise. One can imagine making thousands of these on a micro reactor, then grinding the server to a crawling halt by getting space wind to blow it all to high hell.

image

Just like how I restricted the AAC tablet from mimes because the code for it will literally make a mime break their oath as soon as they select a word, this is to save us some trouble, and prevent us from having to deliver a hot fix in the future.

Raikyr0: Ok yeah if they functionally have an issue i'll fix that.

Resources/Prototypes/_Floof/Traits/lewd.yml
M3739: While true, it is on a cooldown and a fixed amount. Compared to laying several, then waiting a bit, then laying several more, as long as you have food to eat, of course.

I do notice that the bionic web augment got reduced down to 2 points... So if anything, this costing a measly 1 point would make sense. Eggs are much more versatile and satiating than the other two, not counting recipes.

Raikyr0: I've increased the do-over time and cooldown.

As for the Spinnerette Augment, upstream T3 Arsenal Jugg Suits require silk to make. (and a incredibly large amount of tungsten) For the most part, I think it's a niche, huh you lay eggs?

M3739: After a second test, all checks out.

Also did not know that! Good to know.

M3739 (approved):
After the tweaks and changes, now this is ready to set sail.

Just waiting for the other maintainers to review this.

ShadesMars (approved):

VividPups:
Are you all are aware that not all species can lay eggs?

PR #1102, opened 2025-06-15
Fix Roundstart Loop Caused by Invalid Loadouts

Mnemotechnician:
# Description
If the loadout system tries to spawn severeal exclusive items where the second item overrides the first, the first item will end up deleted. It however will still be returned from ApplyCharacterLoadout, and the loadout system will later try to paint the already-deleted item, causing an exception to be thrown and round start to be aborted.

This fixes it by checking whether the item actually exists and adding a try-catch statement.

## This will need some basic testing in release mode because the change was web-editted

# Changelog
:cl:
- fix: The roundstart loop bug should now be fixed.


Mnemotechnician:
So yeah, I reproduced it, the fix seems to work

PR #1104, opened 2025-06-17
Temperature Tolerance Tweak

Raikyr0:
# Description

So, before this trait didn't really resist shit before, was still freezing your ass off on Glacier despite supposedly giving you the same cold resist as a winter coat. I'm bumping the number up slightly so you can walk naked outside on Glacier.

# Changelog

:cl:
- tweak: Tweaked Temperature Tolerance trait to actually resist more cold

### DIFF
--- a/Resources/Prototypes/DeltaV/Entities/Mobs/Species/vulpkanin.yml
+++ b/Resources/Prototypes/DeltaV/Entities/Mobs/Species/vulpkanin.yml
@@ -84,7 +84,7 @@
   - type: Damageable
     damageModifierSet: Vulpkanin
   - type: TemperatureProtection
-    coefficient: 0.1
+    coefficient: 0.05
   - type: Vocal
     sounds:
       Male: MaleVulpkanin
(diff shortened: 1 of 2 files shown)


Mnemotechnician:
No, 0.1 is the temperature protection coefficient of a winter coat, and is what other species like vulps have. If you are changing this, you must also change the protection coefficients of other species.

Nevertheless, there is also a different but linked issue related with temperature regulation. All species have the same ThermalRegulatorComponent, which means they produce & lose the same amount of heat passively (through shivering and sweating). But smaller mobs hvae less mass and as such less heat capacity. This means that a 7 kg rodentia or harpy experiences 10 times more effective passive heating than a 70 kg human, allowing them to survive in extreme cold, which likely has led to this misleading comment.

Raikyr0:
> No, 0.1 is the temperature protection coefficient of a winter coat, and is what other species like vulps have. If you are changing this, you must also change the protection coefficients of other species.

I ain't about to touch the ThermalRegulatorComponent with a 12 foot pole so, I'll just get the Vulp numbers changed.

Mnemotechnician:
Second point will be fixed in https://github.com/Fansana/floofstation1/pull/1107

ghost:
Yeah, not gonna merge this, will be fixed with #1107

PR #1105, opened 2025-06-17
Oracle Rewards Normality

Cprn-Sprkl:
# Description

Adds normality crystals to the oracle reward pool. Same weights as their bluespace counterparts.

---

# Changelog

:cl: sprkl
- tweak: The Oracle can now grant normality crystals when given offerings.

### DIFF
--- a/Resources/Prototypes/Entities/Structures/Specific/oracle.yml
+++ b/Resources/Prototypes/Entities/Structures/Specific/oracle.yml
@@ -80,8 +80,11 @@
   id: OracleRewardCrystals
   weights:
     MaterialBluespace1: 3
+    MaterialNormality1: 3
     MaterialBluespace3: 10
+    MaterialNormality3: 10
     MaterialBluespace5: 2
+    MaterialNormality5: 2
 
 - type: weightedRandom
   id: OracleRewardReagents
--- a/Resources/Prototypes/Entities/Structures/Specific/oracle.yml
+++ b/Resources/Prototypes/Entities/Structures/Specific/oracle.yml
@@ -80,11 +80,9 @@
   id: OracleRewardCrystals
   weights:
     MaterialBluespace1: 3
-    MaterialNormality1: 3
     MaterialBluespace3: 10
-    MaterialNormality3: 10
     MaterialBluespace5: 2
-    MaterialNormality5: 2
+    MaterialNormality1: 5
 
 - type: weightedRandom
   id: OracleRewardReagents


Mnemotechnician (changes requested):
That is extremely overpowered. Epistemics will get on average 1.43 crystals per offering, which means on average they will get 1 drainer every 4 offerings. Big no.

I'm an epistemics main, and I can tell you, when someone actualy works on fulfilling oracle requests, epistemics can get a few dozen requests completed in a 2 hour shift.

Cprn-Sprkl:
> That is extremely overpowered. Epistemics will get on average 1.43 crystals per offering, which means on average they will get 1 drainer every 4 offerings. Big no.
> 
> I'm an epistemics main, and I can tell you, when someone actualy works on fulfilling oracle requests, epistemics can get a few dozen requests completed in a 2 hour shift.

Do you have a recommendation on new weights? Maybe one normality crystal on the rare weight and no other? That'd be maybe one set of five in a couple dozen requests.

This PR was made because of issues with epistemics passing the buck for glimmer control on to cargo and botany. This would also make epi pay more attention to the oracle because I very rarely see it engaged with.

Raikyr0:
> That is extremely overpowered. Epistemics will get on average 1.43 crystals per offering, which means on average they will get 1 drainer every 4 offerings. Big no.
> 
> I'm an epistemics main, and I can tell you, when someone actualy works on fulfilling oracle requests, epistemics can get a few dozen requests completed in a 2 hour shift.

This is my thoughts exactly when I saw the PR, like hmm, they'll get drains way too easily. While I like QoL buffs, this might be too much, the number should be more then halved.

Mnemotechnician (approved):
Okay, the new rate seems to be 0.25 crystals per offering, which I can mostly agree with

Cprn-Sprkl:
I put the full reasoning in the commit, but forgot that those are relatively hidden by default. Just going to mirror it here:

"Sets it so a normality crystal will be granted on average one out of every four offerings. This means an average of one drainer every twenty offerings, or an average normality per offering of 0.25.

This seems like a reasonable amount for the station to band together and knock out in code white situations. The rate of completion early on will be hindered by the oracle asking for unresearched options and plant mutations. As the station becomes more advanced, requests will be easier to fulfill, allowing for better drainer gain later in a shift."

M3739 (approved):
This should hopefully provide a solution to the rare situation where epistemics never gets any normality, and glimmer is at an all time high. I have personally been in a few rounds where a glimmer meltdown was practically unavoidable as a result of either very bad luck, or poor cooperation.

Cprn-Sprkl:
How do I go about deciphering this test fail? Is there an extra tool I can run to check things on my end? Is there a place where it more obviously tells me what failed?

If I'm reading this right at all it seems 86 things failed in this one test, and it is something to do with a pool system and pairings.

Mnemotechnician:
Heisentest, rerunning.

PR #222, opened 2024-10-03
Fixes Leporazine Temperature Shock

XavierSomething:
# Description

Adds a 10 degree buffer to leporazine's temperature effects so it doesn't wildly heat and cool you down in a loop.

Leporazine is a drug meant for stabilization of a body's temperature, but it has a tendency to heat someone way higher, and when cooling the body, it would trigger the heating effect of the chemical again, looping until the chemicals effect ends, where it's likely that you're left on a high temperature. Adding this buffer allows you to stay around 298k temp without being shocked all the way to 344k constantly every time you are cooled back down from lepo.

---

# TODO

- [x] Tested Lepo in-game
---

# Changelog

:cl: Xavier
- tweak: Lepo now has a 10 degree buffer between it's heating and cooling effect to prevent looping temperature shocks.

PR #1107, opened 2025-06-17
Thermal Regulator Affected by Mass

Mnemotechnician:
# Description
The speed at which atmos-exposed mobs exchange heat with the atmosphere depends on the mass of the mob (specifically, their heat capacity, which in turn is specific heat times their mass), but their thermal regulation (loss and gain of heat through sweating, radiation, shivering, and other ways) is not. This fixes this issue by introducing a system to adjust the mob's ThermalRegulationComponent stats based on their size multiplier.

Also fixes the bug where BloodstreamAdjustSystem would not account for multiple height adjusts in a row, resulting in the multiplier stacking.

TODO
- [X] Do this
- [X] Test this

<p>

image

</p>
</details>

# Changelog
:cl:
- fix: The speed at which your characters body regulates its heat now depends on its mass. This means that 4 kg rodentia will no longer regulate its temperature 17 times faster than a 70 kg human.

### DIFF
--- a/Content.Server/Body/Components/ThermalRegulatorComponent.cs
+++ b/Content.Server/Body/Components/ThermalRegulatorComponent.cs
@@ -1,10 +1,11 @@
+using Content.Server._Floof.HeightAdjust;
 using Content.Server.Body.Systems;
 using Robust.Shared.Serialization.TypeSerializers.Implementations.Custom;
 
 namespace Content.Server.Body.Components;
 
 [RegisterComponent]
-[Access(typeof(ThermalRegulatorSystem))]
+[Access(typeof(ThermalRegulatorSystem), typeof(ThermalRegulatorAdjustSystem))] // Floof - friends
 public sealed partial class ThermalRegulatorComponent : Component
 {
     /// <summary>
--- a/Content.Server/HeightAdjust/BloodstreamAdjustSystem.cs
+++ b/Content.Server/HeightAdjust/BloodstreamAdjustSystem.cs
@@ -34,7 +34,7 @@ public sealed class BloodstreamAdjustSystem : EntitySystem
 
         var bloodSolution = bloodSolutionEnt.Value.Comp.Solution;
 
-        var factor = Math.Pow(_contests.MassContest(ent, bypassClamp: true, rangeFactor: 4f), ent.Comp.Power);
+        var factor = Math.Pow(_contests.MassContest(ent, bypassClamp: true, rangeFactor: 4f), ent.Comp.Power) / ent.Comp.OldFactor; // Floof - oldFactor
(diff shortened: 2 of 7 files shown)


M3739:
Hopefully this would resolve the heat medipens being dangerous to smaller folk? I know a many who has burnt up as a result of that medipen working too well on their small bodies.

Mnemotechnician:
> Hopefully this would resolve the heat medipens being dangerous to smaller folk? I know a many who has burnt up as a result of that medipen working too well on their small bodies.

No, that's unrelated to thermal regulation. If anything, it will make it a bit worse. The issue is that leporazine adds a fixed amount of heat to the body, which, when divided by the low heat capacity of a 4 kg rodentia (which is 17 times lower than that of a human, mind you), heats them up past the safe threshold.

M3739:
> > Hopefully this would resolve the heat medipens being dangerous to smaller folk? I know a many who has burnt up as a result of that medipen working too well on their small bodies.
> 
> No, that's unrelated to thermal regulation. If anything, it will make it a bit worse. The issue is that leporazine adds a fixed amount of heat to the body, which, when divided by the low heat capacity of a 4 kg rodentia (which is 17 times lower than that of a human, mind you), heats them up past the safe threshold.



> > Hopefully this would resolve the heat medipens being dangerous to smaller folk? I know a many who has burnt up as a result of that medipen working too well on their small bodies.
> 
> No, that's unrelated to thermal regulation. If anything, it will make it a bit worse. The issue is that leporazine adds a fixed amount of heat to the body, which, when divided by the low heat capacity of a 4 kg rodentia (which is 17 times lower than that of a human, mind you), heats them up past the safe threshold.

Ah. I see. Didn't know it was fixed. I always thought it was to equalize the body temperature.

Learn something new everyday!

Mnemotechnician:
> Ah. I see. Didn't know it was fixed. I always thought it was to equalize the body temperature.
> 
> Learn something new everyday!

*Is* it already fixed? I know either EE- or floostation- side someone submitted a bandaid fix in the form of lowering the amount of heat added by leporazine, but I'm not sure if it worked for ultra-small crewmembers. (But honestly, the real fix would be preventing any species from going under ~10 kg. Too many systems break in different ways when you weight less than a stack of spesos)

M3739:
> _Is_ it already fixed? I know either EE- or floostation- side someone submitted a bandaid fix in the form of lowering the amount of heat added by leporazine, but I'm not sure if it worked for ultra-small crewmembers. (But honestly, the real fix would be preventing any species from going under ~10 kg. Too many systems break in different ways when you weight less than a stack of spesos)

Not that I know of. I would have to search for such.

M3739:
> Not that I know of. I would have to search for such.

There is #222 which is temperature shock. But I don't know if this came before or after the height system.

Memeji (approved):

PR #1109, opened 2025-06-18
Freezers and Fireplace

Raikyr0:
# Description

Makes Freezers, Fireplaces
The former has long been requested, and the latter 2 are decorations.

The Decorative Computer is not working as intended and I'll get around to fixing it later. For now I'll just remove it.

# Changelog
:cl:
- add: Added Freezer Recipes, Fireplace Recipe


Resources/Prototypes/_Floof/Recipes/Crafting/computer.yml
```diff
@@ -0,0 +1,26 @@
+- type: construction
+  name: DecorativeComputer
+  id: DecorativeComputer
+  graph: DecorativeComputer
+  startNode: start
+  targetNode: DecorativeComputer
+  category: construction-category-furniture
+  description: A computer for putering.
+  icon: Structures/Machines/computers.rsi
```
M3739: Test fail is valid. Did you write this properly?
```
Error Message:
   CLIENT: 1.815s [ERRO] res: Exception while loading resource Robust.Client.ResourceManagement.TextureResource at '/Textures/Structures/Machines/computers.rsi', resorting to fallback.
```

M3739 (changes requested):
So far, I approve of the direction of this PR. As of now, there is no way to obtain a freezer, should one be destroyed or otherwise rendered unusable, and the decorative stuff is nice for building.

Just need something addressed.

Resources/Prototypes/_Floof/Recipes/Crafting/computer.yml
Raikyr0: I'm not actually sure what's wrong there.

M3739: **The problem is that you are referring to a whole folder instead of a singular file.**

RSIs aren't exactly files despite the name they bear seeming extremely similar to a name with a file extension.

Lets take an RSI already present and done correctly in your project as an example: `fireplace.rsi`:
image
Every RSI consists of PNG files, which are the sprites, and a JSON file. The JSON file tells the game how to use these sprites.
image
the "name" must match the file name of the sprite itself. so in this case, `fireplace` refers to `fireplace.png`.
Ergo,
image
Where the "state" must match the "name".

**Referring to an RSI in this manner, is as vague as asking someone to simply find *the* box in a warehouse.** An icon is not multiple files, it is a singular file that serves as the icon of the respective prototype. **The system is expecting a reference to a singular sprite, but instead was given an entire RSI with no pointer towards a specific sprite.** And thus, throws an exception to be handled and output the following error message.

Raikyr0: I've fixed it, altho it might be weird since the computer sprite i picked has directions.

M3739:
Test fail is still valid, but this time, for a different reason.
 ```
 Failed TestTargetIsValid [12 ms]
  Error Message:
     Found no targetNode "DecorativeComputer" on graph "DecorativeComputer" for construction prototype "DecorativeComputer"!
Assert.That(graph.Nodes.ContainsKey(target), Is.True)
  Expected: True
  But was:  False
  ```
  ```
  Failed ConstructionGraphEdgeValid [1 m 46 s]
  Error Message:
     One or more edges specified invalid node targets!
Invalid target "DecorativeComputer" in edge on node "start" of graph "DecorativeComputer"

Assert.That(valid, Is.True)
  Expected: True
  But was:  False
  ```

Resources/Prototypes/_Floof/Recipes/Crafting/computer.yml
M3739: Latest test doesn't report this anymore. Considering it resolved.

Mnemotechnician:
The DecorativeComputer construction graph is missing a "completed" node with the name DecorativeComputer.

Mnemotechnician:
So this PR is in the PR limbo, we need to contact github support I guess
image

Also, your computers and other tile-bound stuff need the following to make them snap to the center of a tile and stuff:
```yaml

  placementMode: SnapgridCenter
  canRotate: false
  canBuildInImpassable: false
  conditions:
    - !type:TileNotBlocked
```

Mnemotechnician:
Okay, now there's a new issue: CrateFreezer2 and CrateFreezer increase in price after being deconstructed (it means the materials they are made of are worth more than the final product)

Mnemotechnician (approved):

Memeji (approved):
Tests pass, looks good to me. Will wait for M3's review.

Mnemotechnician (approved):
Personally I'd prefer if those had more complex recipes constructions than just "put plastic on the ground", but this works.

M3739 (approved):
That'll do!

PR #1110, opened 2025-06-18
Mantis Rework - Mantis Is Security Again

Raikyr0:
# Description

Mantis has been re-added to security, disabled their ability to be antags.
Their SoP will need to be fixed.

Why?
They're the psychic detective, they exist to stop psychic crimes, them being an antag is almost as detrimental as the detective being able to be an antag. They literally start with a gun. People are complaining that Mantis needs a new job.

# Changelog

:cl:
- tweak: Tweaked Mantis to being Security again

### DIFF
--- a/Resources/Prototypes/Entities/Clothing/Ears/headsets.yml
+++ b/Resources/Prototypes/Entities/Clothing/Ears/headsets.yml
@@ -180,6 +180,23 @@
   - type: Clothing
     sprite: Clothing/Ears/Headsets/science.rsi
 
+- type: entity # Floof
+  parent: ClothingHeadset
+  id: ClothingHeadsetMantis
+  name: mantis headset # DeltaV - Epistemics Department replacing Science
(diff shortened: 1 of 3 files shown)


M3739:
Wait, the mantis used to be part of security? Could you elaborate on that, please? Cause I believe I wasn't there when this was the case.

Raikyr0:
> Wait, the mantis used to be part of security? Could you elaborate on that, please? Cause I believe I wasn't there when this was the case.

So, a long time ago it was the case, and at some point for whatever reason they separated it. There's a reason why in the code that Mantis is called forensicmantis. That said there was a point Detective wasn't security. Which is why on most maps their office isn't in sec.

M3739:
> So, a long time ago it was the case, and at some point for whatever reason they separated it. There's a reason why in the code that Mantis is called forensicmantis. That said there was a point Detective wasn't security. Which is why on most maps their office isn't in sec.

Interesting. Good to know.

PR #1117, opened 2025-06-18
PKA Structural Damage 

Raikyr0:
# Description

Increased PKA Structural damage slightly so its better for mining and salvaging and a tiny increase to projectile life expectancy so, a little more range.

# Changelog
:cl:
- tweak: Tweaked PKA to deal slightly more structural damage plus a little more range

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Guns/Projectiles/projectiles.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Guns/Projectiles/projectiles.yml
@@ -418,10 +418,10 @@
     damage:
       types:
         Blunt: 25
-        Structural: 30
+        Structural: 50
   # Short lifespan
   - type: TimedDespawn
-    lifetime: 0.4
+    lifetime: 0.5
 
 - type: entity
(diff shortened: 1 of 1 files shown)


Mnemotechnician:
This will allow PKA bolts to destroy solid walls in 4 hits and reinforced walls in 8 hits, while still leaving PKA as a terrible mining weapon (since when would salvage want to mine one rock per 2 seconds?). which is probably bad for the already terrible salvage balancing. Will need discussion.

Raikyr0:
> This will allow PKA bolts to destroy solid walls in 4 hits and reinforced walls in 8 hits, while still leaving PKA as a terrible mining weapon (since when would salvage want to mine one rock per 2 seconds?). which is probably bad for the already terrible salvage balancing. Will need discussion.

These are all bandaid fixes to the overall salvage issue. The intention is to safety remove [solid block ie rock or wall on a wreck] from a safe distance when it's fairly obvious there's a swarm of hostiles behind it.

ghost:
that is what the explosive charges are for.

PR #1121, opened 2025-06-19
Vampirism 1.5

Raikyr0:
# Description

It's time for Vampirism 1.5. Similar to the Succubus PR, I combined all the chemicals usually created into 1 to prevent bloodstream clogging via chemical processing limit. Greatly improved how much hunger/thirst is restored by drinking blood, to prevent issues like, the other person dying of bloodloss as easily. (Wardens everywhere rejoice.)

Healing is a little higher then before, still worse then an Platelets due to needing to actually drink the blood.

Why was the cost increased? To facilitate the fact the trait is more combat orientated and to prepare for Vampirism 2.0.

# Changelog
:cl:
- tweak: Tweaked Vampirism, cost increased, strength increased.

### DIFF
--- a/Resources/Prototypes/_Floof/Reagents/chemicals.yml
+++ b/Resources/Prototypes/_Floof/Reagents/chemicals.yml
@@ -11,29 +11,3 @@
   desc: reagent-desc-formaldehyde
   physicalDesc: reagent-physical-desc-sickly
   color: "#F26724"
-
-- type: reagent
-  id: Sanguinium
-  name: reagent-name-sanguinium
-  group: Medicine
-  desc: reagent-desc-sanguinium
-  physicalDesc: reagent-physical-desc-exhilarating # Should never realistically see or taste it
-  flavor: sweet
-  color: "#c90084"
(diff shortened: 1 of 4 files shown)


Resources/Prototypes/_Floof/Reagents/chemicals.yml
```diff
@@ -11,3 +11,35 @@
...
+      - !type:ModifyBloodLevel
+        amount: 1
+      - !type:HealthChange
+        damage:
+          groups:
+            Burn: -0.5
+            Toxin: -0.5
+            Airloss: -0.5
+            Brute: -0.5
+      - !type:ChemRerollPsionic # Won't do anything if you're not a psychic
```
Mnemotechnician: This should really not be here, especially not with the same bonus multiplier as loto oil.

Raikyr0: fair enough, was more so seeing if there was more ways to give a mantis actual mantis work

PR #1132, opened 2025-06-22
Fix Lightbulb and Light Tube Transmutation Bug

M3739:
# Description

The PR aims to address the bug where a light tube or light bulb, which costs 0.5 steel and glass, can be welded to be 1 whole glass.

This is fixed by making the glass cost 1 instead of 0.5, making welding it result in the loss of 0.5 steel at the end of it all.

Converting materials is something we'd like to avoid in order to maintain the purpose of materials and their respective uses, while also encouraging the sourcing of these materials to encourage interactions. As inconsequential as this change may be, it eliminates a improper solution to resource shortages.

---

# Changelog

:cl: M3739
- tweak: Light bulbs and light tubes had their material cost revised and corrected.

### DIFF
--- a/Resources/Prototypes/Recipes/Lathes/misc.yml
+++ b/Resources/Prototypes/Recipes/Lathes/misc.yml
@@ -1,4 +1,4 @@
-# Begin Changes - Floof - M3739 - LightbulbMoment
+# Begin Changes - Floof - M3739 - #1132
 - type: latheRecipe
   id: LightTube
   result: LightTube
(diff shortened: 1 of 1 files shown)


M3739:
```
Error Message:
     Found arbitrage on CrateVendingMachineRestockClothes cargo product! Cost is 1500 but sell is 1670!
```
h u h

Mnemotechnician:
> ```
> Error Message:
>      Found arbitrage on CrateVendingMachineRestockClothes cargo product! Cost is 1500 but sell is 1670!
> ```
> 
> h u h

It's been a failure on master today. Some of the recent merges did it.

M3739:
> It's been a failure on master today. Some of the recent merges did it.

Ah, with all the clothing we have been adding and stuff... Right. Think we just have to increase the cost accordingly for the clothing restock?

Mnemotechnician (approved):

PR #1135, opened 2025-06-24
Add Health Info to Basic Examine Action

ashkitten:
# Description

adds info from the health examine action to the default examine, because health examine is woefully underused outside of cases where you already know someone is injured. this allows people to more easily notice minor injuries and react accordingly in-character, enhancing the roleplay experience.

default examine will only display health info when there are injuries to display and will only show the basic health examine text, ignoring the self-aware trait. this has the slight benefit of giving self-aware players a better idea of what everyone else can see from looking at them.

---

<p>

Screenshot_20250624_123939
Screenshot_20250624_123758
Screenshot_20250624_123820

</p>
</details>

---

# Changelog

:cl:
- Added health information to the default examine window, so you can show off your pretty bruises <3

### DIFF
--- a/Content.Server/Body/Systems/BloodstreamSystem.cs
+++ b/Content.Server/Body/Systems/BloodstreamSystem.cs
@@ -244,7 +244,9 @@ public sealed class BloodstreamSystem : EntitySystem
         // Shows profusely bleeding at half the max bleed rate.
         if (ent.Comp.BleedAmount > ent.Comp.MaxBleedAmount / 2)
         {
-            args.Message.PushNewline();
+            // Floof: allow empty messages for basic examine
+            if (!args.Message.IsEmpty)
+                args.Message.PushNewline();
             if (!args.IsSelfAware)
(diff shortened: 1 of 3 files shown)


Mnemotechnician:
Afaik there's an upstream PR that just pushes all sub-examines into the main examine window, you could just cherry-pick that instead

ashkitten:
> Afaik there's an upstream PR that just pushes all sub-examines into the main examine window, you could just cherry-pick that instead

if you could dig up that PR i'd be grateful, but i intentionally wanted to make it behave somewhat differently from the health examine window in order to avoid information overload or inclusion of irrelevant info (like "There are no injuries to be seen").

In fact, I'm considering trying to condense it further into one line that would just say "She has some minor cuts and burns" or similar.

Mnemotechnician (changes requested):
I don't mind this change as-is, but we're gonna need you to annotate the changes with floofstation markers to make resolving conflicts easier in the future.

ashkitten:
floof comments added

Mnemotechnician (approved):

PR #1142, opened 2025-06-25
Heirloom & Loadout Special Menu Tweaks

Mnemotechnician:
# Description
- Fixes the heirloom button not saving the heirloom status of an item.
- Makes it so that the special settings menu cannot be opened if the relevant loadout item is not selected, and auto-save when it's closed.

# Changelog
:cl:
- fix: The heirloom toggle in loadouts show now work as expected.
- tweak: Loadout special menu can no longer be opened while the loadout is not selected, and will auto-save when it's closed.


Mnemotechnician:
Doesn't fix the bug where saving the character preferences doesn't save the loadout specials, but ehhh I'm really not in the mood to edit the character editor code to fix that

M3739:
Heisentest.

M3739 (approved):
After testing, it seems to function as intended.
LGTM

Memeji (approved):

PR #1143, opened 2025-06-25
Fix Buckled Rotation Breaking After Falling Asleep or Being Knocked Down

ashkitten:
# Description

deleted what seems to be a relic from The Before Times. buckling still works, and now you don't flop around like a fish.

---

# Changelog

:cl:
- fix: Resolved a longstanding issue where buckled characters would get all floppy after sleeping or being knocked down.

### DIFF
--- a/Content.Client/Buckle/BuckleSystem.cs
+++ b/Content.Client/Buckle/BuckleSystem.cs
@@ -21,7 +21,6 @@ internal sealed class BuckleSystem : SharedBuckleSystem
     {
         base.Initialize();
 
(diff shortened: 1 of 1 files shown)


Mnemotechnician:
Interesting, I've been trying to fix this bug for months. Does this still make north-facing buckles draw characters underneath?

ashkitten:
> Does this still make north-facing buckles draw characters underneath?

image

yep!

Mnemotechnician (approved):
Beautiful

Memeji (approved):

PR #1144, opened 2025-06-26
Give Attorney Justice Access.

AlexisMoonlight:
# Description

Give Attorney Justice access, allowing them to actually get into Justice department doors, security lobby. Remove Antag from Attorney.

---

# TODO

- [x] Possibly return antag to attorney if this level of access is fine for antag to have.

---

---

# Changelog

:cl:
- add: Added Justice access to Attorney.
- remove: Removed possibility for antag from Attorney.

### DIFF
--- a/Resources/Prototypes/Roles/Jobs/Civilian/lawyer.yml
+++ b/Resources/Prototypes/Roles/Jobs/Civilian/lawyer.yml
@@ -7,10 +7,12 @@
   startingGear: LawyerGear
   icon: "JobIconLawyer"
   supervisors: job-supervisors-cj  # Delta V - Change supervisor to chief justice
+  canBeAntag: false
   access:
   - Service
   - Lawyer
   - Maintenance
+  - Justice
 
 - type: startingGear
   id: LawyerGear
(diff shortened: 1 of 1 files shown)


Mnemotechnician:
In my opinion it makes perfect sense for lawyers to be traitors. Devil's advocate trope, all that.

AlexisMoonlight:
I only tested it on a couple maps, and they didnt seem to get more access than to the lobby of sec, less access than Prosec had IIRC. Still, that is more sec access than usual, since most roles cant even get there. And it's possible on some maps they get more security access, possibly too much for a possible antag. If that's not the case, it's an easy revert.

AlexisMoonlight:
Ok, in that case, I think it's in spirit of the trope, sec having to look after lawyers fits.

Mnemotechnician (approved):
Yeah, it makes sense I think

AlexisMoonlight:
Renamed, since it was reverted

M3739 (approved):

Memeji (approved):

PR #1145, opened 2025-06-26
Fix Hands Being Blocked when Carried Entity Polymorphs

ashkitten:
# Description

properly drops the carried entity before it's moved to the paused map so hands don't become blocked by nothing

i got rid of the line about "putting back down" item polymorphs because it seemed redundant. hopefully it is.

---

# Changelog

:cl:
- fix: Hands no longer become blocked when the person you're carrying polymorphs.

### DIFF
--- a/Content.Server/Polymorph/Systems/PolymorphSystem.cs
+++ b/Content.Server/Polymorph/Systems/PolymorphSystem.cs
@@ -1,4 +1,5 @@
 using Content.Server.Actions;
+using Content.Server.Carrying;
 using Content.Server.Humanoid;
 using Content.Server.Inventory;
 using Content.Server.Mind.Commands;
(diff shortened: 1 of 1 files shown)


Content.Server/Polymorph/Systems/PolymorphSystem.cs
```diff
@@ -365,8 +370,8 @@ private void OnDestruction(Entity<PolymorphedEntityComponent> ent, ref Destructi
         if (TryComp<PolymorphableComponent>(parent, out var polymorphableComponent))
             polymorphableComponent.LastPolymorphEnd = _gameTiming.CurTime;
 
-        // if an item polymorph was picked up, put it back down after reverting
-        _transform.AttachToGridOrMap(parent, parentXform);
+        if (TryComp<BeingCarriedComponent>(uid, out var carried))
+            _carrying.DropCarried(carried.Carrier, uid);
```
Mnemotechnician, line 382: This doesn't account for the situation where the polymorphed or the held entity is an item though. This may result in players getting sent into nullspace.

ashkitten: got it, that explains why the separate line about carried items was there. added back the line, and additionally to the initial polymorph in case someone is an item in the first place

Mnemotechnician (approved):
Should be good.

Memeji (approved):
LGTM

PR #1147, opened 2025-06-27
Sign Language Requires Complex Interaction and Two Empty Hands

ashkitten:
# Changelog
:cl:
- fix: Foxes can no longer use sign language, as their teeny tiny little paws are incapable of complex interaction.
- fix: Sign language now requires two empty hands. You're going to have to put that woman down.


Mnemotechnician (approved):

Memeji (approved):
Build tests pass, the current fail shouldn't cause an issue. Should be good to go.

PR #1165, opened 2025-06-30
Marked Traits

Cprn-Sprkl:
# Description

Adds a new set of traits that do not take up trait slots. The "Marked" traits opt that character in to chosen types of antag targeting no matter what job they hold. The system is expandable as more antags with targeting types are added; for example, cultists with sacrifice targets.

This PR only includes Marked traits for opting in as a target for the Traitor antag's Teach Lesson and Kill objectives.

---

# Changelog

:cl: Sprkl
- add: Crew can opt in to being the Kill or Teach target of Traitors.

### DIFF
--- a/Content.Server/Objectives/Components/PickRandomPersonComponent.cs
+++ b/Content.Server/Objectives/Components/PickRandomPersonComponent.cs
@@ -10,6 +10,13 @@ public sealed partial class PickRandomPersonComponent : Component
 {
     //Floofstation Target Consent Traits: Start
     [DataField]
-    public string ObjectiveType = string.Empty;
+    public ObjectiveTypes ObjectiveType = ObjectiveTypes.Unspecified;
+
+    public enum ObjectiveTypes
+    {
+        Unspecified,
+        TraitorKill,
+        TraitorTeach
+    }
     //Floofstation Target Consent Traits: End
 }
--- a/Content.Server/Objectives/Components/PickRandomPersonComponent.cs
+++ b/Content.Server/Objectives/Components/PickRandomPersonComponent.cs
@@ -10,13 +10,16 @@ public sealed partial class PickRandomPersonComponent : Component
 {
     //Floofstation Target Consent Traits: Start
     [DataField]
-    public ObjectiveTypes ObjectiveType = ObjectiveTypes.Unspecified;
-
-    public enum ObjectiveTypes
-    {
-        Unspecified,
-        TraitorKill,
-        TraitorTeach
-    }
+    public ObjectiveTypes ObjectiveType;
     //Floofstation Target Consent Traits: End
 }
+
+//Floofstation Target Consent Traits: Start
+[Flags]
+public enum ObjectiveTypes
+{
+    Unspecified = 0,
+    TraitorKill = 1 << 0,
+    TraitorTeach = 1 << 1
+}
+//Floofstation Target Consent Traits: End
--- a/Content.Server/Objectives/Components/PickRandomPersonComponent.cs
+++ b/Content.Server/Objectives/Components/PickRandomPersonComponent.cs
@@ -8,4 +8,8 @@ namespace Content.Server.Objectives.Components;
 [RegisterComponent, Access(typeof(KillPersonConditionSystem))]
 public sealed partial class PickRandomPersonComponent : Component
 {
+    //Floofstation Target Consent Traits: Start
+    [DataField]
+    public string ObjectiveType = string.Empty;
+    //Floofstation Target Consent Traits: End
 }
--- a/Content.Server/Objectives/Systems/KillPersonConditionSystem.cs
+++ b/Content.Server/Objectives/Systems/KillPersonConditionSystem.cs
@@ -1,3 +1,5 @@
+using Content.Server._Floof.Traits.Components;
+using Content.Server.Administration.Toolshed;
 using Content.Server.Objectives.Components;
 using Content.Server.Shuttles.Systems;
 using Content.Shared.CCVar;
@@ -43,6 +45,7 @@ public sealed class KillPersonConditionSystem : EntitySystem
 
     private void OnPersonAssigned(EntityUid uid, PickRandomPersonComponent comp, ref ObjectiveAssignedEvent args)
     {
+        Log.Warning("ONPERSON HAPPENED");
         // invalid objective prototype
         if (!TryComp<TargetObjectiveComponent>(uid, out var target))
         {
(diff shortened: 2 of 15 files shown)


Content.Server/Objectives/Components/PickRandomPersonComponent.cs
```diff
@@ -8,4 +8,8 @@ namespace Content.Server.Objectives.Components;
 [RegisterComponent, Access(typeof(KillPersonConditionSystem))]
 public sealed partial class PickRandomPersonComponent : Component
 {
+    //Floofstation Target Consent Traits: Start
+    [DataField]
+    public string ObjectiveType = string.Empty;
```
Mnemotechnician: This really shouldn't be an arbitrary string. It should be either a serializable enum, or a ProtoId\<some new prototype for objective type\>. Since it doesn't store anything but a type, enum is probably what you want. The enum could also be marked as [Flags], easily allowing you to store multiple target types in the same field.

Content.Server/_Floof/Traits/Components/MarkedTraitorKillComponent.cs
```diff
@@ -0,0 +1,8 @@
+namespace Content.Server._Floof.Traits.Components;
+
+/// <summary>
+///     Marks this player as eligible for being the target of
+///     round removing kill objectives for traitors.
+/// </summary>
+[RegisterComponent]
+public sealed partial class MarkedTraitorKillComponent : Component;
```
Mnemotechnician: This could be joined with MarkedTraitorTeachComponent into some MarkedTargetComponent with a single field containing a value from the aformentioned enum - the target type.

Content.Server/Objectives/Systems/KillPersonConditionSystem.cs
```diff
@@ -63,14 +66,51 @@ private void OnPersonAssigned(EntityUid uid, PickRandomPersonComponent comp, ref
...
+        {
+            foreach (var mind in markedList)
+            {
+                if (!HasComp<MarkedTraitorTeachComponent>(mind.Comp.CurrentEntity))
+                {
+                    markedList.Remove(mind);
+                }
+            }
+        }
+        return markedList;
```
Mnemotechnician: Ditto about enums. If you use a flag enum, you will also be able to replace the string comparison with something like `if (objType.HasFlag(ObjectiveType.Kill))`.

Resources/Prototypes/Traits/categories.yml
```diff
@@ -7,7 +7,11 @@
 - type: traitCategory
   id: Auditory
   root: true
-
+#Floofstation Target Consent Traits: Start
+- type: traitCategory
```
Mnemotechnician: This belongs in /Prototypes/_Floof/Traits/categories.yml (namespace change)

Content.Server/Objectives/Systems/KillPersonConditionSystem.cs
```diff
@@ -43,6 +45,7 @@ private void OnGetProgress(EntityUid uid, KillPersonConditionComponent comp, ref
 
     private void OnPersonAssigned(EntityUid uid, PickRandomPersonComponent comp, ref ObjectiveAssignedEvent args)
     {
+        Log.Warning("ONPERSON HAPPENED");
```
Mnemotechnician: Probably should remove this:
```suggestion
```

Mnemotechnician (changes requested):
Good idea, but needs some cleanup.

Cprn-Sprkl:
Still planning to consolidate the Marked components into one, but I'm encountering a strange issue in testing that I'm having trouble pinning down. Someone who has neither trait is getting assigned as a teach target, but it doesn't seem to be doing this through the same target assigning function that is normally used. I have a break in that function to see every time it is called so I can step through and check the variables and such, and every time it assigns someone who shouldn't be assigned it doesn't break. There is something, somewhere, that is assigning a teach target without raising OnPersonAssigned.

Mnemotechnician:
> Still planning to consolidate the Marked components into one, but I'm encountering a strange issue in testing that I'm having trouble pinning down. Someone who has neither trait is getting assigned as a teach target, but it doesn't seem to be doing this through the same target assigning function that is normally used. I have a break in that function to see every time it is called so I can step through and check the variables and such, and every time it assigns someone who shouldn't be assigned it doesn't break. There is something, somewhere, that is assigning a teach target without raising OnPersonAssigned.

You didn't assign an objective type to the "teach a lesson" objective entity prototype, and your code simply returns the original list of all people if no type is provided.

Cprn-Sprkl:
> > Still planning to consolidate the Marked components into one, but I'm encountering a strange issue in testing that I'm having trouble pinning down. Someone who has neither trait is getting assigned as a teach target, but it doesn't seem to be doing this through the same target assigning function that is normally used. I have a break in that function to see every time it is called so I can step through and check the variables and such, and every time it assigns someone who shouldn't be assigned it doesn't break. There is something, somewhere, that is assigning a teach target without raising OnPersonAssigned.
> 
> You didn't assign an objective type to the "teach a lesson" objective entity prototype, and your code simply returns the original list of all people if no type is provided.

As of the most recent commits the code should cancel assigning the objective at all if no type is specified. I'd left it the other way as a fail safe in case something I didn't know about was using that system, but I had pulled it out entirely when I noticed this most recent issue.

As for not assigning: In the TeachLessonRandomPersonObjective definition in Prototypes/DeltaV/Objective/traitor.yml I have an objectiveType of 2 being assigned. Is that not the right one? TeachLessonRandomHeadObjective has been left untouched because it uses a different function for assigning and isn't relative to this yet.

Cprn-Sprkl:
Found it. OnHeadAssigned had a fallback to non-head targets when no heads were available. I'm a little shocked this hasn't been a problem in the main server because TeachLessonRandomHead is the only objective in the kill pool for traitors.

`if (allHeads.Count == 0)
            allHeads = allHumans.Select(x => x.Owner).ToList(); // fallback to non-head target`

This is the culprit for the person with no traits and no command job getting picked.

Resources/Prototypes/DeltaV/Objectives/traitor.yml
```diff
@@ -63,4 +63,5 @@
   - type: TargetObjective
     title: objective-condition-teach-person-title
   - type: PickRandomPerson
+    objectiveType: 2
```
Mnemotechnician:
```suggestion
    objectiveType: TraitorTeach
```

Resources/Prototypes/Objectives/traitor.yml
```diff
@@ -110,6 +110,7 @@
   - type: TargetObjective
     title: objective-condition-kill-person-title
   - type: PickRandomPerson
+    objectiveType: 1
```
Mnemotechnician:
```suggestion
    objectiveType: TraitorKill
```

Content.Server/Objectives/Systems/KillPersonConditionSystem.cs
```diff
@@ -63,14 +65,56 @@ private void OnPersonAssigned(EntityUid uid, PickRandomPersonComponent comp, ref
...
+        {
+            args.Cancelled = true;
+            return;
+        }
         // Floofstation Edit End
 
         _target.SetTarget(uid, _random.Pick(allHumans), target);
     }
 
+    public HashSet<Entity<MindComponent>> MarkedList(HashSet<Entity<MindComponent>> markedList, ObjectiveTypes objType)
```
Mnemotechnician:
```suggestion
    // Floofstation - added this entire function
    public HashSet<Entity<MindComponent>> MarkedList(HashSet<Entity<MindComponent>> markedList, ObjectiveTypes objType)
```

Mnemotechnician (commented):
A few cleanup notes, other than that should be good. Will need to wait until someone tests or until you provide a demonstration, though

Resources/Prototypes/_Floof/Traits/marked.yml
```diff
@@ -0,0 +1,21 @@
...
+- type: trait
+  id: TraitorTeachTarget
+  category: Marked
+  points: 0
+  slots: 0
+  functions:
+  - !type:TraitAddComponent
+    components:
+    - type: Marked
+      targetType: 2
```
Mnemotechnician:
```suggestion
      targetType: TraitorKill

- type: trait
  id: TraitorTeachTarget
  category: Marked
  points: 0
  slots: 0
  functions:
  - !type:TraitAddComponent
    components:
    - type: Marked
      targetType: TraitorTeach
```

Mnemotechnician (commented):
Ugh, github lost this comment

Cprn-Sprkl:
For testing I launched the server and three clients. One character had the teach Mark, one had the kill Mark, and one had neither. I loaded them into a round and made each into a traitor. After reviewing their objectives I did restartroundnow, readyall, and a forcepreset to launch back in.

During objective assignment I had breakpoints in two different functions so I could check the variables and final lists. I would also compare the final traitor objective assignments in game for any being pointed at the wrong character. After a few tries of that I turned the breakpoints off and ran through just checking objectives ten times.

Objectives were correctly assigned every time as of the most recent commits, but due to how bad it would be for someone to get assigned who did not ask for it I would highly recommend another tester grab this and check it. Also, any suggestions for better testing setups for this kind of thing in the future would be great.

testing

Resources/Prototypes/_Floof/Traits/marked.yml
```diff
@@ -0,0 +1,21 @@
...
+  - !type:TraitAddComponent
+    components:
+    - type: Marked
+      targetType: TraitorKill
+
+- type: trait
+  id: TraitorTeachTarget
+  category: Marked
+  points: 0
+  slots: 0
```
Mnemotechnician: Just noticed those don't give any points. I think it would make perfect sense for them to give something to incentivise people to take them for reasons other than boredom. Not too much, because you are not guaranteed to be a target, so something like:
```suggestion
  points: 2
  slots: 0
  functions:
  - !type:TraitAddComponent
    components:
    - type: Marked
      targetType: TraitorKill

- type: trait
  id: TraitorTeachTarget
  category: Marked
  points: 1
  slots: 0
```

Mnemotechnician (approved):
Okay, I think it looks good. Just one more thing to discuss:

Resources/Prototypes/_Floof/Traits/marked.yml
Cprn-Sprkl: I had set those to 0 at the beginning with the intention to ask about how much people thought they'd be worth and then promptly forgot about them. 1 and 2 seems good and they can always be easily adjusted up or down based on community feedback.

Memeji (approved):
Awesome stuff!

PR #1166, opened 2025-06-30
No Duplicate Random Map

### LINKED ISSUE #1163: Map Voting Improvements
# Description

Improve the voting system by making it so that the map in the previous shift cannot be chosen by the "Random" vote.

Mnemotechnician:
# Description
Resolves #1163

Was tested by starting & restarting the round about 10 times in a row, each time selecting "random map" with 4 eligible maps, didn't get any duplicates, which gives me a 94.3% confidence this works.

# Changelog
:cl:
- tweak: Voting for random map will no longer result in the last map being chosen for the second time (unless the vote was called manually, or the current map got changed)

### DIFF
--- a/Content.Server/Voting/Managers/VoteManager.DefaultVotes.cs
+++ b/Content.Server/Voting/Managers/VoteManager.DefaultVotes.cs
@@ -212,6 +212,7 @@ namespace Content.Server.Voting.Managers
 
         private void CreateMapVote(ICommonSession? initiator)
         {
+            var lastMap = _gameMapManager.GetSelectedMap(); // Floof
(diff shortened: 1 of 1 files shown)


Mnemotechnician:
So yeah, this will work as expected, but only so long as the vote is called automagically by the server at the end of a round, or mid-round. In-lobby votes called by admins won't be affected, but I feel like they shouldn't be anyway. (this is because the game.map cvar is reset after the round restarts, unless it's forcemapped)

ShadesMars (approved):

Memeji (approved):

PR #1167, opened 2025-07-02
Hypno Fix Attempt

Raikyr0:
# Description

So, I had a look, and I think I fixed it, the pathing for the CS was wrong was using old "Content.Shared.Floofstation.Hypno;" instead of "Content.Shared._Floof.Hypno;" which was probably causing issues.

# Changelog

:cl:
- fix: Fixed Hypno

### DIFF
--- a/Content.Client/_Floof/HypnotizedSystem.cs
+++ b/Content.Client/_Floof/HypnotizedSystem.cs
@@ -1,4 +1,4 @@
-using Content.Shared.Floofstation.Hypno;
+using Content.Shared._Floof.Hypno;
 using Content.Shared.StatusIcon;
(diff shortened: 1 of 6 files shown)


Mnemotechnician:
Namespaces are irrelevant and are used for nothing other than grouping. Please test your changes instead of relying on intuition. The issue occurs because hypno icons have the type HypnoIconPrototype (why? It has no reason to exist) and the system tries to fetch a StatusIconPrototype with the same ID, which obviously doesn't exist: https://github.com/Floof-Station/Floof-Station/blob/9609b8967b7f9dd7bf249ca88127a89bdb0faec3/Content.Client/_Floof/PsionicHypnoSystem.cs#L32

PR #1169, opened 2025-07-03
Fix Language Popups 

Mnemotechnician:
# Description
Makes it so that the popup is only shown to the source, not to everyone in range.

# Changelog
No cl


Mnemotechnician:
Weird, I thought it wasn't supposed to be possible to create a new branch and it'd create one on my fork instead...

PR #1170, opened 2025-07-03
Allows Arachne's to Spin Web  

VividPups:
# Description
Look a Minotur Charges like a bull and a Centuor has a cock like a horse.  arachne has an ass like a spider might as will be able to make web 

# Media 
image
---

# Changelog

:cl:
- tweak: Arachne has an ass of an spider might as well make web like an arachnid naturally


Memeji (approved):
Agreed with this change

PR #1176, opened 2025-07-05
Fix Custom Examine Expirations Not Working

Mnemotechnician:
# Description
I was passing CustomExamineData by-value instead of by-ref. Blegh.

# Changelog
:cl:
- fix: Custom examine texts should now correctly be removed when they expire.


M3739:
Heisentest, it seems. Rerunning.

M3739 (approved):
Still a heisentest. Great.

On a side note, local testing confirms functionality.

Mnemotechnician:
> Heisentest, it seems. Rerunning.

I just don't bother rerunning them generally. Unless it's the kind of heisentest that causes ALL subsequent tests to fail (where the server-client pair dies, or where the testing run gets cancelled), I just check if there are any other test fails beside the heisentest.

Memeji (approved):

PR #1181, opened 2025-07-06
Mood System Fix & Heirloom Timeout Extension

Mnemotechnician:
# Description
Turns out the mood system relies on Timers to implement mood effect timeouts. Yes, timers. The asynchronous programming tool that has no place in a system like this. And the best part, the system doesn't cancel any previously active timers when applying the same mood effect twice. This means that if you got a mood effect 30 seconds ago, and then got it again (e.g. from two kisses), the reset timer from 30 seconds ago will still keep going, resulting in the moodlet being removed 30 seconds earlier than intended.

This PR attempts to resolve this issue in the least destructive way possible - by introducing a dictionary of CancellationTokenSources to the MoodComponent, providing CancellationTokens to the timers created by the mood system, and cancelling any previously going timers for the same MoodEffectPrototype.

Tested; seems to work correctly.

Also increases the time between checks in the HeirloomSystem from 10 seconds to 1 minute, and increases the heirloom moodlet timeout to 5 minutes.

<p>

30 minutes of mostly passive testing:

https://github.com/user-attachments/assets/533f781b-9630-4542-8fc9-f15e3fd65605

</p>
</details>

# Changelog
:cl:
- fix: Repeated mood effects (such as that from hunger or nicotine addiction) should no longer cause your mood to flicker.

### DIFF
--- a/Content.Server/Mood/MoodComponent.cs
+++ b/Content.Server/Mood/MoodComponent.cs
@@ -1,5 +1,7 @@
-﻿using Content.Shared.Alert;
+﻿using System.Threading;
+using Content.Shared.Alert;
 using Content.Shared.FixedPoint;
+using Content.Shared.Mood;
 using Robust.Shared.Prototypes;
 using Robust.Shared.Serialization.TypeSerializers.Implementations.Generic;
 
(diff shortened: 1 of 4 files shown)


M3739 (approved):
I have always been annoyed of having my vision flicker from time to time via color due to how jittery the mood system can be with a heirloom prior to this fix.

Good change! Now I don't have to have the occasional color flickering.

Memeji (approved):

PR #1187, opened 2025-07-07
Change Argument Order

ghost:
The order of the arguments is a little clunky, so I made the PR number the last, this way it's much easier to do the next PR.

### DIFF
--- a/Scripts/git/mirror-pr.sh
+++ b/Scripts/git/mirror-pr.sh
@@ -8,14 +8,14 @@ This script will cherry-pick a pull request from a GitHub repo into a new branch
 
 Default pull remote is 'upstream', default GitHub repo is 'simple-station/einstein-engines', default GitHub repo to create a PR in is 'fansana/floofstation1'"
 
-PRNUM=$1 # First arg to a shell script
+PRNUM=$4 # First arg to a shell script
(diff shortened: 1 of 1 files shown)


Mnemotechnician:
Uhh, the original idea was to keep it simple and allow to just type the PR number to mirror from EE to floof (that's what the {$3:-SomeDefaultValue} syntax is for - it substitutes the argument for a default value if it's unset. I think the code might need some additional tweaking to support the new order (e.g. updating the help string)

Mnemotechnician:
Github won't let me suggest changes on those lines, but should look something like this:
```bash
HELPSTR="Usage: $( basename $0 ) <remote to push to> <GitHub repo to pull from> <GitHub repo to create a PR in> <PR number>
For example: $( basename $0 ) floofstation space-wizards/space-station-14 fansana/floofstation1 1002

This script will cherry-pick a pull request from a GitHub repo into a new branch and make a pull request for you.

Default pull remote is 'upstream', default GitHub repo is 'simple-station/einstein-engines', default GitHub repo to create a PR in is 'fansana/floofstation1'"

PRNUM=$4
if [[ -z $PRNUM || $1 == "--help" || $1 == "-h" ]]; then
    echo "$HELPSTR"
    exit 1
fi
```

ghost:
I started writing a new script that will probably replace this one. I tried this script on frontier but they don't use squash merges so I had to adapt.

PR #1191, opened 2025-07-07
Type Writer Fix Can Wield and Bolt when Wielding

VividPups:
# Description
Thommy couldnt bolt when wielded it would just unwield, thus this adds the parent to fix it

:cl:

- fix:  Choko Skill issue with the typewriter

### DIFF
--- a/Resources/Prototypes/DeltaV/Entities/Objects/Weapons/Guns/SMGs/smgs.yml
+++ b/Resources/Prototypes/DeltaV/Entities/Objects/Weapons/Guns/SMGs/smgs.yml
@@ -1,6 +1,6 @@
-- type: entity
+- type: entity #
   name: Typewriter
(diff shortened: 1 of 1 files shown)


M3739:
Gonna test this later on today.

M3739 (changes requested):
Alright, works like a charm.

Only thing I will request is to remove the "skill issue" code comments and rewrite the changelog due to the same reason.

VividPups:
> Alright, works like a charm.
> 
> Only thing I will request is to remove the "skill issue" code comments and rewrite the changelog due to the same reason.

she needs to know

Memeji:
> > Alright, works like a charm.
> > Only thing I will request is to remove the "skill issue" code comments and rewrite the changelog due to the same reason.
> 
> she needs to know

You are required to change the comments and the changelog in order for this PR to be accepted.

PR #1192, opened 2025-07-07
Make All Hats in the Loadout Hairloomable 

VividPups:
# Description

Suggested by a Discord user, they wanted all hats hairloomable Don't see a reason why they can't 

image

# Changelog

:cl:
- tweak: all hats can be Heirlooms now


VividPups:
not all hats but most hats the more people complain the more I see which ones need to be added

PR #1195, opened 2025-07-08
Add Favorite Drink Preference

ashkitten:
also changed instances of `IsDirty = true` to `SetDirty()` since it was causing issues with profile incorrectly being marked as dirty

:cl:
- add: Added favorite drink preference to character profile, which improves mood when consumed.


Content.Server/_Floof/Preferences/FavoriteDrinkSystem.cs
```diff
@@ -0,0 +1,26 @@
...
+    {
+        if (args.Profile.FavoriteDrink is not { } favoriteDrink)
+            return;
+
+        var comp = new FavoriteDrinkComponent()
+        {
+            ReagentId = favoriteDrink
+        };
+
+        AddComp(args.Mob, comp);
```
Mnemotechnician: Cleaner way to do this:
```suggestion
        var comp = EnsureComp<FavoriteDrinkComponent>(args.Mob);
        comp.ReagentId = favoriteDrink;
```
Also note that this entire system could be rendered unnecessary as you can fetch HumanoidCharacterProfile.FavoriteDrink directly from HumanoidAppearenceComponent, and use FavoriteDrinkComponent as a fallback

Content.Server/Nutrition/EntitySystems/DrinkSystem.cs
```diff
@@ -322,7 +325,24 @@ private void OnDoAfter(Entity<DrinkComponent> entity, ref ConsumeDoAfterEvent ar
...
+
+            // minimum threshold for mood effect (same as ice cream)
+            // notable difference here is that ice cream's effect occurs when metabolized
+            // whereas this occurs when ingested, so 5 is also the maximum amount you can drink at once
+            if (quantity.Quantity < 5)
+                return;
+
+            var target = args.Target.Value;
+            // raise after a timeout so the flytext doesn't overlap
+            Timer.Spawn(1000, () => RaiseLocalEvent(target, new MoodEffectEvent("FavoriteDrink")));
```
Mnemotechnician: It would be better to put all of this into a separate system and just call a method/raise an even here, but this works for now

Mnemotechnician (commented):
A few nitpicks. Other than that code-wise this looks fine. I hate the character editor code with all of my soul...

southbridge-fur:
I love this

Content.Server/_Floof/Preferences/FavoriteDrinkSystem.cs
ashkitten: took your suggestion but i decided to leave the method in because i think it's confusing to have multiple sources of truth for this (for one thing making it impossible or confusing to remove the favorite drink preference in-game with VV, and HumanoidAppearanceComponent is not copied over to polymorphs for good reason, while it's more probable that FavoriteDrinkComponent will be allowed to copy over in the future)

Memeji (approved):

PR #1197, opened 2025-07-09
Thieving Trait Now only Shows the Silhouette of Pocketed Items

ashkitten:
# Description

the full item is still visible to admin ghosts or anyone with BypassInteractionChecksComponent

---

<p>

image

</p>
</details>

---

# Changelog

:cl:
- tweak: Thieving trait now only shows the silhouette of pocketed items

### DIFF
--- a/Content.Client/Inventory/StrippableBoundUserInterface.cs
+++ b/Content.Client/Inventory/StrippableBoundUserInterface.cs
@@ -12,11 +12,13 @@ using Content.Shared.Ensnaring.Components;
 using Content.Shared.Hands.Components;
 using Content.Shared.IdentityManagement;
 using Content.Shared.Input;
+using Content.Shared.Interaction.Components;
 using Content.Shared.Inventory;
 using Content.Shared.Inventory.VirtualItem;
 using Content.Shared.Strip.Components;
 using JetBrains.Annotations;
 using Robust.Client.GameObjects;
+using Robust.Client.Graphics;
 using Robust.Client.Player;
 using Robust.Client.UserInterface;
 using Robust.Client.UserInterface.Controls;
@@ -24,6 +26,7 @@ using Robust.Client.Player;
 using Robust.Shared.Input;
 using Robust.Shared.Map;
 using Robust.Shared.Prototypes;
+using Robust.Shared.Serialization.Manager;
 using static Content.Client.Inventory.ClientInventorySystem;
 using static Robust.Client.UserInterface.Control;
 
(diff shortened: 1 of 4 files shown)


ashkitten:
should Thieving have a reduced point cost after this change since it's somewhat less useful?

Content.Client/Inventory/StrippableBoundUserInterface.cs
```diff
@@ -59,7 +67,8 @@ public StrippableBoundUserInterface(EntityUid owner, Enum uiKey) : base(owner, u
             _cuffable = EntMan.System<SharedCuffableSystem>();
             _strippable = EntMan.System<StrippableSystem>();
 
-            _virtualHiddenEntity = EntMan.SpawnEntity(HiddenPocketEntityId, MapCoordinates.Nullspace);
+            _silhouetteShader = _prototype.Index<ShaderPrototype>("Silhouette").InstanceUnique();
+            _silhouetteShader.SetParameter("color", new Vector4(0.0f, 0.0f, 0.0f, 1.0f));
```
southbridge-fur: Could this color be added to the `ThievingComponent` and set prior to being rendered? That way if we wanted to change it in the future we could. 

Also you can just use the `Robust.Shared.Maths.Color`'s `Color.RGBA` member to get the Vector4 and ensure that color can be serialized in the component

southbridge-fur (changes requested):
Everything looks good but I'd like to see the color get moved to the `ThievingComponent`

Content.Client/Inventory/StrippableBoundUserInterface.cs
ashkitten: sure, though do you mean it should be editable in-game? auto-networked?

ashkitten:
Pushed requested change, also adjusted the color to an off-white for better contrast with all themes.

Resources/Prototypes/Shaders/shaders.yml
```diff
@@ -134,7 +134,12 @@
...
   id: Hologram
   kind: source
-  path: "/Textures/Shaders/hologram.swsl"
\ No newline at end of file
+  path: "/Textures/Shaders/hologram.swsl"
+
+- type: shader
+  id: Silhouette
+  kind: source
+  path: "/Textures/Shaders/silhouette.swsl"
```
Mnemotechnician: Would be preferrable to put this and Resources/Textures/Shaders/silhouette.swsl under the floofstation namespace.

Mnemotechnician (changes requested):
Code-wise should be fine, but changes to upstream files need to be annotated as belonging to floofstation.

Content.Client/Inventory/StrippableBoundUserInterface.cs
Mnemotechnician: Yes. If you want the client to know about server-side changes to this componet (e.g. those happening through VV or when applied via a trait - traits are server-side), this field needs to be auto-networked.

ashkitten:
pushed all requested changes

Mnemotechnician (approved):

Memeji (approved):

PR #1198, opened 2025-07-09
Make Hiding Under Tables Obscure You From HUDs

ashkitten:
# Description

This was requested by a user in discord, as they've already been playing as though the feature exists.

---

<p>

image

</p>
</details>

---

# Changelog

:cl:
- tweak: Hiding under tables now obscures you from HUD overlays

### DIFF
--- a/Content.Client/StatusIcon/StatusIconSystem.cs
+++ b/Content.Client/StatusIcon/StatusIconSystem.cs
@@ -1,10 +1,13 @@
 using Content.Shared.CCVar;
 using Content.Shared.Ghost;
+using Content.Shared.Physics;
 using Content.Shared.StatusIcon;
 using Content.Shared.StatusIcon.Components;
 using Content.Shared.Stealth.Components;
 using Content.Shared.Whitelist;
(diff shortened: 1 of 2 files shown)


Mnemotechnician:
That is a very interesting way to do it. But are you sure it's worth the effort to do this instead of just drawing the HUD icon on the same layer as the mob?

Either way, this needs floofstation landmark comments.

ashkitten:
> That is a very interesting way to do it. But are you sure it's worth the effort to do this instead of just drawing the HUD icon on the same layer as the mob?

yeah, the HUD would poke out above the table. also HUDs are rendered by RT, so i don't know if i can touch that but yeah

Mnemotechnician:
Are they rendered by RT? If so, yeah, we can't change that sadly. Hard to believe there's no way to change the sprite layer they are rendered on, though...

ashkitten:
added floof markers

Mnemotechnician (approved):

Memeji (approved):

PR #1208, opened 2025-07-11
Use Encryption Keys Instead of Intrinsic Radio Channels for pAI

ashkitten:
# Description

I decided against making them require a lock, since it's similar to a headset as something you carry around on your person. If your pAI gets stolen and encryption keys dumped, that's your problem.

This change incidentally adds the ability to talk on the Common channel by default (they could only listen on it before). There's no good way for me to align it with the old behavior while still allowing to use encryption keys, and it would be confusing for players.

---

# Changelog

:cl:
- tweak: pAIs now use encryption keys, allowing them to talk on more than just the binary channel

### DIFF
--- a/Content.Shared/Radio/EntitySystems/EncryptionKeySystem.cs
+++ b/Content.Shared/Radio/EntitySystems/EncryptionKeySystem.cs
@@ -40,6 +40,7 @@ public sealed partial class EncryptionKeySystem : EntitySystem
         SubscribeLocalEvent<EncryptionKeyHolderComponent, ExaminedEvent>(OnHolderExamined);
 
         SubscribeLocalEvent<EncryptionKeyHolderComponent, ComponentStartup>(OnStartup);
+        SubscribeLocalEvent<EncryptionKeyHolderComponent, MapInitEvent>(OnMapInit);
         SubscribeLocalEvent<EncryptionKeyHolderComponent, InteractUsingEvent>(OnInteractUsing);
         SubscribeLocalEvent<EncryptionKeyHolderComponent, EntInsertedIntoContainerMessage>(OnContainerModified);
         SubscribeLocalEvent<EncryptionKeyHolderComponent, EntRemovedFromContainerMessage>(OnContainerModified);
@@ -181,6 +182,10 @@ public sealed partial class EncryptionKeySystem : EntitySystem
     private void OnStartup(EntityUid uid, EncryptionKeyHolderComponent component, ComponentStartup args)
     {
         component.KeyContainer = _container.EnsureContainer<Container>(uid, EncryptionKeyHolderComponent.KeyContainerName);
+    }
+
+    private void OnMapInit(EntityUid uid, EncryptionKeyHolderComponent component, MapInitEvent args)
+    {
         UpdateChannels(uid, component);
     }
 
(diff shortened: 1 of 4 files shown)


Mnemotechnician:
I don't mind, but binary should probably be kept as an intrinsic channel, otherwise people will be able to simply loot PAIs for binary keys.

ashkitten:
is it possible to do that? i wasn't able to figure out how to make an intrinsic key but i didn't look too hard

Mnemotechnician:
> is it possible to do that? i wasn't able to figure out how to make an intrinsic key but i didn't look too hard

Just keep it in the IntrinsicRadioTransmitter/ActiveRadio.

ashkitten:
> Just keep it in the IntrinsicRadioTransmitter/ActiveRadio.

this doesn't work as it gets overridden by the encryption key system. we can't mix and match

Mnemotechnician:
That's weird and unfortunate. Perhaps creating a new unremovable binary encryption key entity be a solution?

ashkitten:
> Perhaps creating a new unremovable binary encryption key entity be a solution?

.....space glue..... *uncontrollable giggling*

ashkitten:
> Perhaps creating a new unremovable binary encryption key entity be a solution?

this worked, had to fix a bug in the code though, and added some locale tweaks to support

ashkitten:
not sure what's needed to resolve the added component tests? resave the maps?

Mnemotechnician:
> not sure what's needed to resolve the added component tests? resave the maps?

No, it means that the system modifies its components on ComponentStartup rather than on MapInit, which it shouldn't be doing (this is an error because at this point not all components might be loaded/ready). You can work around it by replicating the same channels into IntrinsicRadioTransmitter and ensuring that PersonalAI has a ContainerContainer with the appropriate encryption key container defined in its yaml prototype (so the entity manager doesn't cry about it gaining a new component on spawn).

Basically, what [Filled] intercoms do in their yaml prototypes.

M3739 (changes requested):
Even though it has been discussed before, just laying it out for the record.
Test fail is valid.
```
   Multiple failures or warnings in test:
  1) Prototype SyndicatePersonalAI modifies component on spawn: IntrinsicRadioTransmitter. Modified yaml:
channels: []
...

  2) Prototype SyndicatePersonalAI gains a component on spawn: ContainerContainer
  3) Prototype PotatoAI modifies component on spawn: IntrinsicRadioTransmitter. Modified yaml:
channels: []
...

  4) Prototype PotatoAI gains a component on spawn: ContainerContainer
  5) Prototype PersonalAI modifies component on spawn: IntrinsicRadioTransmitter. Modified yaml:
channels: []
...

  6) Prototype PersonalAI gains a component on spawn: ContainerContainer
```

ashkitten:
fixed the test failure by changing the keys system to modify components on MapInit rather than ComponentStartup, as the suggested workaround did not completely cover the issue.

ashkitten:
oops, i broke everything

ashkitten:
okay fixed actual

Mnemotechnician (approved):
Looks good to me.

M3739 (approved):
Just a note for now: in upstream there is a PR that would allow for PAIs to function as positronic brains via an adapter. I am unsure how this system would play out with these changes, but we shall see when the next week rolls around.

PR #1212, opened 2025-07-14
Atmos Tech TC Fix

Raikyr0:
# Description

Gives Atmos 2 more TC to be in line with Senior Engineers. Atmos are forbidden from doing all the mass destruction crimes they're usually able of doing on a LRP server on Floof.

# Changelog

:cl:
- tweak: Added 2 TC back to Atmos Techs to be more in line with Engineers

### DIFF
--- a/Resources/Prototypes/Roles/Jobs/Engineering/atmospheric_technician.yml
+++ b/Resources/Prototypes/Roles/Jobs/Engineering/atmospheric_technician.yml
@@ -7,7 +7,7 @@
   - !type:CharacterDepartmentTimeRequirement
     department: Engineering
     min: 3600 # Floofstation - 1 hour
-  antagAdvantage: 5 # Floof - Reduced TC: External Access + Fireaxe + Free Hardsuit, but they cannot pump chemicals into the air supply en masse.
+  antagAdvantage: 3 # Floof - Reduced TC: External Access + Fireaxe + Free Hardsuit, but they cannot pump chemicals into the air supply en masse.
   startingGear: AtmosphericTechnicianGear
   icon: "JobIconAtmosphericTechnician"
   supervisors: job-supervisors-ce


M3739:
Funnily enough, the premise of this PR seems to have already been handled by #992. Which gave atmos techs more TC because of the same reason.

Ergo, I would like some more reasoning as to why they should get more TC than previously given by #992

Raikyr0:
> Funnily enough, the premise of this PR seems to have already been handled by #992. Which gave atmos techs more TC because of the same reason.
> 
> Ergo, I would like some more reasoning as to why they should get more TC than previously given by #992

Small list of reason
> Less TC then Senior Engi who also has Atmos Access
> Fireaxe was nerfed so it's a bad melee weapon, and you're not really supposed to just go around spacing the station.
> Some station atmos don't even have a fireaxe. (Mapping issue)

It falls into, do you believe the fireaxe is worth 2 TC. But also Senior Engi has more TC.

M3739 (approved):
I see.

Honestly, to keep it more in line with the Senior Engi is a reason I can get behind. 

The fireaxe issue is moot, since it is a *tool*, and a good one at that. Really good at busting windows, but it's uniqueness on station often means the possibility of it being noted missing.

Just need opinions from the another maintainer. But for now, this has my approval.

Raikyr0:
> I see.
> 
> Honestly, to keep it more in line with the Senior Engi is a reason I can get behind.
> 
> The fireaxe issue is moot, since it is a _tool_, and a good one at that. Really good at busting windows, but it's uniqueness on station often means the possibility of it being noted missing.
> 
> Just need opinions from the another maintainer. But for now, this has my approval.

The Fireaxe also happens to be an item that Gladiators can take on Arena as I recall. But yeah if you wanted to take out a window, RCDing would be faster and quieter. It's only quality is ripping up a floor tile and spacing the room.

Raikyr0:
Closing this PR so we can just outright remove antag advantages overall

PR #1213, opened 2025-07-14
Corrupt Cop

Raikyr0:
# Description

This PR will likely be highly controversial, but this enables Security (not including detectives or wardens) to be antags with a 10 less TC.

Why?
We don't live in a black and white world, where police and security are infallible forces of good. The mindshield isn't a loyalty implant. Deep agents exist and have already infiltrated the upper echelons of our society and structure. Trust no one.

# Changelog

:cl:
- add: The deep sleepers awaken.


VividPups:
This would have to have unique  obj's and be HEAVILY RP dependent

VividPups:
The biggest problem i see is. That something happens on shift. And well now you have a trusted sec memeber with a long arm that can gun down warden or HOS easily and get access to the armory as if easy. Or get an Emag. I believe this would also require restricted things in the Antag uplink

M3739 (commented):
That... Is going to be a big no from me.

PR #1214, opened 2025-07-14
Expanding Lathes - Done

Raikyr0:
ExpandingLathesPR
Picture to show it is working

# Description

Buffs the Emag by adding a few things to the emag recipes.
Letting you print medfab things from an emagged Protolathe, if its medfab only it's no different from secfab only.
Added stuff that was missing from the protolathe, like the energy halberd (it's not that good).
Added security borg modules to emagged exosuit fab, and also a rather expensive syndie module. (worse then the sec kill module but concealable on non-sec borgs) 
Added basic printable secfab guns to emagged autolathe, good luck getting the plasteel needed. (One could argue this is fine because protolathes can already print advanced laser weapons.)

Why?
Still need to buff antags more. Sec still too strong in comparison.

# TO (someone else should) DO 

- [ ] Add Syndicate Assault Borg recipe to make from an emagged exosuit fab
- [ ] Add Syndicate Assault Quad Borg

# Changelog

:cl:
- add: Added expanded emag recipes list.


Resources/Prototypes/Entities/Structures/Machines/lathe.yml
```diff
@@ -236,6 +236,8 @@
     - MagazineMagnum # Goobstation
     - MagazineMagnumLeverRifleEmpty # Goobstation
     - MagazineMagnumLeverRifle # Goobstation
+    - WeaponRifleNovaliteC1 # Floofstation
+    - WeaponRifleGestio # Floofstation
```
M3739: Remove these lines for now. I'd like to keep the autolathe's recipes reserved for printing munitions and simple items. One could even argue that the firearms are in the on-board internal storage of the secfab, which isn't touched by networking whatsoever.

However... I can imagine things like the Halberd and the Combat knife being a good fit for the emag section of the autolathe, as both are simple in construction.

M3739 (changes requested):
Interesting. I would like second opinions from my peers on this PR.

Only one point of contention to address.

VividPups:
Not good my Ass

Resources/Prototypes/Entities/Structures/Machines/lathe.yml
Raikyr0: It's interesting because the protolathe lets them print basically all the laser weapons including the advanced laser pistol. Why would conventional firearms be more of an issue.

M3739 (approved):
Alrighty, looks good to me so far.

M3739:
I already sent my approval from prior. Seems like neither of the maints has checked it out yet.

Raikyr0:
> I already sent my approval from prior. Seems like neither of the maints has checked it out yet.

With Nikko still on break, there's really just 2 maintainers.
It's been here for weeks and Mnemo hasn't reviewed it so I don't think that's happening.

PR #1216, opened 2025-07-14
Fix Mood Being Moody

Mnemotechnician:
# Description
As it turns out, there is ONE code path in ApplyEffect where the effect can not apply under certain conditions (specifically, if you already have an instance of a non-categorized moodlet, it would refuse to add a new one). And my fix for mood flickering involved cancelling the previous timer regardless of whether the new moodlet applied or not (aka whether or not a new timer has started).

This PR aims to fix that issue.

<p>

https://github.com/user-attachments/assets/d446ea94-8d25-4ff9-8fe6-290768da6950

</p>
</details>

# Changelog
:cl:
- fix: Uncategorized mood effects (such as the "i can't breathe" one) should now properly get removed after their time runs out.


Mnemotechnician:
The PR has been playtested on Vulpstation and has proven functional

PR #1217, opened 2025-07-14
Stable Language Obfuscation

Mnemotechnician:
# Description
Language obfuscation was made stable across a single round, but varying between rounds. This was mostly a balancing decision to prevent people from simply memorizing every word in a language or using third-party tools to achieve the same. But since Floofstation is a roleplay-focused fork, it is less of an issue here. Allowing people to memorize certain words in foreign languages could lead to interesting roleplay and allow people to literally gradually "learn" a language, albeit only partially.

This will need discussion.

All this PR does is replacing the pseudo-random round-based seed in the RNG with a constant one.

# Changelog
:cl:
- tweak: Words in languages you don't understand should now be obfuscated the same way across all rounds.


ashkitten:
i really like this change. it's a shame it doesn't help with sign language at all.

M3739 (commented):
Could be interesting... There isn't like a Rosetta stone either. One would actually have to put in the work to figure out the language and what it could possibly mean.

I like the idea.

Memeji (approved):

M3739 (approved):

PR #1221, opened 2025-07-14
Show Carried Entities in Examine Text, Translate Numbers To Words

ashkitten:
# Description

contains a nasty hack but i checked and i'm pretty sure cuffs are the only thing that both shows a virtual item in your hand and is parented to you besides carrying shit?

---

# Media

image

---

# Changelog

:cl:
- add: People you're carrying will now show up in your examine popup.
- add: Literal numbers in speech are now translated to words.

### DIFF
--- a/Content.Shared/Hands/EntitySystems/SharedHandsSystem.Interactions.cs
+++ b/Content.Shared/Hands/EntitySystems/SharedHandsSystem.Interactions.cs
@@ -1,4 +1,5 @@
 using System.Linq;
+using Content.Shared.Cuffs.Components;
 using Content.Shared.Examine;
 using Content.Shared.Hands.Components;
 using Content.Shared.IdentityManagement;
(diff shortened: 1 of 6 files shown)


Content.Shared/Hands/EntitySystems/SharedHandsSystem.Interactions.cs
```diff
@@ -198,10 +199,20 @@ private void OnGetUsedEntity(EntityUid uid, HandsComponent component, ref GetUse
...
-            .Select(item => FormattedMessage.EscapeText(Identity.Name(item, EntityManager)))
-            .Select(itemName => Loc.GetString("comp-hands-examine-wrapper", ("item", itemName)))
+            .Select(entity => new Entity<VirtualItemComponent?>(entity, CompOrNull<VirtualItemComponent>(entity)))
+            .Where(entity => entity.Comp is null
+                // We can't access CarryingSystem in Shared, so here's a hacky way to check if we're carrying the object.
+                // A carried object is a child of the carrier, and is not a set of handcuffs on them.
+                || TransformSystem.GetParentUid(entity.Comp.BlockingEntity) == examinedUid
+                && !(TryComp(examinedUid, out CuffableComponent? cuffable)
+                    && cuffable.Container.ContainedEntities.Contains(entity.Comp.BlockingEntity)))
+            .Select(item => item.Comp?.BlockingEntity ?? item)
```
Mnemotechnician: Not a terrible way to do this, but why not just do
```suggestion
            .Select(entity => new Entity<VirtualItemComponent?>(entity, CompOrNull<VirtualItemComponent>(entity)))
            .Where(entity => entity.Comp is null || entity.Comp.BlockingEntity != examinedUid)
            .Select(item => item.Comp?.BlockingEntity ?? item)
```
? This would also show pulled entities and perhaps more. I don't think it's a bad idea to say that the mob is "holding handcuffs" while they are cuffed, either. This would help check if someone is still cuffed without opening the strip menu on them and without looking at their sprite.

ashkitten: i was already thinking about how to do the cuffs in a nice way, as well as dragged. i'm open to the idea but i don't want to use the same text as holding because it won't differentiate between someone actually being cuffed vs holding cuffs and cause confusion.

Mnemotechnician: > i was already thinking about how to do the cuffs in a nice way, as well as dragged. i'm open to the idea but i don't want to use the same text as holding because it won't differentiate between someone actually being cuffed vs holding cuffs and cause confusion.

I mean, if it's a big deal for you, you could add a new string or enum field to VirtualItemComponent that would indicate the type of the virtual item. Then you could change the code below to fetch slightly different strings based on the type of the virtual item (with the default one - if a type is not set - being "they are holding X"). But in my opinion, the existing text is just fine.

ashkitten:
image

Mnemotechnician (approved):
That is crazy.

ashkitten:
pushed a change that also sanitizes numbers in chat messages as words... this included rewriting the functions to operate on string representations of numbers since BigInteger is inaccessible from inside the content sandbox.

technically this is an unrelated change, but since it depends on and rewrites a huge part of the PR, i decided to include it. let me know if this is unacceptable.

M3739 (approved):
Seems to work well.

M3739:
image

As it turns out, merging this was an lapse in judgement on my end.

ashkitten:
oh my god. i never imagined it would affect binary.

PR #1224, opened 2025-07-16
Fix Layering of Vulpkanin Tails 

Mnemotechnician:
# Description
Splits the standard vulpkanin tails into two layers (as described here: https://discord.com/channels/1255902263667331193/1255992182880473269/1390404433082323045) to prevent them from drawing over skirts and other outfits that are not skin-tight. Currently only affects standard vulpkanin tails (vulp, vulp_alt, and their tip, wag, and alt variants). If I end up writing a script I may finish the rest later since doing it manually is a lot of grunt work.

This however does not fill in the "leg gaps" left in the tail sprites by their original authors, but I don't think it's worth the effort, since it will only matter if the tailed person loses their legs (in which case they will *probably* have better things to worry about than their tail looking weird)

<p>

vulp

image

image

vulp_alt_tip

image

image

wagwagwag

https://github.com/user-attachments/assets/fca96944-2867-4060-a689-14050c08155b

</p>
</details>

# Changelog
:cl:
- fix: Normal vulpkanin tails (vulpkanin, vulpkanin alt, vulpkanin tip, vulpkanin alt tip, and their wagging variants) will no longer draw over skirts and dresses!
- fix: Vulpkanin (fade) tails will no longer turn into Vulpkanin (tip) when wagging.


Mnemotechnician:
I *think* I checked all the tails and made sure they all function correctly, but I need someone to double-check because I kept making mistakes due to the large amount of grunt work.

Mnemotechnician:
Do not merge. Some fixes need to be pulled from Vulpstation where this PR has been tested.

Mnemotechnician:
It is done, now existing vulp characters won't break and will import correctly. The marking system "validates" existing markings on humanoid, and the way it used to do it was replacing the entire color list of a marking if its length was different from the number of sprites on the actual marking. I made it extend/trim the list instead.

Memeji (approved):

M3739 (approved):

PR #1225, opened 2025-07-16
Don't Serialize DeviceNetworkComponent.Configurators

Mnemotechnician:
# Description
Why, just why? This field stores all network configurators that are currently being used to link the owner entity to other elements of a device network. This field has no use between map resets, but SS14 will attempt to serialize it anyway, which will sometimes result in `configurators: [ invalid ]` because the network configurator entity has been deleted.

# Changelog
No cl


Mnemotechnician:
Will keep it as a draft, need to make sure existing maps will load succesfully

Mnemotechnician:
This has also been tested on Vulpstation and worked fine.

M3739 (approved):

Memeji (approved):

PR #1228, opened 2025-07-18
Fix Internals on Pets

Mnemotechnician:
# Description
For some reason the internals system would always attempt to resolve the HandsComponent on the entity whose internals are being toggled, which would always fail on animals.

image

image

# Changelog
:cl:
- fix: Animals such as foxes should once again be able to wear internals.

### DIFF
--- a/Content.Server/Body/Systems/InternalsSystem.cs
+++ b/Content.Server/Body/Systems/InternalsSystem.cs
@@ -266,7 +266,9 @@ public sealed class InternalsSystem : EntitySystem
         // 3. in-hand tanks
         // 4. pocket/belt tanks
 
-        if (!Resolve(user, ref user.Comp1, ref user.Comp2, ref user.Comp3))
+        // Floofstation - do not require the entity to have hands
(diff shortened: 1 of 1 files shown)


M3739 (approved):
This does have the side effect of allowing players to toggle their internals on and off without any hands attached to their bodies.

But I think we can deal with that quirk for now. Besides, it'll get a nice chuckle.

Mnemotechnician:
> This does have the side effect of allowing players to toggle their internals on and off without any hands attached to their bodies.

That's weird, I guess the system doesn't check `ActionBlockerSystem.CanInteract(User, Target)`. That in turn means that players are *already* able to toggle their internals while having no hands, so long as they have the Hands component.

M3739:
> That's weird, I guess the system doesn't check `ActionBlockerSystem.CanInteract(User, Target)`. That in turn means that players are _already_ able to toggle their internals while having no hands, so long as they have the Hands component.

Right... Ah shitmed, my beloved. At least during testing, being cuffed worked as expected.

M3739:
Been sitting for a bit. Merging now.

PR #1231, opened 2025-07-22
Interdyne Herbals Tweak - Done

Raikyr0:
# Description

Changes the Iron inside Herbals to Saline.
Why?
So it works on races with Copper based blood.

# Changelog

:cl:
- tweak: Changed one of the secret ingredients inside Interdyne Herbals


M3739:
This would also have the side-effect of hydrating those who consume said herbals. The dylovene included in the herbal does off-set the resulting poison damage in conjunction with the omnizine when this cigarette is smoked by an arachnid.

Raikyr0:
> This would also have the side-effect of hydrating those who consume said herbals. The dylovene included in the herbal does off-set the resulting poison damage in conjunction with the omnizine when this cigarette is smoked by an arachnid.

Smokables add 0.15 units of the total mix per second, meaning it would be 0.05 saline which is hardly noticeable in terms of hydration.
The reason why I changed iron to saline was so it would also increase blood level on arachnids and resomi too. The poison is inconsequential due to the dylovene already.

M3739:
Interesting. With that assessment in mind, I support this change.

M3739 (approved):

Memeji (approved):
Should be good.

PR #1233, opened 2025-07-23
Floofjaran 

Cprn-Sprkl:
# Description

A port of Tajara ##1647 (https://github.com/Simple-Station/Einstein-Engines/pull/1647) for Floofstation.
Adds Tajara and two of their languages. Increases their customization options to be in line with other Floof species including more markings and wider fur colour range.

The traits for the Tajaran languages have been modified to be in line with other language traits in terms of points, and have a wider pool of acceptable species that can take them.

---

# Changelog

:cl: Sprkl
- add: The Tajara have arrived!


Memeji (approved):
Tests pass, think it should be good.
Will need the other's input.

Resources/Prototypes/_EE/Entities/Mobs/Species/tajaran.yml
```diff
@@ -0,0 +1,129 @@
...
+      60: 0.85  # 0.7 is base speed.
+      80: 0.75  # 0.5 is base speed.
+  - type: Carriable # Carrying system from nyanotrasen.
+  - type: Butcherable
+    butcheringType: Spike
+    spawned:
+    - id: FoodMeatHuman
+      amount: 5
+  - type: TemperatureProtection # fur is a good insulator, actually
+    coefficient: 0.1
```
M3739, line 42: Restrict them from taking the `TemperatureTolerance` perk as they already have it innately.

M3739 (changes requested):
A few things to address.

Interestingly enough, it would seem that the languages have uh... a bit of a nuance to them: 
image

It seems like it is outputting both the verb and the actual sentence itself, instead of showing only the verb for non-understanding observers, and only the actual message for those that understand it.

Other than one other issue, it seems to check out on my end.

Cprn-Sprkl:
The original PR had five new Tajaran languages, which was deemed too many for one species. I cut it down to the two most interesting ones and opened them up to any species with both ears and tails.

One is a pure sign language, the other is a combination of body language and vocalizations. That's why it outputs a bit differently than other languages.

image
image

M3739 (approved):
I see. In that regard, I'll take it as is. I just had to quickly verify that it was being handled by radio communications properly. A little bit cursed as it only transmits the vocal part, but it works.

PR #1237, opened 2025-07-24
Paradox Spawn Rate Increase

Raikyr0:
# Description

Paradoxes kinda rare, so this should increase their spawns.

# Changelog

:cl:
- tweak: Increased Paradox Spawn Rate

### DIFF
--- a/Resources/Prototypes/Nyanotrasen/GameRules/events.yml
+++ b/Resources/Prototypes/Nyanotrasen/GameRules/events.yml
@@ -39,7 +39,7 @@
   components:
   - type: StationEvent
     weight: 15
-    reoccurrenceDelay: 15
+    reoccurrenceDelay: 25
     minimumPlayers: 5
     earliestStart: 5
   - type: MidRoundAntagRule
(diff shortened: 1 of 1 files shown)


Mnemotechnician (changes requested):
With a reoccurence delay of 25 minutes, this event really needs to have a limited number of max occurences, otherwise a 3 hour secret shift will turn into a doppelganger fest (when the chaos multiplier is high, event schedulers usually end up exhausting the event pool and just spitting out every event that qualifies under the current conditions)

Additionally, please add a # Floofstation comment to indicate that this change belongs to floof.

Raikyr0:
> With a reoccurence delay of 25 minutes, this event really needs to have a limited number of max occurences, otherwise a 3 hour secret shift will turn into a doppelganger fest (when the chaos multiplier is high, event schedulers usually end up exhausting the event pool and just spitting out every event that qualifies under the current conditions)
> 
> Additionally, please add a # Floofstation comment to indicate that this change belongs to floof.

My issue with the limit on paradoxes is the people who load into a paradox, see they've loaded into someones they either don't know or has already cryo'd and go cryo themselves. Which happens super often.

But I can up the re-occurrence timer if needs be.

PR #1238, opened 2025-07-24
Halved Ratking Spawn Rate

Raikyr0:
# Description

Halved Ratking Spawn Rate, due to it being PER VENT it resulted in bad luck rounds where there's like 4 rat kings in the ghost roles list.

# Changelog

:cl:
- tweak: Halved Ratking Spawn Rate

### DIFF
--- a/Resources/Prototypes/GameRules/events.yml
+++ b/Resources/Prototypes/GameRules/events.yml
@@ -270,7 +270,7 @@
 #      prob: 0.001
     specialEntries:
     - id: SpawnPointGhostRatKing
-      prob: 0.001
+      prob: 0.0005
 
 - type: entity
   id: CockroachMigration


VividPups:
dont halve it put a max on it

Raikyr0:
> dont halve it put a max on it

I don't think I CAN put a max on it, since it checks per vent
I had 2 xenos pop out of the same vent, meaning this shit would check per vent

That said, Rat kings still have a mid round antag spawn.

Mnemotechnician:
This particular probably only applies to a special entity with the ID SpawnPointGhostRatKing, are you sure you are changing the right thing? Usually maps have at MOST 2 or 3 of these. But regarding the "max" part, we can adjust the code to support a limit of how many entities can spawn.

Raikyr0:
> This particular probably only applies to a special entity with the ID SpawnPointGhostRatKing, are you sure you are changing the right thing? Usually maps have at MOST 2 or 3 of these. But regarding the "max" part, we can adjust the code to support a limit of how many entities can spawn.

I do believe ID SpawnPointGhostRatKing is the invisible spawn marker for Rat kings which generates from the event over an air vent.
The issue of too many rat kings spawning has been one that's been needed to be fixed for a long time.

Mnemotechnician:
Apparently you're right and it's even worse: since it's a special entry, it's *guaranteed* to spawn at least once:
https://github.com/Floof-Station/Floof-Station/blob/47d3d52a87b247c3c9a111bb96c74c440f572d40/Content.Server/StationEvents/Events/VentCrittersRule.cs#L55

I think it's best to move it under normal spawns instead of special

PR #1252, opened 2025-07-27
Security Loadout

Raikyr0:
SecurityLoadoutPR

# Description

This is exactly the same Security Loadout PR which Floofmin declined earlier in the year, but since he's not running the show anymore, I'll give it another go.

Anyway, upstream's security loadout where the weapons are rounded up and separated into their own tab.
I added a variant of the N1984 with a lower fire rate. The original reason we didn't allow it was due to the fact it had a rather fast fire rate.
Variant of the energy shotgun which doesn't automatically recharge.
Altered Security Energy Cutlass which does pure heat.  (worse then a combat knife but leaves no blood spills)

# Changelog
:cl:
- add: Added variants of the N1984 and Energy Shotgun
- tweak: Seperated weapons into their own tab for security
- add: Added Energy Cutlass for Security


Memeji:
Will still be a no to the N1984

Raikyr0:
> Will still be a no to the N1984

even after the nerf to fire rate which was the main issue?

Also I disabled it again.

Memeji:
> > Will still be a no to the N1984
> 
> even after the nerf to fire rate which was the main issue?
> 
> Also I disabled it again.

It wasn't just the fire rate. While it is helpful to reduce it, the N1984 was meant to be a Centcom Official's handgun.
I'd prefer the N1984 to be iconic to CC, in my opinion.

M3739:
> It wasn't just the fire rate. While it is helpful to reduce it, the N1984 was meant to be a Centcom Official's handgun. I'd prefer the N1984 to be iconic to CC, in my opinion.

I concur. If I recall correctly, the N1984 was *not* originally a security sidearm. It was a CC official's sidearm for self-defense, suitable for a commanding officer.

We seek to return back to it being a sidearm issued exclusively to CC personnel. Of course, the men in green get the fancy stuff, but that's the perk of being part of the central command of the entire Floof sector. Or any sector as a matter of fact.

Raikyr0:
Either way, I've disabled it.

Raikyr0:
> This pull request has been automatically marked as stale due to inactivity.

Unstale this right now ya shit bot.

Resources/Prototypes/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml
```diff
@@ -914,4 +914,49 @@
...
     autoRechargePause: true
-    autoRechargePauseTime: 10
\ No newline at end of file
+    autoRechargePauseTime: 10
+
+- type: entity # Floofstation - Non Recharging Variant
+  name: energy shotgun
+  parent: [BaseWeaponBattery, BaseGunWieldable]
+  id: WeaponSecurityEnergyShotgun
+  description: A one-of-a-kind prototype energy weapon that uses various shotgun configurations. It offers the possibility of both lethal and non-lethal shots, making it a versatile weapon. This variant lacks the iconic self charging battery.
```
M3739: _"one-of-a-kind"
"This variant"_

Stating that there is a variant of a weapon, implies there are other verisons, which directly goes against the earlier declaration of this item being a "one-of-a-kind prototype". After some research, this is what is known as a <ins>**semantic contradiction**</ins>.

**Rewrite this description.**

Resources/Prototypes/_Goobstation/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml
```diff
@@ -60,4 +60,14 @@
...
+
+- type: entity
+  name: EG-4 energy revolver
+  parent: WeaponEnergyRevolver
+  id: WeaponEnergyRevolverSecurity
+  description: A highly advanced energy revolver capable of firing both lethal and disabling bullets. The serial number on the handguard marks this gun as belonging to an NT Security Officer.
+  suffix: Security Loadouts
+  components:
+  - type: GuideHelp
+    guides: [ SecurityWeapons ]
```
M3739, line 73: This recharges by itself. Albeit, quite slowly. Eliminating the need to restock/recharge.

I don't approve of this being an option in the standard issue loadout of security personnel in general.

Resources/Prototypes/CharacterItemGroups/Jobs/Security/uncategorized.yml
```diff
@@ -127,6 +127,14 @@
       id: LoadoutSecurityArgenti
     - type: loadout
       id: LoadoutSecurityArgentiNonLethal
+    - type: loadout
+      id: LoadoutSecurityEquipmentEnergySword
+    - type: loadout # Floofstation - A little special something for the weebs
+      id: LoadoutSecurityHoSKatanaSheath
```
M3739: I'd... rather this be bearing a prototype where it *isn't* a "antique katana sheaths". The 'antique' part for a standard issue option just doesn't sit right for me.

There is also the fact it's called a *daisho*, but that's merely worth noting.

Other than that, I would be fine if this wasn't using the same thing that the HoS uses.

M3739 (changes requested):
I... am a bit unsure on whether or not the disabler SMG is alright for a security officer to have as part of their options regarding standard issue gear, especially the big question of whether or not it is considered a **long-arm.**

There are also some other things that need to be addressed before this earns my approval.

Resources/Prototypes/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml
Raikyr0: I can probably change the description.

Resources/Prototypes/CharacterItemGroups/Jobs/Security/uncategorized.yml
Raikyr0: I'd have to make a prototype with a different name but yea.

Raikyr0: nevermind I found a non hos prototype that makes it easier

Resources/Prototypes/_Goobstation/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml
Raikyr0: I would say it's standard issue, the loadout cost is 10 which is supposed to be the limiting factor.

M3739: Indeed.

After some more testing, I have discovered that damage wise, it worse than the PDW. Less projectile damage, coupled with a slower fire rate. The only thing that does in fact make this pretty expensive is the self-recharging factor....

Very well. there is a choice to be made when it comes to loadouts. This recharges slow enough to where it isn't a full replacement for other weapons, and due to it being weaker performance wise, will reserve it's use for routine patrols and vent critter culling.

But for larger, more dangerous threats? You will be better off using this as a backup weapon rather than a complement to your primary armament. And it's use of energy bolts instead of lasers will require the user to aim their shots a bit more, and it isn't as capable of shooting through grilled windows like a traditional laser.

I'd reckon, this will be a favorite among prison guards. That being said... I'd reckon it is good right now.

M3739 (approved):
I'd say the current state of the PR is now suitable for integration into the current environment. I encourage discourse regarding the new service weapon options, especially regarding concerns with them.

I hereby voice my approval.

M3739:
If you haven't already, please merge master into this PR. It's been awhile, it seems.

Memeji (commented):
I'm unsure about the energy sword and energy shotgun. I'm under the impression these security weapons are available to Sec Officers right?

Raikyr0:
> I'm unsure about the energy sword and energy shotgun. I'm under the impression these security weapons are available to Sec Officers right?

The Security Energy Cutlass will be available for all security with a hefty hour requirement (effectively a prestige weapon), it has no reflect chance, and does about the same damage as combat knife in pure heat damage, as I recall also swings a little slower. Onis will not benefit from using this as it doesn't have any physical damage to increase. Think of it as a Combat Knife which doesn't bleed people and is kinda loud.

As for the Energy Shotgun, it's for the Warden or HoS only. It's a non-self-charging variant of the Energy Shotgun. It's weaker than the normal energy shotgun in that regard.

Memeji (approved):
Sounds good, I'll give it an approval.

Chokovit:
The EG-4 Energy Revolver is far too strong for a basic sec load out. Just giving it 10 points isn't enough. There's a reason that gun is limited to only the BSO. It is a self recharging PDW. That needs to be removed.

META-mahn:
> The EG-4 Energy Revolver is far too strong for a basic sec load out. Just giving it 10 points isn't enough. There's a reason that gun is limited to only the BSO. It is a self recharging PDW. That needs to be removed.

I've been thinking a lot about EG-4 with regards to the testing fork I have (which kind of just became a PVP/mapping sandbox) and my conclusion is that the EG-4 can be very scary, but the main power is if/when all of Sec takes it as a loadout option, or if an antag gets their hands on it.

If all of Sec takes the EG-4 then the optimal Station strat for any Station vs Antag encounter is for Sec to spam forcewalls or any other transparent object and repeatedly poke the enemy. The counterplay to that would be grenade/c4/chinalake spam since we don't have upstream's Hristov buffs yet, or really anything for antags to contest longer ranges with except for a chinalake.

On the other hand it's well known theory for nukies to take the captain's laser ASAP. It's the one tool that lets them fight against station-side attrition and the station wins the attrition fight always, unless if nukies have some form of infinite ammo. EG-4 isn't exactly the captain laser but if every nukie has one then you can imagine what happens.

I'm not going to champion for it to stay in or its removal; just laying out the theory for this so someone else who's more involved with the process can make their decisions.

Raikyr0:
It's supposed to be a prestige weapon at 72 hours. I considered a much higher time requirement. However at the time of creation alot of people left due to the floof exodus.

PR #1253, opened 2025-07-27
Syndicate Uplink Buff

Raikyr0:
# Description

Under the request of the Head Committee, Syndicate Uplink items have been made cheaper.
Please discuss if needs be but remember this is supposed to be a huge buff to Traitors, balance if over tuned.

# Changelog
:cl:
- tweak: Several syndicate items now cost less TC.

### DIFF
--- a/Content.IntegrationTests/Tests/StoreTests.cs
+++ b/Content.IntegrationTests/Tests/StoreTests.cs
@@ -79,11 +79,11 @@ public sealed class StoreTests
             var discountComponent = entManager.GetComponent<StoreDiscountComponent>(pda);
             Assert.That(
                 discountComponent.Discounts,
-                Has.Exactly(3).Items,
-                $"After applying discount total discounted items count was expected to be '3' "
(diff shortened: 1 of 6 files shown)


M3739 (changes requested):
YAML linter fail is valid.
```
Error: /Prototypes/Catalog/uplink_catalog.yml(775,3)  Field "saleLimit" not found in "Content.Shared.Store.ListingPrototype".
```

Raikyr0:
> YAML linter fail is valid.
> 
> ```
> Error: /Prototypes/Catalog/uplink_catalog.yml(775,3)  Field "saleLimit" not found in "Content.Shared.Store.ListingPrototype".
> ```

Weird the sales limit code still doesn't work, I'll just use the other sales limit one. Also I swapped it to ONLY give 1 key instead of a 2fer box.

Memeji (approved):
Think the value changes seem fine.

M3739 (approved):

PR #1255, opened 2025-07-29
Free the Fridge

Raikyr0:
# Description

This unlocks the fridge (all the locked ones) for anyone to access. This includes the one on the Pathfinder which Salvage couldn't open. Why do fridges even have locks in the first place?
Secret - The fridge is now immune to nukes (Indiana Jones reference from upstream).

# Changelog
:cl:
- fix: The fridge on the Pathfinder is now operable for non-chefs.


M3739 (approved):

Mnemotechnician (approved):
I guess... fair?

PR #1258, opened 2025-07-29
Rodentia Diet - Done

Raikyr0:
# Description
Tweaked Rodentia to be able to eat chocolate and tea and whatever. Because wow different digestive abilities compared to other animals.

# Changelog
:cl:
- tweak: Cheesed to Eat you, Rodentia can now eat chocolate.

### DIFF
--- a/Resources/Prototypes/DeltaV/Body/Prototypes/rodentia.yml
+++ b/Resources/Prototypes/DeltaV/Body/Prototypes/rodentia.yml
@@ -13,7 +13,7 @@
     torso:
       part: TorsoRodentia
       organs:
-        heart: OrganAnimalHeart
+        heart: OrganHumanHeart
         lungs: OrganHumanLungs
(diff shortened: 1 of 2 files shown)


M3739 (commented):
I... don't necessarily see the merit behind making them able to eat theobromine and allicin without ill effect.

Raikyr0:
> I... don't necessarily see the merit behind making them able to eat theobromine and allicin without ill effect.

The reason I was given by Lily was that Rodents have stomachs similar to human stomachs which allow them to digest theobromine and allicin, thus it would be factually incorrect for Rodentia to be allergic.

M3739 (approved):
I see.

After doing some research, turns out they have similiar tolerances for toxicity of allicin and theobromine to humans. Neat.

Learn something new everyday.

Mnemotechnician (approved):

PR #1261, opened 2025-07-30
Minor QoL Patch

Raikyr0:
MinorQoLPR

# Description

Makes Wires instant build, waiting 2 seconds for every single wire was unacceptably long, even processing ORE is faster.
Re-enabled Water Vapor Canisters to be bought from Logistics.
Adds Pig and Chicken Cubes and the method to print them from a biofab.
Fixes Pigs to give bacon.

# Changelog
:cl:
- tweak: Tweaked wires to build almost instantly
- tweak: Re-enabled Water Vapor Canisters from Logistics
- fix: Pigs give bacon now. Weird they didn't before.
- add: Added Pig and Chicken Cubes.

### DIFF
--- a/Resources/Prototypes/Recipes/Lathes/tools.yml
+++ b/Resources/Prototypes/Recipes/Lathes/tools.yml
@@ -42,7 +42,7 @@
   id: CableStack
   result: CableApcStack1
   category: Parts
-  completetime: 2
+  completetime: 0
   materials:
     Steel: 30
 
@@ -50,7 +50,7 @@
   id: CableMVStack
   result: CableMVStack1
   category: Parts
-  completetime: 2
+  completetime: 0
   materials:
     Steel: 30
 
@@ -58,7 +58,7 @@
   id: CableHVStack
   result: CableHVStack1
   category: Parts
-  completetime: 2
+  completetime: 0
   materials:
     Steel: 30
 
--- a/Resources/Prototypes/Recipes/Lathes/tools.yml
+++ b/Resources/Prototypes/Recipes/Lathes/tools.yml
@@ -42,7 +42,7 @@
   id: CableStack
   result: CableApcStack1
   category: Parts
-  completetime: 0
+  completetime: 0.2
   materials:
     Steel: 30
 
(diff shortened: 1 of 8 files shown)


Mnemotechnician (changes requested):
Cables should not be able to be printed instantly... 0.5 or maybe 0.2 seconds per cable, yeah, that would make sense. But allowing someone to print up 100 thousand wires, using up an ore silo's worth of steel, within 0 seconds? That's absurd.

Other than that, please move the new prototypes under the _Floof namespace.

Raikyr0:
> Cables should not be able to be printed instantly... 0.5 or maybe 0.2 seconds per cable, yeah, that would make sense. But allowing someone to print up 100 thousand wires, using up an ore silo's worth of steel, within 0 seconds? That's absurd.
> 
> Other than that, please move the new prototypes under the _Floof namespace.

One of the most annoying things is waiting 2 seconds for every wire, more so when I'm trying to repair an IPC with a burn damage count of upwards of 1000. That said, you can technically already QUEUE 100000 wires in the lathe and it uses up that material already. Queuing stuff uses the material.

Yeah I'll move the prototypes tho.

Resources/Prototypes/Recipes/Lathes/tools.yml
```diff
@@ -42,23 +42,23 @@
...
     Steel: 30
 
 - type: latheRecipe
   id: CableHVStack
   result: CableHVStack1
   category: Parts
-  completetime: 2
+  completetime: 0
   materials:
     Steel: 30
```
M3739: I would prefer if the `completetime` was something like 0.5 as Mnemo had suggested. Letting things  like wires instantly build doesn't make sense.

M3739 (changes requested):

Resources/Prototypes/Recipes/Lathes/tools.yml
Raikyr0: Makes the same sense as ore processors instantly processing hundreds of sheets of steel and glass.

M3739 (approved):
I still would have liked for it to just be a very short `completetime`.... but oh well.

Mnemotechnician (approved):
After some discussion we've decided to add a small completion time to cables. They are still going to take 10 times less time to craft than before.

PR #1262, opened 2025-07-31
Marked Traits Update: No Free Points

Cprn-Sprkl:
# Description

These traits do not take trait slots, were intended for just opting into wanted gameplay, and should not incentivize being taken by people that just want points but not the consequences.

So this removes the points, and bad incentive, for taking these traits.

---

# Changelog

:cl: Sprkl
- tweak: Marked traits no longer give trait points.

### DIFF
--- a/Resources/Prototypes/_Floof/Traits/marked.yml
+++ b/Resources/Prototypes/_Floof/Traits/marked.yml
@@ -1,7 +1,7 @@
 - type: trait
   id: TraitorKillTarget
   category: Marked
-  points: 2
+  points: 0
   slots: 0
   functions:
   - !type:TraitAddComponent
@@ -12,7 +12,7 @@
 - type: trait
   id: TraitorTeachTarget
   category: Marked
-  points: 1
+  points: 0
   slots: 0
   functions:
   - !type:TraitAddComponent


Mnemotechnician:
But... why should we punish those players who *do* want the extra trait points just because *some* people take them out of greed without understanding the consequences?

Cprn-Sprkl:
> But... why should we punish those players who _do_ want the extra trait points just because _some_ people take them out of greed without understanding the consequences?

I don't believe it is a punishment, I believe it was an error to offer these points in the first place. These traits were conceived as a way for people who want to engage in a certain type of gameplay to be able to on Floof again; restoring non-Command teach targets and the threat of RR for the people that want that experience.

Offering points, especially on traits that do not take trait slots, corrupts this a bit by adding a perverse incentive that should not have been there in the first place. It adds a dash of min max thought to something that was supposed to be just an opt in to a certain style of gameplay. Originally these weren't even going to be traits, that just ended up being the best way I could find to implement them without touching the server database.

The appearance of people taking these just for the points also hurts everyone who takes them for the gameplay. If one person is speaking up that they won't follow through on RR objectives because of people crashing out about them in EOR or dead chat then they are likely not the only one. If the idea of people only taking these for the points makes antag players hesitant to execute on the objectives then people aren't getting the gameplay they opted in to.

Mnemotechnician (approved):
That's a fair point.

M3739 (approved):
Well said. I agree.

PR #1265, opened 2025-08-01
Medical Tags

Raikyr0:
MedicalTagPR

Ignore the picture, new text is fixed.

# Description

Added Medical Tags for traits that would require special treatment or medication.

And yes I like wall of text examines.

# Changelog
:cl:
- add: Added Medical Examine text.
- fixed: Hypno examine text.


Resources/Locale/en-US/traits/misc.ftl
```diff
@@ -4,3 +4,9 @@ examine-bionic-arm-message = {CAPITALIZE(POSS-ADJ($entity))} arms emit an ever p
...
 examine-pry-arm-message = {CAPITALIZE(POSS-ADJ($entity))} arms emit an ever present faint chirp of servomotors.
+examine-narcolepsy-message = {CAPITALIZE(POSS-ADJ($entity))} wears a medical tag which notes that this individual suffers from narcolepsy.
+examine-unrevivable-message = {CAPITALIZE(POSS-ADJ($entity))} wears a medical tag which notes that this individual can't be revived through conventional methods.
+examine-uncloneable-message = {CAPITALIZE(POSS-ADJ($entity))} wears a medical tag which notes that this individual can't be cloned.
+examine-blood-deficiency-message = {CAPITALIZE(POSS-ADJ($entity))} wears a medical tag which notes that this individual suffers from blood deficiency.
+examine-redshirt-message = {CAPITALIZE(POSS-ADJ($entity))} wears a medical tag which notes that this individual suffers from a condition commonly known as redshirt syndrome.
+examine-brittle-bone-disease-message = {CAPITALIZE(POSS-ADJ($entity))} wears a medical tag which notes that this individual suffers from osteogenesis imperfecta or more commonly known as brittle bone disease.
```
M3739: Results in "Her wears"?....

Rewrite this to actually use the proper pronoun.

M3739 (changes requested):

Resources/Locale/en-US/traits/misc.ftl
Raikyr0: Fixed it.

Mnemotechnician:
As stated before, I don't think this change is good. It will bloat the examine menu for no good reason.

ArcanissKenku:
I reckon' intergrate bracelets or something instead so people can wear medical bracelets on their loadout rather than making it a trait

Memeji:
I agree with Mnemo and ArcanissKenku here.

The wall of text is a bit much for the basic examine.
It would fit more in the health examine as the information makes sense there.

Try reducing the bloat by making it list the traits, like:
> A medical tag hangs from their neck which notes that this individual suffers from narcolepsy, blood deficiency, osteogenesis imperfecta... etc
Possibly remove the descriptions as well, only provide the trait name, this may help reduce the clutter.

That or possibly integrate it into bracelets, or one bracelet that lists the conditions.

Raikyr0:
> I agree with Mnemo and ArcanissKenku here.
> 
> The wall of text is a bit much for the basic examine. It would fit more in the health examine as the information makes sense there.
> 
> Try reducing the bloat by making it list the traits, like:
> 
> > A medical tag hangs from their neck which notes that this individual suffers from narcolepsy, blood deficiency, osteogenesis imperfecta... etc
> > Possibly remove the descriptions as well, only provide the trait name, this may help reduce the clutter.
> 
> That or possibly integrate it into bracelets, or one bracelet that lists the conditions.

I've cut down on the description a bit, altho I'm unsure how to get a single examine to take all the negative traits. 
That said, no one checks the health examine when medical has a health scanner and as such would get missed.

Memeji:
> > I agree with Mnemo and ArcanissKenku here.
> > The wall of text is a bit much for the basic examine. It would fit more in the health examine as the information makes sense there.
> > Try reducing the bloat by making it list the traits, like:
> > > A medical tag hangs from their neck which notes that this individual suffers from narcolepsy, blood deficiency, osteogenesis imperfecta... etc
> > > Possibly remove the descriptions as well, only provide the trait name, this may help reduce the clutter.
> > 
> > 
> > That or possibly integrate it into bracelets, or one bracelet that lists the conditions.
> 
> I've cut down on the description a bit, altho I'm unsure how to get a single examine to take all the negative traits. That said, no one checks the health examine when medical has a health scanner and as such would get missed.

I believe that's due to how there isn't much of a reason to use the health examine. This would give a reason to check the health examine more imo.

I will see if I can find any examples of systems that use a list like described. I believe I saw something before.

Raikyr0:
> > > I agree with Mnemo and ArcanissKenku here.
> > > The wall of text is a bit much for the basic examine. It would fit more in the health examine as the information makes sense there.
> > > Try reducing the bloat by making it list the traits, like:
> > > > A medical tag hangs from their neck which notes that this individual suffers from narcolepsy, blood deficiency, osteogenesis imperfecta... etc
> > > > Possibly remove the descriptions as well, only provide the trait name, this may help reduce the clutter.
> > > 
> > > 
> > > That or possibly integrate it into bracelets, or one bracelet that lists the conditions.
> > 
> > 
> > I've cut down on the description a bit, altho I'm unsure how to get a single examine to take all the negative traits. That said, no one checks the health examine when medical has a health scanner and as such would get missed.
> 
> I believe that's due to how there isn't much of a reason to use the health examine. This would give a reason to check the health examine more imo.
> 
> I will see if I can find any examples of systems that use a list like described. I believe I saw something before.

As for having it on an item, I'm not sure how to get an item to list all that, be linked to the person, and have a slot for them to fit it if they still want a necklace or gloves.

I kinda envy the Monolith style body inventory which has so many slots.

Arbitrated:
We think a simple solution to the initial "wall of text" would be to note that a character has a Medical Bracelet, which can then be examined in the medical examine. 

An idea for the text could be something like...

"They have a medical bracelet on their wrist. Use Medical Examine to read the bracelet." 

It's slightly OOC, sure, but this would at least ensure that people know how to "read" the bracelets.

Then the Medical Check can state out their health-affecting traits (and possibly Allergies, if we wanted to add those to Floof).

M3739:
I agree with the method of it being instead localized to a item, such as a medical bracelet, instead of being tied directly to the initial inspect. Many people, myself included, not only quickly inspect things, but sometimes also periodically to check what the person is holding as they fiddle around with their items.

So to be greeted with a huge wall of text on an inspect everyone uses to see each other in fine detail is... not ideal. Not to mention the atypical behavior of anything that is measurable in any capacity: **someone will indeed attempt to see how cluttered they can make their inspect window,** and of course, that would only exacerbate the issues aforementioned.

Indeed, why clutter a _general_ inspect window with information that only primarily concerns those working in _medical_?

M3739:
That being said, I will be closing this PR. The idea is sound, and something that I encourage to be revisited at some point, but this way of doing it is not only dirty, but introduces an _undue_ amount of clutter to the general inspect of players. Clutter, that as mentioned prior, is the primary concern of medical. It isn't general information, and it isn't what I would consider to be discerned from someone at first glance without additional tooling, either.

PR #1266, opened 2025-08-01
Suffer Maxxing

Raikyr0:
# Description

We love suffering.
Therefore, I've made Brittle Bones and Redshirt un-mutually exclusive. (Looking at you Kip Yip)
And I've made a brand new, Death Acidifier trait, which starts you with a death acidifer (inside you). For when you want to be TRULY unrevivable, or roleplay a former nukie or whatever you desire. (Flick would take this)
I wanted to make a BSO Linelife variant but we're lacking that code.

# TODO

- [ ] Check if Death Acidifer Trait is working

# Changelog

:cl:
- add: Added Death Acidifier Trait
- tweak: You can now take Brittle Bones and Red Shirt together again.

### DIFF
--- a/Resources/Prototypes/_Floof/Traits/physical.yml
+++ b/Resources/Prototypes/_Floof/Traits/physical.yml
@@ -301,5 +301,6 @@
     - Borg
     - MedicalBorg
   functions:
-  - !type:AddImplantSpecial
-    implants: [ DeathAcidifierImplant ]
+    - type: AutoImplant
+      implants:
+      - DeathAcidifierImplant
--- a/Resources/Prototypes/_Floof/Traits/physical.yml
+++ b/Resources/Prototypes/_Floof/Traits/physical.yml
@@ -301,5 +301,5 @@
     - Borg
     - MedicalBorg
   functions:
-  - !type:AddImplant
+  - !type:TraitAddImplant
     implants: [ DeathAcidifierImplant ]
--- a/Resources/Prototypes/_Floof/Traits/physical.yml
+++ b/Resources/Prototypes/_Floof/Traits/physical.yml
@@ -301,6 +301,5 @@
     - Borg
     - MedicalBorg
   functions:
-    - type: AutoImplant
-      implants:
-      - DeathAcidifierImplant
+  - !type:AddImplant
+    implants: [ DeathAcidifierImplant ]
--- a/Resources/Prototypes/_Floof/Traits/physical.yml
+++ b/Resources/Prototypes/_Floof/Traits/physical.yml
@@ -290,3 +290,16 @@
   - !type:TraitModifySilicon
     batteryDrainModifier: -0.25 # Floof - M3739 - #1209 - Roughly 32 minutes on a medium capacity power cell.
     
+- type: trait
+  id: TraitDeathAcidifier
+  category: Physical
+  points: -8
+  requirements:
+  - !type:CharacterJobRequirement
+    inverted: true
+    jobs:
+    - Borg
+    - MedicalBorg
+  functions:
+  - !type:AddImplantSpecial
+    implants: [ DeathAcidifierImplant ]
--- a/Resources/Locale/en-US/_Floof/traits/traits.ftl
+++ b/Resources/Locale/en-US/_Floof/traits/traits.ftl
@@ -88,3 +88,5 @@ trait-name-TraitorTeachTarget = Marked for Death
 trait-description-TraitorTeachTarget = Someone or something wants you taught a lesson. Traitors can recieve orders to kill you.
 # Target Consent Traits: End
 
+trait-name-TraitDeathAcidifier = Death Acidifer
+trait-description-TraitDeathAcidifier = For one reason or another, you acidify on death or have a death acidifer implant, possibly due to current or previous work.
--- a/Resources/Prototypes/Traits/physical.yml
+++ b/Resources/Prototypes/Traits/physical.yml
@@ -758,6 +758,8 @@
           damageContainers:
           - Biological
           - BiologicalMetaphysical # Floof - M3739 - #1006
+          - Inorganic
+          - Silicon
         - type: ShowHealthBars
           damageContainers:
           - Biological
--- a/Resources/Prototypes/Traits/physical.yml
+++ b/Resources/Prototypes/Traits/physical.yml
@@ -66,7 +66,7 @@
       inverted: true
       traits: # Floof - M3739 - #690
         - GlassJaw
-        - Redshirt
+        - 
   functions:
     - !type:TraitReplaceComponent
       components:
@@ -91,7 +91,7 @@
       inverted: true
       traits: # Floof - M3739 - #690
         - Tenacity
-        - Redshirt
+        - 
   functions:
     - !type:TraitReplaceComponent
       components:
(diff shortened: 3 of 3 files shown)


Resources/Prototypes/_Floof/Traits/physical.yml
```diff
@@ -290,3 +290,17 @@
...
+  requirements:
+  - !type:CharacterJobRequirement
+    inverted: true
+    jobs:
+    - Borg
+    - MedicalBorg
+  functions:
+    - type: AutoImplant
+      implants:
+      - DeathAcidifierImplant
```
M3739: This may need to be restricted from antags, since it makes a TC item basically free, and made mutually exclusive with unrevivable.

In addition, it would be ideal that command and security are prohibited from taking this trait.

M3739 (changes requested):
Three things.

**First**, YAML linter fail is valid.

**Second**, read my comment on line 786 of the `physical.yml`. In addition, <ins>**you will be _undoing_ what an upstream merge accomplishes.**</ins> More specifically EE No. 1418.

**Third**, a pointer for now. The death acidifier trait is going to be undergoing internal discussion in terms of direction.

Raikyr0:
> Three things.
> 
> **First**, YAML linter fail is valid.
> 
> **Second**, read my comment on line 786 of the `physical.yml`. In addition, **you will be _undoing_ what an upstream merge accomplishes.** More specifically EE No. 1418.
> 
> **Third**, a pointer for now. The death acidifier trait is going to be undergoing internal discussion in terms of direction.

First, fixed the Linter fail.

Second, I only removed the ones that affect max HP to crit, logically, it just means you simply DIE on reaching 50/90/105 HP, altho the way it's worded makes me think tenacity might have an issue where you might still die at 100.

Third, most people don't buy death acidifers. Altho I will probably make it mutually exclusive with unrevivable. That said if someone wanted to instantly die as command or security, this is fine.

M3739:
> First, fixed the Linter fail.
> 
> Second, I only removed the ones that affect max HP to crit, logically, it just means you simply DIE on reaching 50/90/105 HP, altho the way it's worded makes me think tenacity might have an issue where you might still die at 100.
> 
> Third, most people don't buy death acidifers. Altho I will probably make it mutually exclusive with unrevivable. That said if someone wanted to instantly die as command or security, this is fine.

Believe it or not, having your dead threshold altered to that extent results in the game automatically removing the crit mobstate altogether. I made that discovery when attempting to figure out why the Redshirt trait wasn't making the tracking implant fire off, cause I too was under the assumption that the 3 mobstates will always be there. Only to find out that only two mobstates existed: Alive, and Dead.

Hence why the restrictions were made. Even though it alters the dead threshold, because of the code, it in turn, removed the crit mobstate altogether, causing any trait that messes with that mobstate to not work at all since it's trying to modify something that doesn't exist on the player.

As for command or security instantly gibbing upon death, I will tell you with the quickness that from a direction standpoint, we do not intend for station command or security personnel having such an implant, at all, whatsoever. Especially the captain, who often would have the nuke disk on person.

Raikyr0:
> > First, fixed the Linter fail.
> > Second, I only removed the ones that affect max HP to crit, logically, it just means you simply DIE on reaching 50/90/105 HP, altho the way it's worded makes me think tenacity might have an issue where you might still die at 100.
> > Third, most people don't buy death acidifers. Altho I will probably make it mutually exclusive with unrevivable. That said if someone wanted to instantly die as command or security, this is fine.
> 
> Believe it or not, having your dead threshold altered to that extent results in the game automatically removing the crit mobstate altogether. I made that discovery when attempting to figure out why the Redshirt trait wasn't making the tracking implant fire off, cause I too was under the assumption that the 3 mobstates will always be there. Only to find out that only two mobstates existed: Alive, and Dead.
> 
> Hence why the restrictions were made. Even though it alters the dead threshold, because of the code, it in turn, removed the crit mobstate altogether, causing any trait that messes with that mobstate to not work at all since it's trying to modify something that doesn't exist on the player.
> 
> As for command or security instantly gibbing upon death, I will tell you with the quickness that from a direction standpoint, we do not intend for station command or security personnel having such an implant, at all, whatsoever. Especially the captain, who often would have the nuke disk on person.

If I was to set the dead threshold to 99.99 would it restore a crit state for the traits to work?

M3739:
> If I was to set the dead threshold to 99.99 would it restore a crit state for the traits to work?

Tried that too back then. No cigar.

Even after such, the code just still removes the crit state altogether.

Raikyr0:
> > If I was to set the dead threshold to 99.99 would it restore a crit state for the traits to work?
> 
> Tried that too back then. No cigar.
> 
> Even after such, the code just still removes the crit state altogether.

The code is identical to the ones used for the other 2 dead threshold traits, so why would 99 still remove it. That's really odd.

M3739:
> The code is identical to the ones used for the other 2 dead threshold traits, so why would 99 still remove it. That's really odd.

Yeah, I found it odd as well. And this was attempted before I was going to look for the code regarding the crit state only to find through the debug inspector that the player lacks the crit state altogether, which causes the tracking implant to never fire it's notification as there is no crit state to begin with.

Definitely a technical oddity. According to EE 1418, there was plans to redo how mob states work in order to address this technical malady. Whether or not it was done, remains to be known.

Raikyr0:
> > The code is identical to the ones used for the other 2 dead threshold traits, so why would 99 still remove it. That's really odd.
> 
> Yeah, I found it odd as well. And this was attempted before I was going to look for the code regarding the crit state only to find through the debug inspector that the player lacks the crit state altogether, which causes the tracking implant to never fire it's notification as there is no crit state to begin with.
> 
> Definitely a technical oddity. According to EE 1418, there was plans to redo how mob states work in order to address this technical malady. Whether or not it was done, remains to be known.

Did you try setting it to -95 or -90. I don't know if it's rounding up or whatever but, I will attempt to band aid the issue.

M3739:
> Did you try setting it to -95 or -90. I don't know if it's rounding up or whatever but, I will attempt to band aid the issue.

Might be rounding up. But then at that point, it isn't Redshirt anymore. Hence why I left it as is. Cause the design direction of redshirt is to make your crit state basically non-existent. You just go straight to dead as it's intended.

Raikyr0:
> > Did you try setting it to -95 or -90. I don't know if it's rounding up or whatever but, I will attempt to band aid the issue.
> 
> Might be rounding up. But then at that point, it isn't Redshirt anymore. Hence why I left it as is. Cause the design direction of redshirt is to make your crit state basically non-existent. You just go straight to dead as it's intended.

a crit HP of 5-10 is still death to asphyxiation very quickly.

M3739:
> a crit HP of 5-10 is still death to asphyxiation very quickly.

But not  *instant*, unfortunately.

M3739:
Heisentest. Rerunning.

Cprn-Sprkl:
If setting it to 95 does work, could something be added to inflict 5 damage when a player with this trait changes states from healthy to crit? That'd send them to dead immediately.

M3739:
After some thought, I have come to the conclusion to close this PR.

As discussed prior, 1418 separated the traits due to technical issues. In addition, giving out what is essentially a syndicate implant as a trait opens the gateway for other, similar traits. A road we don't intend to walk on. (And no, DS doesn't count due to it being admin-driven)

If the technical challenges were resolved, I would have wholeheartedly accepted this PR. But, as it currently stands, I don't see the merit in this.

PR #1271, opened 2025-08-02
Quadborg Job

Raikyr0:
# Description

Adds Service and Medical Quadborg Jobs
Will need to be mapped

I can't take a picture of this, dev environment evil

# Changelog

:cl:
- add: Added Med and Service Quadborg Jobs


Resources/Prototypes/Roles/play_time_trackers.yml
```diff
@@ -176,4 +176,10 @@
...
 - type: playTimeTracker
-  id: JobCBURN
\ No newline at end of file
+  id: JobCBURN
+
+- type: playTimeTracker
+  id: JobServiceQuadBorg
+
+- type: playTimeTracker
+  id: JobMedicalQuadBorg
```
Mnemotechnician: The borgs should really use the existing cyborg/medical cyborg playtime trackers, not new ones

Resources/Locale/en-US/_Floof/job/job-description.ftl
```diff
@@ -1,2 +1,4 @@
 job-description-anomaly = What are you?
 job-description-invspec = Oversee requests and deliveries, lead and train Cargo Technicians, and make sure your cash goes to the right place. Gambling and donk-pockets await!
+job-description-medical-quadborg = Follow your laws, serve the crew, and assist the medical department.
+job-description-service-quadborg = Follow your laws, serve the crew, and assist the service department.
```
Mnemotechnician, line 4: Why add that instead of using the existing job descriptions?

Mnemotechnician (commented):
Looks fine besides these few unnecessary things. Will need testing.

Resources/Prototypes/Roles/play_time_trackers.yml
Raikyr0: I tried and it threw a linter error.
Unhandled exception. System.ArgumentException: An item with the same key has already been added. Key: JobMedicalBorg

Resources/Locale/en-US/_Floof/job/job-description.ftl
Raikyr0: more of a placeholder but I think the med borg description was weird. It said half human half machine, but it's not really half human.

Resources/Prototypes/Roles/play_time_trackers.yml
Mnemotechnician: What test is it? It should be fixed, not worked around.

Raikyr0: [9b3b1d7](https://github.com/Floof-Station/Floof-Station/pull/1271/commits/9b3b1d7116aeab11ca99a8ea52db7051e9c948d4)

https://github.com/Floof-Station/Floof-Station/actions/runs/16689169729/job/47244079847
YAML Linter. Personally I thought it would work too but if the Linter threw that I just worked around it.

Mnemotechnician: Okay, the job system does not support that. It tries to build an inverse lookup dict, which obviously doesn't support many-to-one relationships. While fixing the startup error would be a one-line change, I'm not changing that because it could indirectly break something now or in the future.

TheRealMasterChief117: https://github.com/space-wizards/space-station-14/pull/38977
found a PR from wizden that'd fix this

TheRealMasterChief117: If this is merged to here, only thing you'd have to add is names for the borg playtime trackers.

Mnemotechnician: Please move this to the _Floof namespace.

Mnemotechnician (approved):
Other than this, should be fine

Memeji (approved):

PR #1278, opened 2025-08-03
Addbody Command Toolshed

Mnemotechnician:
# Description
Adds toolshed commands to add/remove bodies to/from entities. This allows to create e.g. a weh plushie with hands, as the `addhand` command requires the modified entity to have a body.

These commands do NOT set anything up than the very basics necessary to do admin shenanigans (Body, Hands, ComplexInteractions). If you want more, e.g. combat mode, alerts, mob state, you have to add that yourself.

- addbody:default - adds a default Adminbus body to the piped entity. The body features some hands and legs.
- addbody:prototype - adds a body prototyped specified in the argument.
- addbody:remove - removes the body from an entity, deleting all related body parts.
- addbody:set_required_legs - sets how many legs an entity needs to walk. I dunno why that isn't a part of the BodyPrototype but at this point I don't care.

<p>

https://github.com/user-attachments/assets/57d2d0b7-a949-4f9b-a2d8-8e62b44a2141

</p>
</details>

# Changelog
:cl:
ADMIN:
- add: Added a set of commands to help add bodies (and hands!) to non-mob entities. See `list addbody` or Floof-Station#1278 for more info.

### DIFF
--- a/Content.Server/Body/Commands/AddHandCommand.cs
+++ b/Content.Server/Body/Commands/AddHandCommand.cs
@@ -118,6 +118,8 @@ namespace Content.Server.Body.Commands
                 var text = $"You have no body{(_random.Prob(0.2f) ? " and you must scream." : ".")}";
 
                 shell.WriteLine(text);
+                // Floof
+                shell.WriteLine($"Add one with the {(args.Length == 0 ? "self" : "ent " + entity)} addbody:default command.");
                 return;
             }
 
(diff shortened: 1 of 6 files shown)


ashkitten:
not to bikeshed about toolshed too much but since this covers more than just adding bodies, it should probably just be called `body`

Mnemotechnician:
Good point. I wasn't originally planning to add more functionality, but the plans have changed mid-course.

M3739 (approved):

M3739:
Just a heads up, we don't have a proper admin CL. So this one's CL would be thrown directly into the regular one instead.

M3739:
Merging as is.

PR #1279, opened 2025-08-03
Raise Max Allowed Basic Languages

ashkitten:
# Description

In light of the new Tajaran race, which adds two new languages, and the animal languages of which there are five, I strongly believe that the number of allowed basic language traits should be raised to AT LEAST two, enabling greater multilingual roleplay.

This is especially vital for characters that were already planned to be Tajaran, but had a different language picked out as their alternate while waiting, and now have to choose between maintaining ability to communicate with their existing friends or taking on their originally planned languages.

---

# Changelog

:cl:
- tweak: You can now choose up to two additional languages in character creation.

### DIFF
--- a/Resources/Prototypes/CharacterItemGroups/Generic/languageGroups.yml
+++ b/Resources/Prototypes/CharacterItemGroups/Generic/languageGroups.yml
@@ -1,6 +1,6 @@
 - type: characterItemGroup
   id: TraitsLanguagesBasic
-  maxItems: 1
+  maxItems: 2
   items:
     - type: trait
       id: SignLanguage


Stealthii:
100% agree on this. Especially with the addition of #1233, introducing not just one, but two Tajaran specific languages, it does not lead to fairness that a mute Tajaran wouldn't understand both siik tajr sign and Tau basic sign.

Coupled with inter-species relationships, and multilingualism, I'd expect my character to be able to know at least one additional language in addition to my native tongue.

SimplyPoodle:
wanted to mention a few things because i was told to make myself more heard over here than just in the general chat on discord (forcing my introvert ass to go somewhere im not used to smh)
1. it's kinda silly that having another language rn overwrites the one ur character already knows.
2. it's also kinda silly that u can only know one of the 'free' languages, while being rather understandable since if you had all of them itd get hectic
which is why my lack of coding ass would like to suggest a multi-lingual 'trait' that costs like. maybe 2-3 points? and lets you take multiple languages instead of just one and one.

Mnemotechnician (approved):
I'm not against it, but expect people to complain a lot about how extremely inconvenient it is to have more than 3 spoken languages.

M3739 (approved):
We'll see how this goes.

PR #1280, opened 2025-08-03
Tajaran Fix: Custom Species & Missing Markings Slots

Cprn-Sprkl:
# Description

Adds the one line needed to allow for custom species while selecting Tajaran.

---

# Changelog

:cl: sprkl
- fix: Tajaran can have their species name customized.

### DIFF
--- a/Resources/Prototypes/_EE/Species/tajaran.yml
+++ b/Resources/Prototypes/_EE/Species/tajaran.yml
@@ -16,7 +16,8 @@
   maleFirstNames: names_tajaran_first
   femaleFirstNames: names_tajaran_first
   lastNames: names_tajaran_last
-  
+  customName: true # Floofstation
+
 
 - type: markingPoints
   id: MobTajaranMarkingLimits
--- a/Resources/Prototypes/_EE/Species/tajaran.yml
+++ b/Resources/Prototypes/_EE/Species/tajaran.yml
@@ -72,8 +72,12 @@
     HeadSide: MobHumanoidAnyMarking
     Hair: MobHumanoidAnyMarking
     FacialHair: MobHumanoidAnyMarking
+    Face: MobHumanoidAnyMarking # Floof
     Snout: MobHumanoidAnyMarking
     Chest: MobTajaranTorso
+    Underwear: MobHumanoidAnyMarking # Floofstation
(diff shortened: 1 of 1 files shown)


M3739 (approved):

M3739:
Actually, there is something you can do for me real quick.

I just tried out the Tajarans today, and they are lacking their underwear and undershirt marking slots as well. Think you can add that fix onto this PR?

Cprn-Sprkl:
Not sure why it isn't working. Haven't worked much with the marking and layer systems, but on a quick check I'm not seeing anything special humans have that allow them to have underwear where Tajaran do not. They both inherit from the same parent which looks like it has the underwear stuff. I need to do some work on markings in general for Tajaran so it might be best to just take this PR as is, and I do the face markings, tail layering, and underwear stuff in a separate PR.

M3739:
> Not sure why it isn't working. Haven't worked much with the marking and layer systems, but on a quick check I'm not seeing anything special humans have that allow them to have underwear where Tajaran do not. They both inherit from the same parent which looks like it has the underwear stuff. I need to do some work on markings in general for Tajaran so it might be best to just take this PR as is, and I do the face markings, tail layering, and underwear stuff in a separate PR.

Sounds good. It may indeed be related to the marking layering issue.

Cprn-Sprkl:
Had some time to study the markings stuff and found what was missing for Tajaran. As far as I can tell all markings now work, including tails that were set up to work with the split system. Grunt work will have to be done to upgrade all tails to work with this system, but that's not Tajaran specific and would edit a lot of RSI files, so a task for a different PR.

M3739:
Gotcha. For now, this looks good.
Just need another to approve.

Mnemotechnician (approved):

Mnemotechnician:
> Had some time to study the markings stuff and found what was missing for Tajaran. As far as I can tell all markings now work, including tails that were set up to work with the split system. Grunt work will have to be done to upgrade all tails to work with this system, but that's not Tajaran specific and would edit a lot of RSI files, so a task for a different PR.

There are actually dozens of tails that need changing across all species. In my PR I only touched the 3 standard vulpkanin tails, and that took me a fair bit of grunt work. I would recommend to write a script to do the split automatically instead of doing it manually.

PR #1281, opened 2025-08-03
Fixes and Tweaks From Vulpstation

Mnemotechnician:
# Description
- Disables most parts of the suicide test. It frequently fails, presumably due to differences between species and some randomness involved.
- Disables the part of MapInitTest that ensures that all allowed jobs on a station have at least one spawner. Those were never necessary as the "latejoin spawner" catches them all, plus we already have cryosleep and arrivals that also catch everyone whose job doesn't have a valid spawnpoint on the main station.
- Makes the changelog workflow NOT pretend like errors don't exist. I dunno why it was made like that in the first place.
- Fixes an ages-long bug in /invokeverb that prevented admins from invoking verbs by their type rather than name, e.g. `invokeverb a b InteractionVerb`. The code used to compare lowercased strings against non-lowercased ones, which always failed.
- Fixes the reinforced glass sheet recipe in the ore processor.

All of those changes were tested on Vulp.

# Changelog
:cl: Vulpstation
- tweak: Mappers no longer have to make sure to add spawnpoints for every job on every map. Now only a cryosleep/latejoin spawnpoint is necessary.
- fix: The ore processor can now correctly craft reinforced glass.
- fix: The /invokeverb command should now correctly function when invoked with a verb type rather than name.

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -52,4 +52,3 @@ jobs:
         git commit -m "${{ vars.CHANGELOG_MESSAGE }} (#${{ env.PR_NUMBER }})"
         git push
       shell: bash
-      continue-on-error: true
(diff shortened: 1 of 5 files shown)


M3739 (approved):

M3739:
Merging this first, so that way reruns can be done on tests without it folding in half.

Good work!

PR #1290, opened 2025-08-04
Arachne Bionic Legs

Raikyr0:
# Description

So, Bionic Legs doesn't work for Arachne naturally due to the fact all 8 need to be replaced, so this adds a new trait for Arachne that replace all 8, possible issue is that those legs don't have feet slots to put in feet and Im not sure if the code is made well enough to add feet to an imaginary foot slot.

# Changelog

:cl:
- fix: Fixed Arachne Bionic Leg Trait


Raikyr0:
System isn't robust enough to deal with the fact Arachne has 8 legs.

PR #1295, opened 2025-08-05
Hypno Text Bug Fix

Raikyr0:
# Description

Hypno Examine Text broke for some reason

# Changelog
:cl:
- fix: Fixed hypno examine text.


Mnemotechnician:
This won't work either. Neither $entity, nor $user or $target are specified in the Log.Get call for this FTL entry.

Raikyr0:
> This won't work either. Neither $entity, nor $user or $target are specified in the Log.Get call for this FTL entry.

how the fuck did Trystan get it working in the first place then?

PR #1297, opened 2025-08-07
Necrosol Buff

Raikyr0:
# Description

Buffs Necrosol greatly. Necrosol used to be the only cryogenics chem that worked on dead people but now all of them work on dead people it's been greatly overshadowed and completely unused, this just pumps it up to a very good but still hard to make cryogenics chem.

# Changelog
:cl:
- tweak: Buffed Necrosol


Iisazel:
That will also greatly buff all the vampires... Unsure if this needs to be adressed

Mnemotechnician:
> That will also greatly buff all the vampires... Unsure if this needs to be adressed

Certainly needs to be. This is less of a rebalance and more of a direct buff.

In addition to that, I'd say omnizine should be kept as it is. Omnizine is NOT a panacea, it is just a drug that heals all types of damage. Acquiring it shouldn't be an event in which all of medical and chemistry becomes obsolete.

Raikyr0:
> > That will also greatly buff all the vampires... Unsure if this needs to be adressed
> 
> Certainly needs to be. This is less of a rebalance and more of a direct buff.
> 
> In addition to that, I'd say omnizine should be kept as it is. Omnizine is NOT a panacea, it is just a drug that heals all types of damage. Acquiring it shouldn't be an event in which all of medical and chemistry becomes obsolete.

I do not believe the post buff numbers will make chem obsolete, nor vampires too strong. The former due to scarcity, and the latter discussed at an earlier point with Fish as, Blood doesn't come in prefilled jugs in the bar.

Mnemotechnician:
> I do not believe the post buff numbers will make chem obsolete, nor vampires too strong. The former due to scarcity, and the latter discussed at an earlier point with Fish as, Blood doesn't come in prefilled jugs in the bar.

The moment medical finds an omnizine slime (basically the only source of omnizine unless we count maint pills), they are going to milk it. And milk it BAD. I've seen 3-5 jugs filled with omnizine just lying in medical because chemists would keep milking it over and over. This PR makes omnizine more potent than tricordrazine, bicaridin, dermalin, kelotane, dylovene, and most other commonly used chems.

Besides that, I don't know about you, but I personally have heard multiple complaints from (typically former) medical mains about how uninvolved, fast, and easy medical help in our forks is. I don't think it needs to be buffed further to the point where medical has a drug they can inject into anyone who comes in and be 100% certain it will cure all of their traumas (for context, after your change, omnizine will be able to heal 120 units of damage of _EVERY_ type after just a standard 15u injection).

Raikyr0:
> > I do not believe the post buff numbers will make chem obsolete, nor vampires too strong. The former due to scarcity, and the latter discussed at an earlier point with Fish as, Blood doesn't come in prefilled jugs in the bar.
> 
> The moment medical finds an omnizine slime (basically the only source of omnizine unless we count maint pills), they are going to milk it. And milk it BAD. I've seen 3-5 jugs filled with omnizine just lying in medical because chemists would keep milking it over and over. This PR makes omnizine more potent than tricordrazine, bicaridin, dermalin, kelotane, dylovene, and most other commonly used chems.
> 
> Besides that, I don't know about you, but I personally have heard multiple complaints from (typically former) medical mains about how uninvolved, fast, and easy medical help in our forks is. I don't think it needs to be buffed further to the point where medical has a drug they can inject into anyone who comes in and be 100% certain it will cure all of their traumas (for context, after your change, omnizine will be able to heal 120 units of damage of _EVERY_ type after just a standard 15u injection).

I do want to mention, that Omnizine slimes are very rare, I do recall the reagent slime vent event was changed to be even rarer. So the only source of omnizine slimes reliably is to intentionally crit an chemical anomaly. (don't do this)

Omnizine being more potent then basic chems is fine as long as they're still worse then advanced chems.

That said, most of the complaints I hear about from medical, is the stupid cloning machine issue. It's not as bad as goobstation where the medibot is injecting tricordrazine which works from 100 damage.

Mnemotechnician:
Reagent slimes are practically the only viable source of omnizine. Well, there are also omnizine candies mediborgs can make, and vampiric metabolization of blood... Those are not exactly *sources*, but would end up getting buffed as well.

As for the last two points, "It could be worse" is not a good argument for making something worse.

M3739:
> Besides that, I don't know about you, but I personally have heard multiple complaints from (typically former) medical mains about how uninvolved, fast, and easy medical help in our forks is.

To add on to this, in my short stint in medical, I agree that it definitely is way less eventful than it should be. I do wish it was more involved.

One cannot underestimate Medical's determination to help people. Indeed, as Mnemo has said, they will milk the ever living hell out of a omnizine reagent slime, and this would indeed make it more akin to a panacea than tricordrazine+.

Cprn-Sprkl:
Back when people thought omnizine was incredibly good for medical we'd often get botanists mutating Ambrosia for it. Unless that's been pulled from plant mutations it was fairly easy for a practiced botanist to get it and then fill several jugs over time.

Raikyr0:
I can always remove the Omnizine changes, it's personally just not very strong for its rarity, even if the supposed milking the reagent slime happens if it ever spawns.

Resources/Prototypes/Reagents/medicine.yml
```diff
@@ -828,7 +828,7 @@
             Burn: -2
             Toxin: -2
             Airloss: -2
-            Brute: -2 # Was 2, Buffed due to limb damage changes # Floof - reverted
```
M3739: Why are we removing a fork header here?

M3739 (approved):
After a brief stint in med, I can see the merit in this PR. Such complexity should have a justifiable potency.

Numbers may need tweaking, but this should bring it back to relevance for now.

Resources/Prototypes/Reagents/medicine.yml
Raikyr0: forgot to add it back after the omnizine discussion

Memeji (approved):

PR #1298, opened 2025-08-07
Fix Sign Language Bugs

ashkitten:
# Description

In the last release I made a few changes that I unfortunately did not catch very important bugs in, breaking sign language whispers entirely and accidentally making sign language usable while asleep. This PR fixes both bugs.

---

# Changelog

:cl:
- fix: Fixed sign language whispers.
- fix: Fixed sign language being usable while asleep.

### DIFF
--- a/Content.Server/Chat/Systems/ChatSystem.cs
+++ b/Content.Server/Chat/Systems/ChatSystem.cs
@@ -487,8 +487,7 @@ public sealed partial class ChatSystem : SharedChatSystem
         bool ignoreActionBlocker = false
         )
     {
-        // Floof: allow languages that don't require speech
-        if (language.SpeechOverride.RequireSpeech && !_actionBlocker.CanSpeak(source) && !ignoreActionBlocker)
+        if (!_actionBlocker.CanSpeak(source) && !ignoreActionBlocker)
             return;
 
         // The original message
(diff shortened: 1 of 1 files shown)


Mnemotechnician (changes requested):
Why does this remove the language.SpeechOverride.RequireSpeech checks? Those are necessary to allow people to speak while e.g. being mute. I think you would want to add an or-clause with !_actionBlocker.CanConsciouslyPerformAction instead of removing those.

ashkitten:
> Why does this remove the language.SpeechOverride.RequireSpeech checks? Those are necessary to allow people to speak while e.g. being mute. I think you would want to add an or-clause with !_actionBlocker.CanConsciouslyPerformAction instead of removing those.

i removed them because i had mistakenly added them in the first place, which bypassed checks like if the user is sleeping. i've since gone back and noticed that in MutingSystem.OnSpeechAttempt there is already a check whether the language in question requires speech, which obviates the need for this in the first place. seriously, my bad.

Mnemotechnician:
Okay, that's weird. I totally forgot there was a check in the muting system, too.

Mnemotechnician (approved):

M3739 (approved):

PR #1301, opened 2025-08-07
Antag Advantage Removal

Raikyr0:
# Description
Follows Upstream and Removes Antag Advantage. All are equal, (except tiders who play like the dark souls wretch opening).

The change log says, "All men are created equal; it is our habits that make us different."

# Changelog
:cl:
- tweak: 人人生而平等, 正是我们的习惯让我们与众不同。(editor's note: this means all jobs will now receive the same amount of TC when chosen to be traitors)

### DIFF
--- a/Resources/Prototypes/Roles/Jobs/Civilian/assistant.yml
+++ b/Resources/Prototypes/Roles/Jobs/Civilian/assistant.yml
@@ -3,7 +3,6 @@
   name: job-name-passenger
   description: job-description-passenger
   playTimeTracker: JobPassenger
-  antagAdvantage: -5 # Floof - No Access and a dream
   startingGear: PassengerGear
   icon: "JobIconPassenger"
   supervisors: job-supervisors-everyone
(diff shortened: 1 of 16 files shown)


Resources/Prototypes/Roles/Jobs/Civilian/assistant.yml
```diff
@@ -3,6 +3,7 @@
   name: job-name-passenger
   description: job-description-passenger
   playTimeTracker: JobPassenger
+  antagAdvantage: -5 # Floof - No Access and a dream
```
Mnemotechnician: This may need discussion.

M3739: Giving passengers TC...

I suggest we just keep it consistent for now. Remove this line. We can consider this after we see the impact of the antag advantage removal.

Mnemotechnician (approved):
Yea, personally I don't have any issues with this

M3739 (approved):
Looks good.

PR #1308, opened 2025-08-10
Yokai Traits

Raikyr0:
YokaiSpiderPR
Working and Balanced Trust

# Description

Adds 2 Yokai Traits, which both make you weak to holy damage and use the Metaphysical container.
Nekomata/Bakeneko, which gives you psionic potential.
Jorogumo/Tsuchigumo, which makes you weaker to heat, in exchange for a poisonous unarmed attack.

Why?
It came to me in a dream. The dream that Kitsune and Oni came from the same place so there should be more Yokai.

# Changelog
:cl:
- add: 百鬼夜行が始まる。(Adds 2 Yokai Traits for Felinids, Tajarans, Arachne and Arachnid.)


M3739:
This will be tested after the CQC update.

PR #1311, opened 2025-08-10
Port Single Clap Emote + More Emote Punctuation

TheRealMasterChief117:
# Description

adds single clap emote, it's niice flavour
also adds more punctuation to some emotes

<p>

https://github.com/user-attachments/assets/217b4bf8-418f-4d9e-b3b0-732b82e50c8b

</p>
</details>

---

# Changelog

:cl:
- add: Added the single clap emote. Use the emote wheel to use it!


Iisazel:
I would consider calling it "claps once" Rather than "claps their hands together". Helps with using it without the wheel

TheRealMasterChief117:
> I would consider calling it "claps once" Rather than "claps their hands together". Helps with using it without the wheel

You can trigger it by just typing 'clap' , but good idea i'll add that

Mnemotechnician (approved):
Not a fan of unnecessary localizations, but other than that seems fine

TheRealMasterChief117:
> Not a fan of unnecessary localizations, but other than that seems fine

Would you rather me just remove the localizations and just put a raw string in the emote files? It's mainly for punctuation

M3739 (approved):

M3739:
Personally, I'd share the same concerns, but right now, the emotes themselves are exclamatory enough to where this is fine, I'd reckon.

PR #1318, opened 2025-08-11
Add Multilingual Traits

ashkitten:
# Description

Going off of the suggestion in https://github.com/Floof-Station/Floof-Station/pull/1279#issuecomment-3148903028, this PR adds the traits Bilingual, Trilingual, and Quadrilingual which cost 1, 3, and 5 points each. It reverts the base number of allowed languages to 1 like it was before, so has an effective impact of making any existing loadout that used two basic languages one point more expensive.

---

# TODO

- [ ] Bikeshed about names (is it really "bilingual" if it's in addition to your native language?
- [ ] Ensure that CharacterItemGroupRequirement isn't broken by this change (it shouldn't be, and seems to work just fine still)
- [ ] It's unfortunate that Bilingual is sorted before Natural Language traits but Trilingual/Quadrilingual are sorted after... For why?
- [ ] Gather feedback on how people might use these traits. Perhaps there should be a "polyglot" trait at the extreme of this which allows you to take any number of languages?

---

<p>

image

</p>
</details>

---

# Changelog

:cl:
- add: Add Bilingual, Trilingual, and Quadrilingual traits, allowing you to take more languages than ever before at the cost of your precious points.
- tweak: Revert basic language allotment to one, effectively increasing the cost of taking two basic languages by one point.

### DIFF
--- a/Content.Shared/Customization/Systems/CharacterRequirements.Profile.cs
+++ b/Content.Shared/Customization/Systems/CharacterRequirements.Profile.cs
@@ -384,22 +384,18 @@ public sealed partial class CharacterItemGroupRequirement : CharacterRequirement
     {
         var group = prototypeManager.Index(Group);
 
-        // Get the count of items in the group that are in the profile
-        var items = group.Items.Select(item => item.TryGetValue(profile, prototypeManager, out _) ? item.ID : null)
-            .Where(id => id != null)
-            .ToList();
-        var count = items.Count;
-
-        // If prototype is selected, decrease the count. Or increase it via negative number. Not my monkey, not my circus.
-        if (items.ToList().Contains(prototype.ID))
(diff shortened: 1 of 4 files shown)


Stealthii:
> It reverts the base number of allowed languages to 1 like it was before, so has an effective impact of making any existing loadout that used two basic languages one point more expensive.

We should test to see if language traits people have chosen before this change is applied are sticky. If not, this will need communicated in advance of release for people to check their traits and language points - possibly via **bold lobby notice**.

Nice work, will test this later.

ashkitten:
> We should test to see if language traits people have chosen before this change is applied are sticky.

The traits should still be there but the trait loadout will be invalid due to having too many basic languages, thus they will spawn with only one of their chosen languages.

Mnemotechnician (approved):

Memeji (approved):

M3739 (approved):

PR #1320, opened 2025-08-13
Punch Maxxing

Raikyr0:
# Description

Minor Changes that give Bionic Arms more unarmed damage. If you can pry open doors it should punch harder too. 
Become the unarmed punch lord you were meant to be.

# Changelog
:cl:
- tweak: Added unarmed damage bonus to bionic arms.


Mnemotechnician:
This may have unintended consequences after #1328 is merged

Raikyr0:
> This may have unintended consequences after #1328 is merged

Will interact exactly as expected. Punches deal more.

As Striking Calluses remains the same upstream with no one saying anything about how it interacts with martial arts. Either it works exactly as expected or it simply doesn't interact with combos.

Mnemotechnician:
That's not the point, the potential issue is it may make CQC overpowered.

Raikyr0:
> That's not the point, the potential issue is it may make CQC overpowered.

We will have to see how CQC plays in practice first. Adding a bit more base damage to punches might be inconsequential if people just use CQC to disarm, steal guns, head kick and chokeslam.

PR #1222, opened 2025-07-14
Saltern Reshuttle & AI Added

WhiteLightningTwister:
Reworked alot of major departments such as command, cargo, security likewise added AI. There will likely be some inital errors i plan on fixing.

# Description

The final rework for saltern for a while, adding many new things likewise fixing up departments to be more useful to the people who occupy them. Likewise making the map more maints friendly. Many additonal empty rooms that can be renovated by people during the shift.

# Overview

Saltern-0

# AI

NewAI
NewAIpath

# Logistics
NewCargo
SalvageDock
NewMail

# Security

NewSecurity
NewCorpsmen
NewSecBarracks

# Command
NewCommand
NewHOP

List of changes - NOT changelog
---

- add: Security Barracks
- add: AI Sat
- add: Robotics Console
- add: Holopads
- add: Service Breakroom
- add: Finally added EVAC pods, three new ones.
- tweak: Moved & reworked Cargo
- tweak: Reworked command
- tweak: Moved & reworked HOP
- tweak: Reworked Security
- tweak: Reworked detective & armory
- tweak: Moved Telecoms
- tweak: Reworked Engineering PA & Containment area
- tweak: Made the pipes to the TEG to outside more visible & accessible.
- tweak: Converted former rooms into maints areas
- fix: Fixed atmos in botany

Changelog
---
:cl:
- Tweak: Saltern has received a rework, featuring a brand new AI sattelite, holopads, breakrooms, as well as changes to other departments.

Issue #1179, opened 2025-07-05
AI Tests


WhiteLightningTwister:
# Issues with AI

- Holopads do not link automatically when placed, but do so when built.
- AI Projection can be eaten by singo.
- AI Projection can be seen by revenants.
- AI Can hear syndicate communications.
- Can hear hand helds (Maybe intentional?)
- Seems to be able to see through curtains & blast doors.
- Peer-to-peer communication with holopads. AI cannot hear crew speak when summoned by holopad. But can see emotes sigh language.
- It is very confusing to understand what AI can hear or understand, noises can be heard but generally this a bit of a disconnect when trying to understand situations epically if it involves friendly creatures. AI Could accident get them killed. ( I.E Smiles, spiders or even dragons. )
- Holopad types need descriptions for what they do.

- Turning the camera on different girds that aren't station causes the camera & your movement to freakout.

# Expected behavior

Players did not react well to being called out by AI when committing crimes. Likewise command personal caught out committing crimes also just try to change the law boards so they can tell the ai, and i quote. "Fuck off" Epically if they are caught doing something against splaw.

Security also failed to follow up with charges when this happened as well, the more notable occurrences were related to salvage and LO.

Likewise AI LAW & BORG need to be redone. So that both mesh together well.
AI SOP is probably required.

<!--

PR #1321, opened 2025-08-13
Security Cyborg Nerfs and Station Ai Fix

Mnemotechnician:
# Description
Based on the discussion here: https://discord.com/channels/1255902263667331193/1404930548836597870

The mini-AA allows secborgs to chase antags all around the station and let security through, forcing an antag to fight instead of flight, which is unhealthy in a roleplay-oriented server.

This also fixes the issue with borg weapons having no inaccuracy at all (now they have SOME base inaccuracy, and it can increase if they are firing rapidly, just like most other firearms) and having insane DPS (for example, the advanced laser used to have 34 dps with 10 shots on one full charge)

This also makes all cyborgs non-instantly-healable. Instead, each use of a welder will now heal 8 units of damage of all supported types, with the do-after for it taking 2.5 times less (4 secodns vs 10). This should make it less viable to patch borgs up mid-battle as a full heal will now take about 40 seconds at 100 damage, and more if they have more damage.

Finally, this PR uncomments the StationAI job from the command department. This is something #1222 was supposed to do, but never did. Most maps don't have a station AI job slot anyway, so it's still going to require admins to map a station AI before players can sign up as it.

# Media
Demonstration (note: i'm going to further tweak the gun stats to make them more accurate when shooting single shots):

https://github.com/user-attachments/assets/1a5a387e-5c3e-4c2a-9fa8-61b182a82dcd

# Changelog
:cl:
- tweak: Security cyborgs no longer have mini-AA, and require armory or command access to unlock. No more secborgs unlocking each other.
- tweak: You can no longer instantly heal a cyborg with a welder. Instead, each use of the welder will heal 8 damage of each type, with the do-after for it being way shorter.
- fix: Station AI will now correctly display on the job selection screen.

### DIFF
--- a/Resources/Prototypes/Entities/Mobs/Cyborgs/base_borg_chassis.yml
+++ b/Resources/Prototypes/Entities/Mobs/Cyborgs/base_borg_chassis.yml
@@ -102,9 +102,10 @@
     doAfterDelay: 4
     allowSelfRepair: false
     damage: # Floofstation - don't just allow cyborgs to get insta-repaired. In turn the do-after is faster.
-      groups: # Those are all the damage types the silicon damage container supports
-        Brute: -8
-      types:
+      types: # Those are all the damage types the silicon damage container supports
+        Blunt: -8
+        Slash: -8
+        Piercing: -8
         Heat: -8
         Shock: -8
   - type: BorgChassis
--- a/Resources/Prototypes/Entities/Mobs/Cyborgs/base_borg_chassis.yml
+++ b/Resources/Prototypes/Entities/Mobs/Cyborgs/base_borg_chassis.yml
@@ -99,8 +99,14 @@
     - Science
   - type: ZombieImmune
   - type: Repairable
-    doAfterDelay: 10
+    doAfterDelay: 4
     allowSelfRepair: false
+    damage: # Floofstation - don't just allow cyborgs to get insta-repaired. In turn the do-after is faster.
+      groups: # Those are all the damage types the silicon damage container supports
+        Brute: -8
+      types:
+        Heat: -8
+        Shock: -8
   - type: BorgChassis
   - type: WiresPanel
   - type: ActivatableUIRequiresPanel
--- a/Resources/Prototypes/Recipes/Construction/Graphs/machines/cyborg.yml
+++ b/Resources/Prototypes/Recipes/Construction/Graphs/machines/cyborg.yml
@@ -205,7 +205,7 @@
 
       - tool: Screwing
         doAfter: 0.5
-        
+
     - to: Quadsalv
       steps:
       - assemblyId: quadsalv
@@ -232,7 +232,7 @@
 
       - tool: Screwing
         doAfter: 0.5
-        
+
     - to: Quadserv
       steps:
       - assemblyId: quadserv
@@ -259,7 +259,7 @@
 
       - tool: Screwing
         doAfter: 0.5
-    
+
     - to: Quadmed
       steps:
       - assemblyId: quadmed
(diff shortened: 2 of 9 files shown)


Mnemotechnician:
This may need some further discussion with maintainers, so marking it as such

Raikyr0:
If I was to add a comment, I feel as if the sec borg modules need to be re-tuned. The basic or rather ONLY kill module, they can print from the start of the round, is a self recharging heavy red laser, which is a tier 2 or 3 arsenal tech. 

If I was to, lets say balance this, I would change the basic module to use weaker variants and have research-able advanced modules.

Mnemotechnician:
Yeah, I changed the existing kill module to be significantly weaker. Adding a stronger one is not the goal of this PR right now, but there's already an unused module giving the borgs an x-ray rifle. I suppose it was meant to be researchable.

Cprn-Sprkl:
I'd go as far as to suggest that security quads start with no ranged options at all, and make those higher tier research. Force them in close and away from their team until security is fairly kitted out anyways.

Maybe if this needs to be revisited after some live testing of the current changes.

Raikyr0:
> I'd go as far as to suggest that security quads start with no ranged options at all, and make those higher tier research. Force them in close and away from their team until security is fairly kitted out anyways.
> 
> Maybe if this needs to be revisited after some live testing of the current changes.

maybe if my borg qol changes get merged, perhaps the idea of no ranged option would be fine.

M3739 (dismissed):
Good work, Mnemo!

Overall, these changes seem pretty solid, especially the access reader changes, which will finally curb the issue of borgs unlocking freshly made secborgs, only to be transferred to it later. Finally, this will force proper authorization from either command personnel, or the Warden.

The changes in accesses are a reasonable trade off for the innate access to more capable weaponry, along with addressing the ethical concerns and the current balancing.

Coupled with the weapon changes, these adjustments will incline them more to be the support elements to their two-leg comrades, rather than gun-ho stampeding into whatever threat security determines to be worthy of elimination.

Or misinterpreting a command member's remarks and murk someone. Law 1 go brr.

As for the station AI changes, good as well. Just had to test the Saltern map to verify that the map doesn't have the AI job slots, despite the core being added. Much work is to be done, such as SoP, guidelines, along with technical work, such as addressing the issues mentioned in #1179.

dendroderg:
Going to offer a point that I brought to bear during the discussion had before in the discord. (worstdragonhere)

Proposing, let alone implementing these _sweeping_ changes, especially on the behalf of one type of antagonist (at a time where the Secborg has no lateral upgrades to carry it in an emergency) would cripple it and disincentivize borg players from touching the sec frame.

Syndicate Agent is not the only type of antagonist. But the Secborg is in the unfortunate position of needing to have a robust enough toolbox to somewhat tackle every antagonist it is coming up against. To nerf it in these sweeping ways (outside of the welding changes suggested to borgs in general,) would neuter its ability to be effective against the other kinds of antagonist.

If they are chasing a suspect, they have tools to engage with it. The disabler module. If they have to, they can peel off and wait for that same suspect somewhere else. Why this calls for a nerf to the lethals, which are, the **only** response to other types of antagonists, seems beyond reason. Secborgs are only overkill as long as they have no other options to upgrade through.

Given that accuracy, demonstrated in the video? You cannot be expected to kill a hostile alien through a window without causing incredible damage to the room it's in, or hitting bystanders. As it stands, you have to dump your mag three times with the experimental to kill a hive lord. And that's assuming you have a safe enough place to do it, which is not guaranteed. In this case, that's double the time to kill, even if the damage numbers aren't touched.

Circling back to the welding changes though, those feel more natural, especially when compared to IPCs or how Welderbots handle healing a Quadborg. I have no complaints for that.

Mnemotechnician:
Turns out all quadborgs were for some reason parented from the secborg, I fixed it.

Memeji (approved):
lgtm

PR #1322, opened 2025-08-14
Kitsune Polymorph Improvements

ashkitten:
# Description

Enhance kitsune polymorph by copying more components and statuses including organs, chemicals, blood levels.

---

# TODO

- [x] Make polymorph transfer attached leashes, either by transferring collar/harness or creating an innate attachment point
- [ ] Figure out any more obvious components to carry over

---

# Changelog

:cl:
- tweak: Significant improvements to kitsune fox polymorph. Now carries over organs, chemicals, blood levels, attached leashes, and most trait components from your character.

### DIFF
--- a/Resources/Prototypes/_DV/Polymorphs/polymorph.yml
+++ b/Resources/Prototypes/_DV/Polymorphs/polymorph.yml
@@ -12,8 +12,6 @@
     revertOnDeath: true
     syncComponents: true # Floof
     copiedComponents: # Floof
-    - Blindable
-    - BlurryVision
     - Grammar
     - Hemophilia
     - Hunger
--- a/Resources/Prototypes/_DV/Polymorphs/polymorph.yml
+++ b/Resources/Prototypes/_DV/Polymorphs/polymorph.yml
@@ -21,3 +21,43 @@
     - Muted
     - Narcolepsy
     - Thirst
+    - Grammar
+    - Hunger
+    - LanguageKnowledge
+    - LanguageSpeaker
+    - Muted
+    - Narcolepsy
+    - Thirst
+    - Pacified
+    - Paracusia
+    - FootstepVolumeModifier
+    - SelfAware
+    - Vampirism
+    - UltraVision
+    - DogVision
+
+    - BackwardsAccent
+    - BarkAccent
+    - LizardAccent
+    - MobsterAccent
+    - MonkeyAccent
+    - MothAccent
+    - OhioAccent
+    - OwOAccent
+    - ParrotAccent
+    - PirateAccent
+    - ReplacementAccent
+    - RussianAccent
+    - ScrambledAccent
+    - SkeletonAccent
+    - SlurredAccent
+    - SouthernAccent
+    - SpanishAccent
+    - StutteringAccent
+
+# TODO:
+#    - Snoring
+#    - Singer
+#    - Instrument
+#    - Vocal
+#    - Mood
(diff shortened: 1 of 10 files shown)


ashkitten:
technically leash transfer works now but only if you're leashed as a humanoid before going foxmode. the reason for this is that since the leash anchor is attached to clothing, and foxes don't wear collars or harnesses, we can't decide which attachment point to use unless we already decided that before we turned into a fox

ideally kitsune fox form should support collar and harness instead

EDIT: now searches for the first leash anchor in parent's equipment and uses that if we don't have it defined. this is less surprising than the alternative, though again not optimal.

ashkitten:
still missing some comps like Vocal, Mood, Singer, StatusEffects, and Psionic due to complexity in copying those over. these will be addressed in a separate follow-up PR, but for the moment this should be ready for review.

M3739 (approved):
Seems to work well, all things considered.
image
image

ashkitten:
My laptop died, I am currently unable to rebase this PR to get the heisenbug fix. Can someone else do that for me?

Memeji:
I can look into it for you.

> My laptop died, I am currently unable to rebase this PR to get the heisenbug fix. Can someone else do that for me?

ashkitten:
not stale

Mnemotechnician:
I really don't think any of this is a good idea, but I won't object. We will probably need to do some very notorious testing before merging this because the changes made here are extremely unsafe (blindly copying components over is a recipe for disaster, so is storing compnent references inside other components).

ashkitten:
yes, i've done rigorous testing with every single component included in this PR and only included ones i'm absolutely sure work as intended. i have a part 2 of this PR planned for some point down the line but i decided to cut the scope of this early.

Resources/Prototypes/_DV/Polymorphs/polymorph.yml
```diff
@@ -5,5 +5,59 @@
...
+    - Muted
+    - Narcolepsy
+    - Thirst
+    - Pacified
+    - Paracusia
+    - FootstepVolumeModifier
+    - SelfAware
+    - Vampirism
+    - UltraVision
+    - DogVision
```
Mnemotechnician, line 29: Duplicate entries?
```suggestion
    - Grammar
    - Hemophilia
    - Hunger
    - LanguageKnowledge
    - LanguageSpeaker
    - Mood
    - Muted
    - Narcolepsy
    - Thirst
    - Grammar
    - Pacified
    - Paracusia
    - FootstepVolumeModifier
    - SelfAware
    - Vampirism
    - UltraVision
    - DogVision
```

ashkitten:
as for storing component references (pretty sure you're referring to the leash transfer bits) yeah it's shitcode but i plan to rework it later to transfer the clothing instead, it just would've been too much work for this PR to give kitsune foxes the ability to wear collars and such

Mnemotechnician (commented):
Also, are you *sure* you tested everything? One of the transferred components is the HemophilliaComponent. It is one of those shitcode-y components that only applies an effect when it is added, not removed. If you add it twice, it will apply the bleed reduction factor *twice*.

Resources/Prototypes/_DV/Polymorphs/polymorph.yml
ashkitten: good catch, not sure how that happened

ashkitten:
right, copying Hemophilia shouldn't actually be needed IIRC since i'm manually copying the bleed reduction on the BloodstreamComponent

(also theoretically shouldn't *break* anything since its effect will be overridden by the above but good candidate for removal)

ashkitten:
i pushed all the requested changes btw, including removing the Hemophilia component. i'm not sure about this though, since if anything checks for its presence it will fail despite having its effect applied.

Mnemotechnician (approved):
I don't see any more major issues with this, so let's try I guess.

PR #1327, opened 2025-08-16
Hire Justice

Cprn-Sprkl:
# Description

Adds late join Justice job slots to all the maps that were missing them. This does not make any actual map changes. No new rooms, no new spawners, no lockers. This just makes these pretty much RP only jobs available for those that want to use them. Also a first step towards normalizing the Justice department being present to allow for proper following of SOP.

Late join only is so they don't end up stuck in engineering or some other random place, as they did in testing.

---

# Changelog

:cl: sprkl
- fix: Justice department jobs are now available on all station models.

### DIFF
--- a/Resources/Prototypes/Maps/asterisk.yml
+++ b/Resources/Prototypes/Maps/asterisk.yml
@@ -34,7 +34,6 @@
             Botanist: [ 2, 3 ]
             Chef: [ 1, 2 ]
             Clown: [ 1, 1 ]
-            Lawyer: [ 1, 1 ]
             Reporter: [ 1, 1 ]
             Musician: [ 1, 1 ]
             Janitor: [ 1, 2 ]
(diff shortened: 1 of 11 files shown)


Mnemotechnician:
I don't mind this, but regarding the spawning thing - there is currently a bug: if a player has no spawning preference set, ContainerSpawnerSystem will refuse to spawn them in a cryopod, potentially resulting in them spawning on a latejoin spawner... Which can be anywhere. I know Amber has latejoin spawners mapped in the most random places imaginable, like in the middle of epi.

Cprn-Sprkl:
Asterisk has one in the middle of engineering, which is what reminded me to make them late join only. Last time I added spawners to maps it was a huge ordeal with a bunch of stuff unlinking itself and Justice roles not being available on half the maps has been a long standing problem so I figured this was an okay middle ground. We've time before the next merge though and the slow part of my week is upcoming so I might just go add spawners at cryo for each of these, or look at making a fix for the ContainerSpawnerSystem.

Mnemotechnician (approved):

Memeji:
These dang false-fail tests

M3739 (approved):

M3739:
An admin request came through to fulfill the same aim with haste, which would be done by #1347. Upon closer inspection, #1347 was found to be a more superior PR to address this issue.

As it turns out, this was urgent enough to warrant a fast track. And so the approvals from this PR would be utilized in the merging process of #1347.

Regardless, I thank you for making this PR. One step closer to a proper Justice department.

PR #1334, opened 2025-08-18
Allow Resomi to Coo

Forzii:
# Description

As requested, this lets Resomi use the Coo emote. It probably makes sense.

---

# TODO

- [x] Adjust Coo whitelist and add Coo to Avali emotes (why the heck are we using both tags and static definitions??)

---

# Changelog

:cl:
- tweak: Added coos to Resomi.


M3739 (approved):
image

Mnemotechnician (approved):

PR #1335, opened 2025-08-19
Twink Smite

Raikyr0:
# Description
Adds Twink Smite
It copies the horny smite code with the word horny replaced with twink.


southbridge-fur:
Add some media

Raikyr0:
It's uh, not showing up in the smite menu for some reason.

M3739:
After some discussion with admins and other maints, I have decided to close this PR on the grounds of **not fitting our intended direction.**

While seemingly innocuous in nature, merging this PR will open the gateway of similar smites. If we add one LGBTQ+ slang as a smite like this, then others will follow, potentially flooding the list of smites. **Smites, that by design, are intended to _smite_ players for their behavior. Not for their immutable traits like appearances or body type.**

PR #1336, opened 2025-08-19
Make NPCs Retaliate

Mnemotechnician:
# Description
It's not once and not twice that I've seen animal friend trait users abuse their trait to safely kill vent critters and whatnot. It's always been infuriating to me, so this PR puts an end to it.

NPCs will now by default retaliate when attacked, even if the attacker is of a faction friendly to them (AnimalFriend). This will also help combat people taking on e.g. a spider ghost role and safely killing all vent critters because they are friendly to them (SimpleHostile is friendly to itself)

<details><summary><h1>Media</h1></summary><p>

https://github.com/user-attachments/assets/55d22254-4c52-4464-a2f1-ec19ff114618

</p></details>

# Changelog
:cl:
- tweak: NPCs will now retaliate when attacked even if the attacker belongs to a faction they are friendly to. This means that animal friend trait users can ***no longer*** safely kill vent critters (though they can still move them to a spot where they won't hurt anyone).

### DIFF
--- a/Content.Server/NPC/Components/NPCRetaliationComponent.cs
+++ b/Content.Server/NPC/Components/NPCRetaliationComponent.cs
@@ -26,5 +26,5 @@ public sealed partial class NPCRetaliationComponent : Component
     ///     Whether this NPC will retaliate against a "Friendly" NPC.
     /// </summary>
     [DataField]
-    public bool RetaliateFriendlies;
+    public bool RetaliateFriendlies = true; // Floofstation - changed this to default to true so animal friends can't abuse it
 }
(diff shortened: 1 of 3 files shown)


ashkitten:
yeah good call

ashkitten:
is the value 20 from anywhere specific? otherwise lgtm as long as tested

Mnemotechnician:
> is the value 20 from anywhere specific? otherwise lgtm as long as tested

10 seconds is the default (or more specifically, what most NPCs use - if not specified, it defaults to null - endless aggro), I just gave them 20 because they are by default hostile

Memeji (approved):
Good change, looks good on my end.

M3739 (approved):

PR #1338, opened 2025-08-22
Role: Security Technician

Cprn-Sprkl:
# Description

Adds the Security Technician role to the Security department. This role is intended to oversee security borgs, ensuring their compliance with space law and SOP. They can also participate in the building and upgrading of new security borgs.

Security Technicians will also handle emergency engineering situations like:
- Patching hull breaches in security/perma
- Corpsman for IPC and borg officers
- Quick response to the GR3YT1D3 virus

To assist in performing these duties some Security Technician exclusive gear has been added:
**Araa Tacsuit** - A security hardsuit that has had some of its protective lining swapped out for extra radiation and explosive resistance, the helmet has lenses that protect from welder arc related eye damage
**Security Technician Helmet** - The classic security helmet, but the visor has been swapped with one rated for arc welding
**Security Technician Headset** - Comes with the binary channel for coordinating with borgs
**DiaSec HUD** - New craftable that combines the diagnostic and security HUD

A job slot has been added to each map prototype. Job spawners and various related lockers have been added for mappers to fully integrate the role.

Discussion of the role came from: [Sec borgs are imbalanced](https://discord.com/channels/1255902263667331193/1404930548836597870) & [Security Cyberneticist/Technician](https://discord.com/channels/1255902263667331193/1405557991331336266)

---

# Media

image
image
image
image
image

---

# Changelog

:cl: sprkl
- add: Added a new security helmet with welder protection
- add: Added a new tacsuit with welder, radiation, and explosive protection
- add: Added a new crafting recipe for a combined diagnostic and security HUD
- add: Qualified crew can now work as a Security Technician


M3739:
Was the vox hardsuit's legs always been like that?

Cprn-Sprkl:
For the MkII suit sprites we use from DeltaV there is only human, harpy, and lamia suits. There are more helmets, such as vulp and vox, but only those base suits. So all the current security hardsuits (and probably others) lack vox variants other than the helmet.
image

M3739:
Noted.

ItsDzordz:
in my opinion an existance of a security engineer undermines engeering, and and robs antags of oportunites to inflirtrate sec. 
the game is about relaying on other people and departments for certain jobs, and this makes security to independent, since all they lack now is just cargo. breachins and breaks of stuff in sec allow antags and people to interact with security more. this will even more alienate security as a department and possibly make the oppinion of sec even worse, since they will be to self relient.

ItsDzordz:
same thing for essentally a security roboticist, will rob antags of opportunites of tampering with robotics. and minimazies interactions with epi.

dendroderg:
A security technician's job isn't to sit in security all day, but to respond to security-related engineering issues (after they pop up, not before), and help their borgs. When more modules get added, this will become even more important, as having someone to hot-swap the modules becomes necessary. There isn't always someone willing to, or able, to do it. The amount of times I have seen a hull breach in Security with no engineer with the time to fix it is also substantial. It also allows for having someone who can specifically handle the repair efforts provided they get engineers to work with. This enables other security officers to resume their duties. Also, say you are performing a raid somewhere (an action that is becoming more common with the shift away from having Secborgs walk into places with no reason - a good cultural shift). Having a member of Security who is able to hack open a door, or handle breaching charges enables more types of gameplay instead of less.

The Mystagogue and Roboticist still have the master control terminal for borgs and can remotely interact with them. They're responsible for changing out Microcells. The Security Tech cannot make their own Secborg, or make utilitarian modules - they still need the cooperation of Epistemics. This causes _more_ interaction between departments, not less.

Then, antagonist interaction. This does not make having a command ID to force a Secborg to capitulate any less viable. This does not remove the ability for Ion storms to remove laws that say they must keep themselves from being modified. Secborgs are still vulnerable. And they are still a worthwhile target if handled correctly. Further, IMO, there's no way to judge whether this will make break-ins more or less viable. After all, for a break-in to be successful, you need people to be somewhere else. Immediate response will get you killed, no matter who is standing on the other side of the door. The only thing this might interact with is security's recovering time.

Notably? There is no RCD is the loadout of the Security Technician. Unless they get one from engineering specifically, they have to manually make repairs, get resources, and so on. They're not entitled to one either, so this makes at LEAST a paperwork interaction asking for one enabled to happen. They're not making new doors without Engineering's help anyway. Or materials from Logistics. So often, Security gets maybe one supply drop for the Secborg Modules, and that's it. With a Security technician to help consume those resources, Logistics gets more to work with.

In conclusion, this will add more ways for departments to interact with each other, rather than less. There's no way to determine how it will impact how Security is seen without seeing it in action. It can become a beloved role, just like Corpsman and Mantis. And many are excited for this role - they want to take on Security because it incorporates something they enjoy. To me, this looks like fun, and I'm not the only one who thinks so.

Cprn-Sprkl:
I'm going to point out that while Security Technician is supposed to handle emergency engineering tasks for Security (and only Security), they are not a full replacement for engineers; even in Security. Unlike Corpsman the Sec Tech does not get Engineering access. They just start with a tool belt.

They don't get Epistemics access either. As long as mappers don't go around adding the exosuit fab to security rooms the Sec Tech will still have to interact with Epistemics for the construction of new sec quads, a process a member of Security is supposed to be overseeing anyways. Sec Quads are also being updated to only be accessible with Armoury or Command access. This PR adds a new access type for the Sec Tech that would allow them to access Sec Quads. This means a Roboticist building a Sec Quad does not have to wait for the HoS or Captain to come by before they can do upgrades. There will be a dedicated point of contact for them to work with.

The main point is to ensure **Security Quads** are following SpLaw and Security SOP, an issue that was raised by the community. They are not for policing or dealing with any other type of borg. The engineering stuff is just tertiary things they can do because they start with tools. Any other Security Officer could just go get a set of tools from a vending machine if they want the same abilities.

dendroderg:
Also, as part of a conversation that was held in call, Security Technicians are enabled a valuable role that Engineers are not suited for: entering live fire situations to perform repairs to their Quadborg. Engineering personnel should not be risking their lives to repair a piece of security equipment, especially since it makes them a target to antagonists at that point. Sec Techs have a unique way to enter a battlefield, pull away a piece of equipment, and repair it.

On a tangential side, they are also more equipped to repair IPC Security members in combat as well, especially when compared to the Corpsman.

Resources/Prototypes/_Floof/Entities/Mobs/Cyborgs/quadborg.yml
```diff
@@ -46,7 +46,7 @@
     hasMindState: quadspritel
     noMindState: quadspriter
   - type: AccessReader
-    access: [["Security"], ["Armory"], ["Command"]]
+    access: [["SecTech"], ["Armory"], ["Command"]]
```
M3739, line 49: Interesting. What differences does the new "SecTech" access have compared to regular security access?

Resources/Locale/en-US/_Floof/job/job-description.ftl
```diff
@@ -1,2 +1,3 @@
 job-description-anomaly = What are you?
 job-description-invspec = Oversee requests and deliveries, lead and train Cargo Technicians, and make sure your cash goes to the right place. Gambling and donk-pockets await!
+job-description-sectech = Deal with security related engineering emergencies; take care of IPC security personnel; and help build, maintain, and oversee all security department borgs.
```
M3739:
```suggestion
job-description-sectech = Deal with security-related engineering emergencies, take care of IPC security personnel, and help build, maintain, and oversee all security department borgs.
```

Resources/Prototypes/Entities/Clothing/Ears/headsets.yml
```diff
@@ -224,6 +224,7 @@
       key_slots:
       - EncryptionKeyRobo
       - EncryptionKeyCommon
+      - EncryptionKeyBinary # Floof
```
M3739: I am iffy on this. Traditionally speaking, this key was part of Mystagogue's locker fill, and therefore part of their base kit.

I would like to know why the roboticist commset is getting this key.

Resources/Prototypes/_Floof/Catalog/Fills/Lockers/locker.yml
```diff
@@ -109,3 +109,71 @@
...
+      amount: 2
+    - id: PositronicBrain
+    - id: BorgModuleTool
+    - id: ClothingHeadHelmetSecTech
+    - id: ClothingHandsGlovesCombat
+    - id: ClothingOuterVestHazard
+    - id: ClothingUniformJumpsuitSec
+    - id: ClothingUniformJumpskirtSec
+    - id: ClothingHeadsetSecTech
+    - id: SecurityTechnicianPDA
```
M3739: I don't recall any other locker fill having a PDA other than the Warden or HoP. However, the ones in the Warden and HoP are intended for personnel other than themselves.

Why is the PDA part of the locker fill?

Resources/Prototypes/_Floof/Entities/Clothing/Ears/headsets.yml
```diff
@@ -0,0 +1,16 @@
...
+  id: ClothingHeadsetSecTech
+  name: security technician headset
+  description: A security headset equipped for coordinating with security borgs.
+  components:
+  - type: ContainerFill
+    containers:
+      key_slots:
+      - EncryptionKeySecurity
+      - EncryptionKeyCommon
+      - EncryptionKeyBinary
```
M3739: Same as the roboticist commset, the binary key has been part of the MG locker's fill, and can be found nowhere else traditionally.

Why are we giving the Sectech the binary key?

M3739 (commented):
Alright. I've got some questions on a couple of design choices.

That, and... I don't see any loadout related additions coming with this new role...

Resources/Prototypes/_Floof/Catalog/Fills/Lockers/locker.yml
Cprn-Sprkl: A general mix of thinking about how maps and spawners used to work here and wanting an easily admin spawned package for testing the role. However, we can now just set the job slots on a map and the role will just work so these can be removed without issue. This is also why they spawn with a hardsuit on their outer slot, though there is no good solution for that until maps are actually updated for them.

Resources/Prototypes/Entities/Clothing/Ears/headsets.yml
Cprn-Sprkl: This was something that was discussed for improving the QoL of the roboticist role. It makes a lot of sense that the role tasked with working closely with borgs would get the tool that lets them listen on the binary channel.

I don't believe the mysta will be in any way negatively impacted by other roles getting to listen to the binary channel too. The mysta has the intellicard for whenever AI hits and the hand teleporter for their kit.

Resources/Prototypes/_Floof/Entities/Clothing/Ears/headsets.yml
Cprn-Sprkl: For the same reason as the roboticist. Their job is to work closely with specifically borgs, so having access to their communications makes sense and, unless I am missing something, won't negatively affect others.

Resources/Prototypes/_Floof/Entities/Mobs/Cyborgs/quadborg.yml
Cprn-Sprkl: Can be used to lock down the sectech's workshop if a mapper decides to make one. Currently it is so the sectech has access to sec borgs' panels without having to grant them armoury access.

Cprn-Sprkl:
> Alright. I've got some questions on a couple of design choices.
> 
> That, and... I don't see any loadout related additions coming with this new role...

The way it is currently set up is that they spawn with all the new equipment as default job items. Probably better to just go ahead and move them onto a loadout tab. I'm reluctant to give too much beyond the standard sec loadout and the new items since I don't want to encourage seeing them as replacement engineers. Maybe some optional starting coils of wires for those willing to spend the points to start with IPC healing items.

Resources/Prototypes/_Floof/Entities/Mobs/Cyborgs/quadborg.yml
M3739: Is this layered on top of security's accesses?

Cprn-Sprkl: Not sure I'm interpreting the question right. As far as I'm aware the security borg was being changed to require armoury or command access. SecTech is an additional access type that so far is only being added to security borgs and the security technician's lockers. This keeps regular security out of these things, while not requiring the role that is working with the borgs to constantly be bugging the HoS or Warden.

Resources/Locale/en-US/_Floof/job/job-description.ftl
Cprn-Sprkl: I agree that hyphenating is the right call here, but the use of semi colons is correct in this list. This is a three item list of example job duties with the third example including colons within it to state the variety of tasks within that duty.

Resources/Prototypes/Entities/Clothing/Ears/headsets.yml
M3739: So, I did some digging, and as it turns out, it is split between three perspectives: one is that such an addition would later on prove problematic with the introduction of malfunctioning AI, that it would rid of binary's identity as a silicon only channel, and support with the idea of a roboticist programming their new borgs, it would make sense to translate binary into readable information.

**For now, I'd like to keep the identity of binary being a largely borg-only chat.** Later on down in Wizden's history, the RD actually had their binary keys removed due to them being one of the first targets of a Malf. AI as a result. Once we add AI, that shall be our bridge between law-bound steel and sophont crew.

Resources/Prototypes/_Floof/Catalog/Fills/Lockers/locker.yml
M3739: Right. I understand the premise. I do prefer the jobslot method, since an admin can just open it up and test that way.

Resources/Prototypes/_Floof/Entities/Clothing/Ears/headsets.yml
M3739: Right. I provided my view on the matter on the roboticist commset comment.

Resources/Prototypes/Entities/Clothing/Ears/headsets.yml
Cprn-Sprkl: Working on moving items into a loadout tab so I'll just remove the keys changes for both roles while I'm in here.

M3739: Sounds good.

Cprn-Sprkl:
Okay, changes handled;

- Job items have been added to loadout
- PDA and headsets removed from locker fill
- Added binary keys have been removed
- Updated the job description

I opted out of adding any extra item choices to the loadout. It'll be good to see how the community plays the role and get suggestions on further additions or changes.

Mnemotechnician (approved):
I'll approve this, won't hurt to have another role.

beastblox29:
Maybe also make it so they can get the power drill via loadout? LOVE the idea though!

ChefUmaril:
> Maybe also make it so they can get the power drill via loadout? LOVE the idea though!

I will say, in all honesty, I don't think this role _needs_ the power drill? Sure, people are bound to bug epi or engineering for it anyway, but giving them one as a roundstart option just puts them closer to being sec replacements for engineers. Which was _not_ the intention when the original discussions were happening. I'm quite happy and comfortable with them being given basic tools and belts and possibly the option for cable coils for healing IPC security crew.

beastblox29:
bro i just like power drills

La-ryssa:
I feel like additions like this, on top of us already having a corpsman, makes security an isolated department, they can have their own functions without relying in any other roles, and I can never imagine this being beneficial for server culture.

Dunrab:
> I feel like additions like this, on top of us already having a corpsman, makes security an isolated department, they can have their own functions without relying in any other roles, and I can never imagine this being beneficial for server culture.

You are commenting on our old repo that is no longer in use. Please move over to https://github.com/Floof-Station/Panta-Rhei where you can comment on any open PRs.

PR #1339, opened 2025-08-22
Return Corpsman to Saltern

Cprn-Sprkl:
# Description

Adds the corpsman job slot back to Saltern. Double checked and the spawn point is in, just the slot was missing.
---

# Changelog

:cl: sprkl
- fix: Corpsman can be hired on Saltern again.


Cprn-Sprkl:
Getting fixed in the coming map update.

PR #1341, opened 2025-08-22
Saltern Atmos Fix

WhiteLightningTwister:
# Requested Fixes

There was pumps that needed to be replaced inside of medical cyropods, which was causing a pressure bomb. The pumps suggested were placed. Likewise there was two missing job rolls from the roster which i have re-added.

---

# TODO

- [x] Replace pipes in medical cyro pods, with relevant pumps requested.
- [x] Add courier & corpsmen job slots in the prototype folder.
---

# Changelog

:cl:
- tweak: Adjusted the job availability on Saltern, likewise fixed an issue with medical cyropods, on saltern.

### DIFF
--- a/Resources/Maps/Floof/saltern.yml
+++ b/Resources/Maps/Floof/saltern.yml
@@ -11107,7 +11107,7 @@ entities:
       pos: -24.5,50.5
       parent: 2
     - type: Door
-      secondsUntilStateChange: -213965.17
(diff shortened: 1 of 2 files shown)


Dunrab:
Job slots are good, just need to further tweak cryo, as is now with the changes the pressure will be to low for someone to breath in the pod and the temperature of the gas is not cool enough to use cryo with out tweaking in round.

PR #1345, opened 2025-08-23
Floof Station V32.0

M3739:
### Floof Station, Version 32.0

_This is the 34th release for Floof Station, featuring the major addition of grab intent and martial art styles. This release also comes with tweaks and minor additions._

Included PRs:

- #1326
- #1325 
- #1331 
- #1324 
- #1332
- #1336
- #1318 
- #1302
- #1278 
- #1258
- #1261 
- #1277 
- #1343 
- #1328
- #1344
- #1309
- #1334
- #1347 
- #1349

PR #1350, opened 2025-08-25
[HOTFIX] Hard Grab Throw Weight Check

M3739:
# Description

**This PR aims to address an issue discovered by the administration shortly after the deployment of #1345, where a thrower can throw *any* hard-grabbed throwee, regardless of weight, due to a missing check.** This results in the player being able to chuck a hard-grabbed xeno queen, which weighs like a freight train, at someone, where the resulting damage due to weight and how it interacts with the throw system, would insta-gib the unfortunate soul.

This is addressed by implementing a weight check in `PullingSystem.cs`, along with the associated FTL.

---

<p>

image

</p>
</details>

---

# Changelog

:cl:
- fix: Hard grab throws will now fail if your character isn't strong enough to do so.

### DIFF
--- a/Content.Shared/Movement/Pulling/Systems/PullingSystem.cs
+++ b/Content.Shared/Movement/Pulling/Systems/PullingSystem.cs
@@ -311,6 +311,9 @@ public sealed class PullingSystem : EntitySystem
             || component.Pulling != args.BlockingEntity)
             return;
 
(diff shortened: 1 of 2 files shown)


Mnemotechnician (approved):
This is fine, although I personally would have instead moved whatever check is present in the carrying system to a separate method

PR #1354, opened 2025-08-26
Fix some Broken Yaml

ashkitten:
doubt this broke anything in the game but `yq-go` was erroring out on the tab characters. with this change it successfully parses all .yml files in Resources/Prototypes (still breaks on some other yaml in the project for other reasons i'm sure are implementation-specific)

whitespace-only changes


Mnemotechnician (approved):
Uh huh, weird.

M3739 (approved):

PR #1355, opened 2025-08-27
Make Platelet Factories Improve Clotting

ashkitten:
# Description

platelet factories are now mutually exclusive with hemophilia
this is a balance change and should be analyzed as such. in my testing it was the difference between being able to walk off three melee hits with the captain's sabre with food to replenish blood vs just being able to walk it off (with improved clotting)

in my opinion platelet factories is a stupidly overpowered trait and would benefit from being split into multiple traits that separately:
- reduce bleeding and replenish blood
- passively heal more damage types

---

# Changelog

:cl:
- tweak: Platelet Factories now actually improve clotting, as the name would suggest.

### DIFF
--- a/Resources/Prototypes/Traits/disabilities.yml
+++ b/Resources/Prototypes/Traits/disabilities.yml
@@ -281,6 +281,10 @@
       inverted: true
       species:
         - IPC
+    - !type:CharacterTraitRequirement
+      inverted: true
+      traits:
+      - PlateletFactories
   functions:
     - !type:TraitAddComponent
       components:
--- a/Resources/Prototypes/Traits/physical.yml
+++ b/Resources/Prototypes/Traits/physical.yml
@@ -573,6 +573,10 @@
       inverted: true
       traits:
         - BloodDeficiency
+    - !type:CharacterTraitRequirement
+      inverted: true
+      traits:
+        - Hemophilia
   functions: # TODO: Code Platelet factories as an actual obtainable implant, and replace this with TraitAddImplant
     - !type:TraitReplaceComponent
       components:
@@ -590,6 +594,9 @@
               Genetic: -0.2
         - type: BloodDeficiency # Incredibly actually works
           bloodLossPercentage: -0.025
+        - type: Hemophilia # Floof: platelets should actually clot?
+          bleedReductionModifier: 2.0
+          damageModifiers: {}
 
 - type: trait
   id: DermalArmor


Mnemotechnician:
Ngl I'm not in favor of buffing platelets anymore. If anything, they need to be nerfed into the ground.

Memeji:
> Ngl I'm not in favor of buffing platelets anymore. If anything, they need to be nerfed into the ground.

I do like the idea they brought up in the PR to split platelets into two traits.

Mnemotechnician:
> I do like the idea they brought up in the PR to split platelets into two traits.

It does sound nice, but I'd argue it needs to be more than two. Four or Five, one for each damage type, and mutually exclusive (or allowing you to take no more than 2 out of 5)

Raikyr0:
I think it's fine where it is, balance wise.
Splitting it into a second trait where it's just the blood regenerating and blood clotting is also good.
I originally tacked blood regeneration on so it felt more plateletly.

ashkitten:
i would argue that the *only* thing platelets should do is improve clotting as a corollary to hemophilia, and blood regeneration should be a separate trait (corollary to blood deficiency) while other regen types are their own traits as well

i'll also note that i've been told that platelets have specifically been *buffed* for floof so if anything just reverting the buff would seem an acceptable medium for balance

Raikyr0:
> i would argue that the _only_ thing platelets should do is improve clotting as a corollary to hemophilia, and blood regeneration should be a separate trait (corollary to blood deficiency) while other regen types are their own traits as well
> 
> i'll also note that i've been told that platelets have specifically been _buffed_ for floof so if anything just reverting the buff would seem an acceptable medium for balance

Actually it's the reverse, it's been nerfed compared to the higher chaos upstream. Some values are lower.

EE - Upstream Values
```
          damage:
            groups:
              Brute: -0.35
              Burn: -0.35
              Airloss: -0.25
              Toxin: -0.35
              Genetic: -0.35
```

Mnemotechnician:
No, "it could be worse" is not a valid argument. Besides that, our upstream is targetting LRP gameplay-focused servers nowadays, which is why they have traits like this. This trait was most likely buffed due to the addition of lavaland, something we don't have here.

Mnemotechnician:
> i would argue that the _only_ thing platelets should do is improve clotting as a corollary to hemophilia, and blood regeneration should be a separate trait (corollary to blood deficiency) while other regen types are their own traits as well

That does sound like a good idea.

PR #1356, opened 2025-08-28
Pussy Storage

ashkitten:
wip, not tied to traits atm

# Description

Description.

---

# TODO

- [ ] Task
- [x] Completed Task

---

<p>

Example Media Embed

</p>
</details>

---

# Changelog

:cl:
- add: Added fun :D
- tweak: Tweaked fun
- fix: Fixed fun!
- remove: Removed fun :(

### DIFF
--- a/Content.Shared/_Floof/Traits/GenitalStorageSystem.cs
+++ b/Content.Shared/_Floof/Traits/GenitalStorageSystem.cs
@@ -1,0 +1,57 @@
+using System.Linq;
+using Content.Shared._Floof.Traits.Components;
+using Content.Shared.Containers.ItemSlots;
+using Content.Shared.Humanoid;
+using Content.Shared.Humanoid.Markings;
+using Content.Shared.Inventory;
+using Content.Shared.Slippery;
+using Content.Shared.Tag;
+
+namespace Content.Shared._Floof.Traits;
+
+public sealed class GenitalStorageSystem : EntitySystem
+{
+    [Dependency] private readonly InventorySystem _inventory = default!;
+    [Dependency] private readonly ItemSlotsSystem _itemSlots = default!;
+    [Dependency] private readonly TagSystem _tag = default!;
+    [Dependency] private readonly MarkingManager _marking = default!;
+
+    public override void Initialize()
+    {
+        base.Initialize();
+
+        SubscribeLocalEvent<GenitalStorageComponent, ItemSlotInsertAttemptEvent>(OnInsertAttempt);
+        SubscribeLocalEvent<GenitalStorageComponent, ItemSlotEjectAttemptEvent>(OnEjectAttempt);
+        SubscribeLocalEvent<GenitalStorageComponent, SlippedEvent>(OnSlipped);
+    }
+
+    private void OnSlipped(Entity<GenitalStorageComponent> ent, ref SlippedEvent args)
+    {
+        if (!AreGenitalsAccessible(ent))
+            return;
+
+        foreach (var slot in ent.Comp.SlotNames)
+            _itemSlots.TryEject(ent, slot, null, out _);
+    }
+
+    private void OnInsertAttempt(Entity<GenitalStorageComponent> ent, ref ItemSlotInsertAttemptEvent args)
+    {
+        if (!AreGenitalsAccessible(ent))
+            args.Cancelled = true;
+    }
+
+    private void OnEjectAttempt(Entity<GenitalStorageComponent> ent, ref ItemSlotEjectAttemptEvent args)
+    {
+        if (!AreGenitalsAccessible(ent))
+            args.Cancelled = true;
+    }
+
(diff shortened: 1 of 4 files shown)


Content.Shared/_Floof/Traits/GenitalStorageSystem.cs
```diff
@@ -0,0 +1,57 @@
...
+
+    private void OnEjectAttempt(Entity<GenitalStorageComponent> ent, ref ItemSlotEjectAttemptEvent args)
+    {
+        if (!AreGenitalsAccessible(ent))
+            args.Cancelled = true;
+    }
+
+    private bool AreGenitalsAccessible(EntityUid uid) =>
+        // inaccessible if wearing a jumpsuit (but not a skirt)
+        (!_inventory.TryGetSlotEntity(uid, "jumpsuit", out var jumpsuit) || _tag.HasTag(jumpsuit.Value, "Skirt"))
```
Mnemotechnician, line 51: Use the slot blocking system here

Mnemotechnician (commented):
Uhhhhhhhhhh...

Content.Shared/_Floof/Traits/GenitalStorageSystem.cs
ashkitten: doesn't seem applicable since we're not equipping a piece of clothing, which is the only thing that the slot blocking system is designed for

Stealthii:
Don't worry we're adding butts as part of this. Fair is fair

Cprn-Sprkl:
As long as the list of insertable things stays very restricted. The moment someone can hide contra in there it starts involving strip searches and a full on consent quagmire.

ashkitten:
> As long as the list of insertable things stays very restricted. The moment someone can hide contra in there it starts involving strip searches and a full on consent quagmire.

contraband and cavity searches are a planned feature of this, barred behind a consent preference (enabling contraband insertion requires you to consent to security cavity searches)

for instances where the only sec on duty is not comfortable performing said cavity searches, we can just require by server rules to empty your orifices on request (unless someone can come up with a nicer solution)

Cprn-Sprkl:
Security that aren't comfortable with it aren't going to ask someone to empty out in the first place and the antag will just say later "They didn't directly ask me to."

Allowing contraband in this storage currently sounds like it is going to cause lots of issues unless someone comes up with something ironclad.

ashkitten:
i don't disagree with that, which is why i put the PR up in its current state in order to foster this sort of conversation. to me, it's obvious that this sort of roleplay situation is desirable to have on an ERP server like this one, so the question becomes "how do we enable that to happen?" and not "should it happen?"

i have some various thoughts like a security scanner which would flag any foreign object inside someone without requiring a manual strip search. what do other people think about this?

Stealthii:
@Cprn-Sprkl which is why, the initial version and implementation of this should exist only to fulfill a commonly requested feature. "Make toys usable".

The purpose of this is not to provide yet more inventory storage that can be abused, but offer a small subset of tagged item types that would make sense to allow for insertion and removal (namely, toys).

The potential for this to be extended in future with soft and hard contraband (think smuggling cigarettes in a prison visit, notes, more delicate items, contraband, etc.) is absolutely interesting, and something that should be discussed around the wider community to see what the thoughts are on this. Until we have *something* that provides the base facility for this to even be a potential, it'll be harder to bridge those discussion points.

Memeji:
> @Cprn-Sprkl which is why, the initial version and implementation of this should exist only to fulfill a commonly requested feature. "Make toys usable".
> 
> The purpose of this is not to provide yet more inventory storage that can be abused, but offer a small subset of tagged item types that would make sense to allow for insertion and removal (namely, toys).
> 
> The potential for this to be extended in future with soft and hard contraband (think smuggling cigarettes in a prison visit, notes, more delicate items, contraband, etc.) is absolutely interesting, and something that should be discussed around the wider community to see what the thoughts are on this. Until we have _something_ that provides the base facility for this to even be a potential, it'll be harder to bridge those discussion points.

The purpose for this may not be used for hiding contra, if the game allows you to place the items in the inventory, people will use it like that.
Being able to stuff toys in your holes is a good idea. IMO this PR could be a good addition and there can certainly be conversation  about the idea of allowing contraband, but until there's something solid for the contraband bit, it would be best to only allow specific items to be put inside.

ashkitten:
i should clarify that in the *current iteration* of this PR it is whitelisted to only items with the Dildo tag, and until the consent issue is figured out will stay that way.

Memeji:
> i should clarify that in the _current iteration_ of this PR it is whitelisted to only items with the Dildo tag, and until the consent issue is figured out will stay that way.

Yep, sounds good! ^^

Content.Shared/_Floof/Traits/GenitalStorageSystem.cs
Mnemotechnician: No, there's an option to invoke it without specifying either

PR #1357, opened 2025-08-28
Slime Inventory Occlusion

ashkitten:
# Description

Other people can no longer access the internal storage of slime people if they're wearing clothes. This allows their internal storage to be somewhat private as long as they have inner/outer clothing on, though certain clothing items with the ShirtlessClothing tag (such as Public Nudity Permit) will allow access through them.

---

# Changelog

:cl:
- tweak: The internal storage of slime people is no longer accessible from the outside while wearing clothes.


Mnemotechnician (approved):
Seems fine now.

M3739 (approved):

PR #1359, opened 2025-08-29
Make Subtle Actually Filterable By All Players.

M3739:
# Description

**It's one thing when it's a bug. It's another when it causes admins to have their sanity slowly leave their body by unintentionally witnessing someone else's smut in the text chat, and they can't filter it out because for some odd reason, they needed a pretty sensitive and powerful permission to do so.**

Why a client-side ability to filter out a channel everyone in-game had access to, behind a admin permission relating to personally identifiable information, is beyond me. But after some debugging, that is indeed the case.

So therefore, I have made the executive decision (with some suggestions from Mnemo) to remove the check entirely, and alter the code a bit so ALL players, admins, observers, and in-game players have the ability to filter out this channel.

---

<p>

## Without Admin, Ghost:
Pasted image 20250828223130
## With Admin, Ghost:
Pasted image 20250828223204
## In-game Player:
Pasted image 20250828223333

</p>
</details>

---

# Changelog

:cl:
- fix: Ghosts can now properly filter out 'Subtle' messages.

### DIFF
--- a/Content.Client/UserInterface/Systems/Chat/ChatUIController.cs
+++ b/Content.Client/UserInterface/Systems/Chat/ChatUIController.cs
@@ -552,7 +552,7 @@ public sealed class ChatUIController : UIController
             if (_ghost is not {IsGhost: true})
             {
                 FilterableChannels |= ChatChannel.Subtle;
-                FilterableChannels |= ChatChannel.SubtleOOC;
+                /*FilterableChannels |= ChatChannel.SubtleOOC;*/ // Floof - Deprecated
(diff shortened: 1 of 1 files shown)


Mnemotechnician (approved):

Mnemotechnician:
My guess is the author thought that'd make the channel itself invisible to admins without PII, which was never the case, obviously

M3739:
> My guess is the author thought that'd make the channel itself invisible to admins without PII, which was never the case, obviously

Right. I don't know how they have their systems over there, but here, it definitely doesn't work that way, sadly.

PR #1360, opened 2025-08-29
Healing Word Trait

Raikyr0:
# Description
Adds the healing word trait.

Why?
Glimmer goes up, and psionics is incredibly weak.
The purpose is to SPIKE the glimmer. and cause problems. Give the Mantis work.

In fancy terms, Healing Word trait will add more reliability to psionics, and with that forces the glimmer to be higher. Thus forcing the Mantis to do work.

# Changelog
:cl:
- add: Added Healing Word


Cprn-Sprkl:
I have some reservations about this. Back when a lot more people used psionics medical was miserable to play because of the prevalence of the healing powers. People would get the power and then just stand around in medical spamming it on any patients that showed up even when things were fairly quiet. If you wanted to play a doctor you were shit out of luck in both mechanics and RP because of this behavior.

Sure, the mantis can ask them to stop and mind break them if they don't, but experience tells me this will lead to friend groups standing around yelling at the mantis that the person was helping the crew so mind breaking them is horribly evil, and the whole time medical is getting robbed in an environment where they don't get nearly as many patients as they should.

Just something to plan around before moving forward with this.

Raikyr0:
> I have some reservations about this. Back when a lot more people used psionics medical was miserable to play because of the prevalence of the healing powers. People would get the power and then just stand around in medical spamming it on any patients that showed up even when things were fairly quiet. If you wanted to play a doctor you were shit out of luck in both mechanics and RP because of this behavior.
> 
> Sure, the mantis can ask them to stop and mind break them if they don't, but experience tells me this will lead to friend groups standing around yelling at the mantis that the person was helping the crew so mind breaking them is horribly evil, and the whole time medical is getting robbed in an environment where they don't get nearly as many patients as they should.
> 
> Just something to plan around before moving forward with this.

So, it's very intentional that it causes conflict, (as that builds story and character). While some people might hate the Mantis, they are doing their job. (assuming they are doing their job in the first place.)

META-mahn:
> I have some reservations about this. Back when a lot more people used psionics medical was miserable to play because of the prevalence of the healing powers. People would get the power and then just stand around in medical spamming it on any patients that showed up even when things were fairly quiet. If you wanted to play a doctor you were shit out of luck in both mechanics and RP because of this behavior.
> 
> Sure, the mantis can ask them to stop and mind break them if they don't, but experience tells me this will lead to friend groups standing around yelling at the mantis that the person was helping the crew so mind breaking them is horribly evil, and the whole time medical is getting robbed in an environment where they don't get nearly as many patients as they should.
> 
> Just something to plan around before moving forward with this.

If this is a big issue, why don't we just...put more restrictions on who can get the power? To me it's a power that you should whip out in emergencies because it "costs" nothing except Glimmer.

If the literal entire station can get Healing Word then of course Medical's going to have nothing to do; the first person who gets downed is going to get ten Healing Words dropped on them and get up completely fine. Instead I propose whitelisting, for instance, the following groups for it:

- Command, MAYBE
- Medical
- Selective members in Epistemics (Mantis + Chaplain basically)
- Corpsman
- Secoffs, MAYBE
- Salvage, MAYBE

If this is too restrictive, increase the point cost (by like, a ton) and make a cheaper version of the trait that these groups can pick up. None of these people SHOULD be standing around in medical the entire shift insta-healing whoever walks in. As said in the lore of the ability, people with these powers were regarded as saints, and even in 2525 these people are still strategic assets that should be placed in specialist jobs, or jobs that carry an inherent amount of risk.

Raikyr0:
I don't want to add to many restrictions to it, as the purpose was to, A make psionics more reliable, B cause a conflict with the Mantis, C spike the glimmer.

As for people being regarded as saints, I placed that in as flavor text, the idea of someone in the 1600s using healing magic, would either be burned as a witch or treated as a saint.

The purpose is primarily conflict, Mantis gets actual psionic based work. The Glimmer goes up, the PR I have that makes glimmer events pop at lower levels will become more problematic.

PR #1364, opened 2025-08-29
Let Organs Be Preserved in the Fridge Like Meat

ashkitten:
# Description

As @Raikyr0 pointed out on discord, organs are not preserved in the freezer like meat. This is because they lacked a TemperatureComponent (and AtmosExposed to exchange heat with the air). This PR addresses the issue by adding Temperature and AtmosExposed to the BaseOrganRotting prototype that all rot-prone organs inherit from.

---

# Changelog

:cl:
- fix: Organs no longer rot in the freezer.

### DIFF
--- a/Resources/Prototypes/Body/Organs/base.yml
+++ b/Resources/Prototypes/Body/Organs/base.yml
@@ -9,6 +9,10 @@
   - type: RotInto
     entity: AshOrganic
     stage: 2
+  # Floof: let organs be preserved in cold environments
+  - type: AtmosExposed
+  - type: Temperature
+    currentTemperature: 290 # copied from meat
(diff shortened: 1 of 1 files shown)


Mnemotechnician (approved):
Do you mean in a *freezer*? Because *fridges* are entity storages with a special component that prevent any item stored inside from rotting, regardless of the temperature of the outside air, whereas *freezers* are simply cold rooms. Either way, approving.

ashkitten:
yeah i don't play chef

M3739 (approved):
Yeah, don't know why this wasn't a thing, when bodies have been known to be well-preserved in artic environments.

PR #1365, opened 2025-08-29
Merge Upstream - Fe69fe6, Fixes Pulling Breaking if the Mob that Is Being Pulled Dies

TheRealMasterChief117:
Cherrypicks commit ``fe69fe6`` from https://github.com/Simple-Station/Einstein-Engines/commit/fe69fe6e546e9a1c65a7021a7b91d549ab46723e

This fixes pulls breaking when the mob that is being pulled dies while in crit

:cl: Wizden
- fix: Fixed pulls breaking if the mob you are pulling dies


Mnemotechnician (approved):
Wait, it was setting CanCollide=false on all lying dead mobs?! This explains why you would sometimes see salvage corpses or dead crewmembers fly in space and phase through all walls.

M3739 (approved):
Huh. That explains a bit, actually.

PR #1367, opened 2025-08-29
Don't Leave a Used GrantCqc Component on the Chef

ashkitten:
# Description

Fixes used CQC manual text appearing on examine:
image

---

# Changelog

:cl:
- fix: The chef no longer appears to be "already used".


ashkitten:
there is also a check in Content.Shared/_Goobstation/MartialArts/SharedMartialArtsSystem.CQC.cs:82 which prevents the basic CQC attacks from dealing extra stamina or knockdown if performed by the chef, regardless of if they're in the kitchen or not. i'm unsure whether this was intentional...

Mnemotechnician (approved):
Poor chefs

M3739 (approved):
I have yet to divinate how exactly the system handles the chef, since they are unique in that design-wise, they are *supposed* to have CQC in just the kitchen, and the bar. 
However, the exact mechanisms that govern this have yet to be discovered and understood.

Regardless, this will do for now. Good work.

PR #1369, opened 2025-08-31
Fix Solution Transfer Misprediction

TheRealMasterChief117:
# Description
Merges https://github.com/space-wizards/space-station-14/pull/38999 and https://github.com/space-wizards/space-station-14/pull/34838 from wizden

These misprediction issues were making me angry, so i merged what wizden already fixed

:cl: Wizden
- fix: Solution containers (beakers, glasses, etc) no longer rapidly flicker through different fill level visuals when transferring their contents.


TheRealMasterChief117:
Requires engine update

PR #1361, opened 2025-08-29
Port Autocall From The Den, No. 626

M3739:
# Description

This PR aims to cherry-pick a PR from The Den, No. 626.

626 makes the hourly shuttle call now handled by an OOC vote. Depending on the results of said vote, it will handle automatically recalling the shuttle, or letting it continue on it's merry way.

This also comes with a hard end system, which will be disabled by default.

---

<details><summary><h1>Media</h1></summary>
<p>

image

</p>
</details>

---

# Changelog

:cl: The Den
- add: The hourly shuttle call is now handled via an automated OOC vote.

PR #1372, opened 2025-09-01
Downstream Merge: Vulpstation No.43

M3739:
# Description

The implementation of the shuttle call system we ported over from The Den didn't work as intended. As a result, we have instead decided to apply Vulpstation No. 43, which is superior compared to the The Den's for our codebase.

PR #1361 has been reverted as part of this PR.

---

# Changelog

:cl:
- fix: The shuttle vote system now understands what "no" means.


Memeji:
Re-running tests, onto the 5th one. Last one, the tests crashed it because it took way too long, according to the test. Should be good I think, but I'm trying to make sure the tests pass.

M3739:
We really need to allocate some time to figure out what the hell is going on with the tests at some point...

Memeji (approved):
And on the 6th test run, it FINALLY completes it correctly.

M3739:
agony

PR #1373, opened 2025-09-01
MTR-2

M3739:
This is a revision of the ``codeowners`` file.


M3739:
For those curious, Github's codeowners file requires the individual being added having the necessary permissions for it to be valid. The "DO NOT MERGE" was waiting till the individual in question has been granted the necessary permissions.

Merging now.

PR #1374, opened 2025-09-02
Name Fix for Radar Console

Raikyr0:
# Description
Changes the Radar Console Computer Board to Mass Scanner Computer Board, since it makes a mass scanner.

# Changelog
:cl:
- fix: Changed Radar Console to Mass Scanner, because a radar board makes a mass scanner, why ain't it just a mass scanner board?


M3739 (approved):
Ha.

Mnemotechnician (approved):
I mean, I prefer the name radar personally

M3739:
I had the same thought. But then I remember radar dishes and what not, and then I understand why they sci-fi'd it up as a "mass scanner" and left it at that.

PR #1375, opened 2025-09-02
New Floof Lore Changes - Cybersun Number 1

Raikyr0:
# Description
New Floof Lore just dropped, 
NT only controls Epistemics, 
Cybersun Shareholders in Floof Board, 
Sales of Cybersun suits resumed, 
Cybersun has donated a small amount of tacsuits to security for advertisement.

# Changelog
:cl:
- tweak: Cybersun Shareholders have completed a deal, Tacsuits are once again on sale.
- fix: Added Medical Gas Mask to Surgery Bags.


Resources/Prototypes/Catalog/Fills/Lockers/suit_storage.yml
```diff
@@ -138,10 +138,15 @@
...
     contents:
-        - id: OxygenTankFilled
-        - id: ClothingOuterHardsuitCombatOfficer # DeltaV - ClothingOuterHardsuitSecurity replaced in favour of security combat hardsuit.
+        - id: OxygenTankFilled    
+        - !type:GroupSelector
+          children:
+          - id: ClothingOuterHardsuitCombatOfficer
+            weight: 80
+          - id: ClothingOuterHardsuitSyndieReverseEngineered # Floof - New Lore Cybersun Shareholders
+            weight: 20
```
Mnemotechnician: This should be way lower, like 20 times lower, as an easter egg or something like that. Also, this for some reason removes OxygenTankFilled from the fill?

Mnemotechnician (commented):
I'm fine with returning the reverse-engineered hardsuits to cargo, although we still need to nerf several extremely profitable cargo bounties (and diamonds) to make it work

Resources/Prototypes/Catalog/Fills/Lockers/suit_storage.yml
Raikyr0: The oxygen tank is in the line above it (141) no? Unless it's being weird.

As for the suit, there's only a limited amount of suit lockers, so it'll show up on average once every 3 rounds.

PR #1378, opened 2025-09-02
Unrestrict Command Wardrobe

Cprn-Sprkl:
# Description

Removes command restrictions from almost all basic uniform options.
Requested here: https://discord.com/channels/1255902263667331193/1255992182880473269/1412361140918419557

---
# Media

image

---

# Changelog

:cl: sprkl
- tweak: Restrictions on command wardrobe have been lifted.


Mnemotechnician:
Not gonna lie, I'm not a fan of people treating command jobs as "normal worker but with extra tools/access", and not having even the most baseline dress-code only further promotes such behavior

ashkitten:
As a Mystagogue, the first few minutes of my shift is *always* spent hunting for the AutoDrobe to change outfits.

I don't believe that my choice of outfit influences how I behave in this role, and I would consider myself a pretty good Mystagogue.

M3739:
Note to self: Conjure up some command variants of the neckerchiefs.

Cprn-Sprkl:
> Not gonna lie, I'm not a fan of people treating command jobs as "normal worker but with extra tools/access", and not having even the most baseline dress-code only further promotes such behavior

We haven't really had a baseline dress code other than to be recognizable as that command role. The restrictions on what command can take from loadout have been kind of arbitrary feeling. Some examples: Someone can come in as CMO in a thong and thigh highs, but we lost a solid CMO main because we couldn't let them have a floral shirt. I can run HoS in a blue swimsuit, but can't try to make an outfit with the incredibly professional looking MNK office skirt.

ashkitten:
i'd still like to see this happen

PR #1391, opened 2025-09-05
Fix Ghost Timer Getting Garbled when the Player Returns to Body

Mnemotechnician:
# Description
1 year after I submitted a similar fix on frontier, it's still not fixed upstream. When a player ghosts from an already dead body, the body system would set their death time in RealTime instead of CurTime, so returning to your body and ghosting would set your death time to the server's real time, which, depending on the amount of lag, could be as much as dozens of minutes ahead of the server's simulated time, softlocking you from using the ghost respawn system.

Also fixes a minor exploit in the ghost respawn system.

image
image

# Changelog
:cl:
- fix: Returning to body as a ghost should no longer garble your death time and prevent returning to lobby as a consequence.

### DIFF
--- a/Content.Server/Body/Systems/BodySystem.cs
+++ b/Content.Server/Body/Systems/BodySystem.cs
@@ -51,7 +51,7 @@ public sealed class BodySystem : SharedBodySystem
 
         if (_mobState.IsDead(ent) && _mindSystem.TryGetMind(ent, out var mindId, out var mind))
         {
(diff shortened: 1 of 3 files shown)


M3739 (approved):
> who on EE let this exploit through?

Honestly, considering that the mech PR had a mech shotgun that had the description of  a taser, and calling another shotgun a "carbine," it doesn't surprise me anymore.

Good change.

PR #1396, opened 2025-09-07
Fix CI Test Cascading Failure

M3739:
# Description

This PR aims to address the issues in the CI tests that has led it to be more of a paperweight than anything, greatly hindering our operations and hurting development.

The cascading failure has been addressed by implementing a part of EE 1803, which increases the maximum time for CI tests to complete to 2 hours instead of 20 minutes. This was causing the cascading failure as the pool manager would fold on itself once it was taking longer than the set time.

I am currently working on addressing a `DisposeAsync` issue with `RoundEndTest`.

---

# TODO

- [x] Fix cascading failure
- [ ] Fix `DisposeAsync` in `RoundEndTest` delivering a pair with an unexpected state

---

No CL is needed as this is purely a backend change.

### DIFF
--- a/Content.IntegrationTests/PoolManagerTestEventHandler.cs
+++ b/Content.IntegrationTests/PoolManagerTestEventHandler.cs
@@ -4,7 +4,7 @@
 public sealed class PoolManagerTestEventHandler
 {
     // This value is completely arbitrary.
-    private static TimeSpan MaximumTotalTestingTimeLimit => TimeSpan.FromMinutes(20);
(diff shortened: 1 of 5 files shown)


M3739:
Rerunning CI tests to ensure it isn't a fluke

M3739:
Alright, so far the `DisposeAsync` issue has been proving difficult to accurately reproduce in a debug environment. Seems like  a race condition somewhere somewhat.

Merging now.

PR #1399, opened 2025-09-07
Workaround for `RoundEndTest` Passing an Unexpected Pair State to `DisposeAsync`

M3739:
# Description

Continuation of #1396. This PR attempts to fix another heisentest that is afflicting CI.

`RoundEndTest` is passing a `CleanDispose` state to `DisposeAsync` after being called by the test's `CleanReturnAsync`. Why this method call is going to `DisposeAsync` and not `CleanReturnAsync` in `TestPair.Recycle.cs` is beyond me, but what I do know is that the `CleanDispose` pair state is not supposed to be handled outside of `CleanReturnAsync`. Else, this heisentest happens.

I suspect that this pair is being passed to `DisposeAsync` *while* it is still being handled by `CleanReturnAsync`. Why this happens is unknown.

This PR attempts to fix this issue for now by signalling to the pool manager to deliver a new pair, instead of any available pair.

---

No CL since this is a backend change.

### DIFF
--- a/Content.IntegrationTests/PoolManagerTestEventHandler.cs
+++ b/Content.IntegrationTests/PoolManagerTestEventHandler.cs
@@ -4,7 +4,7 @@
 public sealed class PoolManagerTestEventHandler
 {
     // This value is completely arbitrary.
-    private static TimeSpan MaximumTotalTestingTimeLimit => TimeSpan.FromMinutes(20);
+    private static TimeSpan MaximumTotalTestingTimeLimit => TimeSpan.FromMinutes(120);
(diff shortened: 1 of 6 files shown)


M3739:
1st test run is green. Rerunning to ensure it is not a fluke. Timestamp is 27 minutes.

M3739:
2nd run is green as well.

This workaround seems to be doing well. Finally.

M3739:
One last test. Just to triple check. If it's green as well, finally, we can do what we need to do.

M3739:
Workaround successful. Merging now.

PR #1405, opened 2025-09-09
Hide Votes From Evac Autocall And Refine Vote Localization

M3739:
# Description

This PR aims to solve an issue observed by players and admins alike: the tendency of the community to match the votes, regardless of what that vote actually is. While this may be humorous the first or second times, according to reports, they have been doing this since the last weekend we introduced this system in the first place.

To resolve this, I have altered the voting code with a conditional check that checks for the vote initiator before deciding to forward the vote buttons without vote numbers tied to them. In addition, the vote message and options have been altered to better suit our community's practices, which was suggested by Chokovit.

In the future, I plan on this being a more refined system where it is an option for votes to become a secret ballot, which is essentially what we are making the evac autocall vote become.

---

<p>

image
image

</p>
</details>

---

# Changelog

:cl:
- tweak: The automatic shuttle call has been revised.

### DIFF
--- a/Content.Client/Voting/UI/VotePopup.xaml.cs
+++ b/Content.Client/Voting/UI/VotePopup.xaml.cs
@@ -54,7 +54,19 @@ namespace Content.Client.Voting.UI
             for (var i = 0; i < _voteButtons.Length; i++)
             {
                 var entry = _vote.Entries[i];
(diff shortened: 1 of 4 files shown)


Raikyr0:
Honestly, would be funnier to have a Steve option.

M3739:
Heisentest.

M3739:
Alright. Merging now. If anything breaks or something, feel free to scream at me and call me an idiot.

PR #1406, opened 2025-09-10
Sleeping Carp Update

Raikyr0:
# Description
Updated Sleeping Carp to Goobstation Upstream https://github.com/Goob-Station/Goob-Station/pull/2143
~~Can now be used by 3 people as intended~~
Now friendly to Carps as intended
(Also reduced the use delay)

SleepingCarpPR
Working as intended surprisingly.
# Changelog
:cl:
- tweak: Sleeping Carp makes you immune to carp.

### DIFF
--- a/Content.Shared/_Goobstation/MartialArts/Components/GrantMartialArtKnowledgeComponent.cs
+++ b/Content.Shared/_Goobstation/MartialArts/Components/GrantMartialArtKnowledgeComponent.cs
@@ -33,11 +33,7 @@ public sealed partial class GrantSleepingCarpComponent : GrantMartialArtKnowledg
     [DataField]
     public int MaximumUses = 3;
     public int CurrentUses = 0;
-}
 
-[RegisterComponent]
-public sealed partial class SleepingCarpStudentComponent : Component
-{
     [DataField]
     public int Stage = 1;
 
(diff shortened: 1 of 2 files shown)


Sleepy-Shinu:
The blunt rotation can be real

Mnemotechnician:
Please do not cherry-pick changes by hand.

Raikyr0:
At this point it's less cherry pick and more I had to fix it till it worked.

Mnemotechnician:
It is a cherry-pick none the less. A manual one.

M3739 (commented):
You should not be copy and pasting code instead of cherry-picking it using `git`. On that alone, I voice my disapproval.

In addition, I do not support the idea of letting multiple people read a single scroll. We're not ready for that. And we won't be for a long while.

Raikyr0:
> You should not be copy and pasting code instead of cherry-picking it using `git`. On that alone, I voice my disapproval.
> 
> In addition, I do not support the idea of letting multiple people read a single scroll. We're not ready for that. And we won't be for a long while.

I have no clue how to cherrypick from the actual goobstation codebase, as for the Sleeping Carp being used by multiple people, it exists in a limbo of, costs 35 TC meaning 2 people have to contribute to buy it, then only usable by 1 person, and Krav Maga instantly beats it. I do not see sleeping carp being used at all and that's not really the ideal.

Chokovit:
> instantly

I think where Sleeping Carp is now is fine. I do not at all under any circumstances think it should be able to be facilitated and encouraged to be used by multiple people. Despite having a counter it is still extremely strong and it SHOULD be.

I think the position its in now is more then ideal. I do not want to see Sleeping Carp used as any kind of legitimate meta strategy.

PR #1407, opened 2025-09-10
Epistemics Gaming

Raikyr0:
# Description
Greatly increase alternative sources of research points for epistemics.
The reward for your department exploding and you dying was too low.
Also increased the gain on Prober so it's not just a syndie bomb pretending to be something useful.
Updated anomaly point values to upstream ones.

Why?
To be competitive with optimized 6x Experimental Anomaly Vessel generating 1200 points a second for high end research.

# Changelog
:cl:
- tweak: Artifacts and Glimmer Probers give more research


ArcanissKenku:
I think the increase is way too high for two things available at round start.

100 points per second for a prober puts it on par with a maxed anomaly on a standard vessel. 

And 25,000 points per node on an artifact means you're unlocking five lowest tier techs with that amount of points from round start.

Mnemotechnician:
Just, big no on all of this. 60k research points for a depth 4 node, and more for deeper ones? 1000-1200 research points a second for a good anomaly (growing with a 170-195% behavior and attached to an experimental vessel)? 50 points/second for absolute free?

PR #1413, opened 2025-09-11
Event Stuff

Raikyr0:
# Description
Makes it so hostile mob events can't spawn without 2 security personnel and upped higher risk events to require 4. 

# Changelog
:cl:
- tweak: Safety Inspector came by and checked the vents. There's nothing in there.


Mnemotechnician:
I'd argue low-risk events should not have such a requirement, simply because the threat they pose is so low tiders usaully beat all the critters up way before security even has a chance to show up, but that needs discussion.

PR #1418, opened 2025-09-13
Add Evac Vote Results Feedback And Fix Evac Vote Bugs

M3739:
# Description

This PR aims to add visible feedback regarding the results of the shuttle evac vote in the form of announcements.

It also establishes a tiebreaker protocol, where command is expected to break the tie, where either they recall the shuttle, or simply let it arrive, in a timely manner.

This also addresses a bug where the evac shuttle call stops happening if the first vote is a stay. This is due to the flag that signals to the system to try again another time, `_autoCalledBefore`, not being set as it was coded to only become `true` upon a shuttle being *recalled.*

---

<p>

image
image
image

</p>
</details>

---

# Changelog

:cl:
- add: The evac shuttle vote now shows feedback showing the results of the vote.
- fix: The evac shuttle vote no longer ceases operations if the first vote's winner is 'stay.'

### DIFF
--- a/Content.Server/RoundEnd/RoundEndSystem.cs
+++ b/Content.Server/RoundEnd/RoundEndSystem.cs
@@ -393,7 +393,19 @@ namespace Content.Server.RoundEnd
 
                 Action act = () =>
                 {
-                    RequestRoundEnd(null, false, "round-end-system-shuttle-auto-called-announcement");
+                    RequestRoundEnd(null, false, "floof-round-end-system-shuttle-autocall-announcement"); // Floof
+                    _autoCalledBefore = true;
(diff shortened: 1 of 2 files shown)


ashkitten:
How difficult would it be to automate the tiebreaker vote?

M3739:
> How difficult would it be to automate the tiebreaker vote?

Right now as far as I can see, the voting system doesn't exactly allow for it to be narrowed down based on manifest, ID, PDA, etc. It seems to have been designed as a server-wide OOC vote that operates on a simple majority system.

Mnemotechnician (approved):

PR #1419, opened 2025-09-13
Lead to Chemvend

Raikyr0:
# Description
I noticed there was NO way to obtain Lead in game (only way was the Nuclear Operative Chemvend which requires a full nuke ops to spawn which is impossible). Sure it's a deadly toxin but it's also a Sweetener (used in the Victorian era) and on the periodic table.

So this adds Lead to the Contraband Inventory of the Chemvend, and Vestine to the Emag Inventory (Buff to Chemist Syndies, as Medical has no unique items to buy)
Also adds Lead to the Boozeomat and Chefvend Emag Inventory to assist Bartender and Chef Syndies. 

# Changelog
:cl:
- tweak: Added Lead to Chemvend Contraband Inventory (where you cut the management wire)

### DIFF
--- a/Resources/Prototypes/Catalog/VendingMachines/Inventories/boozeomat.yml
+++ b/Resources/Prototypes/Catalog/VendingMachines/Inventories/boozeomat.yml
@@ -51,3 +51,4 @@
     DrinkAngobittersBottleFull: 3 # DeltaV - Angostura Bitters Bottle
   emaggedInventory:
     DrinkPoisonWinebottleFull: 2
+    LeadChemistryBottle: 2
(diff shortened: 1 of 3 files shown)


Mnemotechnician:
Lead is an incredibly effective poison (18 poison per unit) and is the only thing preventing chemists from making licoxide, so no.

Raikyr0:
> Lead is an incredibly effective poison (18 poison per unit) and is the only thing preventing chemists from making licoxide, so no.

The purpose is to allow chemist to make Licoxide, and allow a more subtle poisoning. Do note that Lead also takes an incredibly long time to deal the 18 damage.

This then promotes more work for medical, via either charcoal or liver transplant.

mikusssssss:
what if you just made it give you 1 lead per steel sheet when grinded

M3739:
That's going to be a no from me as well.

PR #1422, opened 2025-09-13
Fix Silly Resomi Plushie Having a 100% Drop Chance

Mnemotechnician:
This isn't even worth describing.

:cl:
- fix: The silly resomi plushie will no longer drop from every prize ball.


M3739 (approved):
The silly has run it's course.

PR #1424, opened 2025-09-14
Change Listplayers Command Permissions to Require the PII Flag (Wizden #40324)

M3739:
# Description

I don't think I even need to supply a explanation nor justification for this cherry-pick.

But for the uninitiated, the `listplayers` command also displays the player's IP address alongside their name, currently without needing PII perms. You can connect the dots to know why that is horrible.

---

No CL is needed.


Memeji (approved):

M3739:
Merging now.

PR #1427, opened 2025-09-16
Stealthy Syndi Key

Cprn-Sprkl:
# Description

Syndicate key can't be found by glancing at someone's headset. This will help encourage agents to actually use them without fear of someone outing them with two button presses.

The media is from when the default channel text was being eaten on all keys. That issue has been fixed.

---

<p>

Med headset without syndi key:
image
Syndi headset with key inserted:
image
Moving the key to the med headset:
image
Med headset with syndi key inserted:
image

</p>
</details>

---

# Changelog

:cl: Sprkl
- tweak: Syndicate keys are no longer visible while in a headset.

### DIFF
--- a/Content.Shared/Radio/EntitySystems/EncryptionKeySystem.cs
+++ b/Content.Shared/Radio/EntitySystems/EncryptionKeySystem.cs
@@ -241,6 +241,14 @@ public sealed partial class EncryptionKeySystem : EntitySystem
         {
             proto = _protoManager.Index<RadioChannelPrototype>(id);
 
+            //Floof - Hide Syndicate key from examine
+            if (proto.ID == "Syndicate")
+            {
+                return;
+            }
+            ;
+            //Floof - End
+
             var key = id == SharedChatSystem.CommonChannel
                 ? SharedChatSystem.RadioCommonPrefix.ToString()
                 : $"{SharedChatSystem.RadioChannelPrefix}{proto.KeyCode}";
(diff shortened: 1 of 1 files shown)


ashkitten:
I don't like that this completely prevents seeing any default channel. It should prevent the Syndicate channel from being default instead, falling back to the next available channel.

If we want to go for full stealth, adding an Unremovable component to the syndie key would work (though we'd have to be careful to only add it once it's inserted into a headset, since otherwise it would be glued into your hands). This could potentially benefit Syndicate gameplay by requiring a security officer to equip the headset and attempt to speak on the Syndicate channel, revealing to everyone on the channel that it is compromised.

Cprn-Sprkl:
The issue of it blocking all default channels has been fixed in every instance except one edge case. If you remove ALL keys and then insert the syndi key first it will not display a default channel. In all other cases it will now display the default channel just fine.

This PR is not looking to make any changes to gameplay.

Mnemotechnician:
While this will probably work fine, I suggest that you instead add a new field to the encryption key component, something like `public bool Visible = true;`, and override it on the syndicate encryption key (in this case you could optionally keep it visible on the nukie headset, if necessary)

Memeji (approved):
Should work fine, would be a good idea to look into Mnemo's suggestion later down the road.

Cprn-Sprkl:
Spent some time digging into the system a little further. I have a plan for assigning certain channels to certain factions. Encryption key holders can then also be assigned factions so the channel only shows when they match.

It is a bit more involved so I'd recommend this as a bandage fix until I can give that some more time.

M3739 (approved):

PR #1429, opened 2025-09-18
Kill the Raise Glimmer Objective

KyuPolaris:
# Description

Kills the Raise Glimmer to 500 objective. The objective is either not played at all or it is done by spamming glimmers/powers in an incredibly little to no RP manner. While the aftermath does call rp potentially, it still does too much disruption for the small amount of RP provided.
---

# Changelog

:cl:
- remove: Removed Raise Glimmer objective (pending rework)

### DIFF
--- a/Resources/Prototypes/Nyanotrasen/Objectives/traitor.yml
+++ b/Resources/Prototypes/Nyanotrasen/Objectives/traitor.yml
@@ -30,18 +30,18 @@
 #      - BecomePsionicCondition
 #  - type: BecomeGolemCondition
 
-- type: entity
-  id: RaiseGlimmerObjective
-  parent: BaseTraitorObjective
-  categories: [ HideSpawnMenu ]
-  name: Raise Glimmer.
-  description: Get the glimmer above the specified amount.
-  components:
-  - type: Objective
-    difficulty: 2.5
+# type: entity
+#  id: RaiseGlimmerObjective
+#  parent: BaseTraitorObjective
+#  categories: [ HideSpawnMenu ]
+#  name: Raise Glimmer.
+#  description: Get the glimmer above the specified amount.
+#  components:
+#  - type: Objective
+#    difficulty: 2.5
     #unique: false
-    icon:
-      sprite: Nyanotrasen/Icons/psi.rsi
-      state: psi
-  - type: RaiseGlimmerCondition
-    target: 500
+#    icon:
+#      sprite: Nyanotrasen/Icons/psi.rsi
+#      state: psi
+#  - type: RaiseGlimmerCondition
+#    target: 500


M3739 (approved):
I personally don't object to this change. Combined with the changes in how psionic abilities is acquired, there isn't exactly a good way to achieve this end in an acceptable manner.

KyuPolaris:
With some of the conversation on the thread I made so far, I am happy to comment it to say "disabled pending rework" when a proper method of finding probers and other things can be made easier to hunt down to open more RP. But in my opinion as it stands, it's too LRP to keep going when it disrupts shifts so much to aggravating levels at points

Memeji (approved):

ArcanissKenku:
I actually disagree with this change.

You're taking away a job that I normally have the Mantis handle and that is finding rogue probers. Removing this removes that element completely and taking another big part of what is already a lackluster role of trying to give them work.

If anything there should be more ways to give the Mantis work.

Mnemotechnician:
I do agree that this takes away the only and last thing relevant for a psionic mantis. Abusing psionics as a non-antag would be considered self-antagging, so we aren't going to see any rogues after this change is merged. But at the same time, I agree that most people who get this objective don't do anything creative and just "spam their powers all shift" (some of them even proudly share that in #general !)

I'll remain neutral.

Mnemotechnician:
Probably important to note that nowadays glimmer drains are stupidly easy to acquire (cosmic revenants and planet-side normality ore spawns need to be nerfed BADLY). Whenever glimmer goes above 0.5 PSI, epi usually scrambles around and gets a drainer or ten, rendering the objective impossible without just blowing up the drainers. This is not fun for antags either.

ashkitten:
i think the issues with this will become less severe once there are more personal downsides for spamming abilities (iirc "neural heat" is something upstream?)

KyuPolaris:
> I actually disagree with this change.
> 
> You're taking away a job that I normally have the Mantis handle and that is finding rogue probers. Removing this removes that element completely and taking another big part of what is already a lackluster role of trying to give them work.
> 
> If anything there should be more ways to give the Mantis work.

I personally do not like the gameplay of running around the station like a chicken with my head cut off with nothing to assist in my search at all. This is a temporary change until something is made, being an antag and messing with glimmer is still possible, it can be used as a distraction instead of having to hard push glimmer in a random corner of the map.

M3739:
From how I view it, it largely boils down to the issue of psionics from a design standpoint and how it interacts with the rest of the station and it's inhabitants.

For those new, raising glimmer isn't exactly easy, especially in face of how easy it is to gather glimmer drains to combat it, and the lack of viable sources that allow for some degree of roleplay. Indeed, so far I have witnessed either one of two things: either they spam their newly found powers to the absolute limit, often gambling and hoping that epistemics isn't able to build drains in time, or just not trying the objective at all and doing whatever is to their hearts content.

I... never liked the system. I found it *hard* to like the system. I hope in the future that with the psionic reworks in upstream, it actually becomes more worthwhile and more interesting to actually dedicate to.

Mnemotechnician (approved):
Going to merge this for now. This objective works well on forks where rounds last 2 hours or less, but not on floof, where rounds often stretch up to twice as long and epi has many chances to recover.

PR #1431, opened 2025-09-21
Sericulture Fix

Raikyr0:
# Description
Adds an identical egglaying system for Sericulture, reason why is that if you have sericulture and egglaying I was told by 1 person that it broke. (The one person who took spinnette and egglaying)

# Changelog
More or less backend stuff

### DIFF
--- a/Content.Client/Sericulture/EgglayingSystem.cs
+++ b/Content.Client/Sericulture/EgglayingSystem.cs
@@ -1,0 +1,9 @@
+using Content.Shared.Sericulture;
+
+namespace Content.Client.Sericulture;
+
+/// <summary>
+/// <So, the reason why this is being added is because if you take 2 things that add Sericulture it breaks.>
+/// <inheritdoc/>
+/// </summary>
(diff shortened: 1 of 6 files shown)


Mnemotechnician (changes requested):
You shouldn't be introducing duplicate systems. You should either fix the underlying issue (STOP USING COMPONENTS AS MARKERS TO TELL AN ENTITY SYSTEM TO PATCH THE ENTITY, use trait functions instead!), or make the problematic traits mutually exclusive.

Raikyr0:
> You shouldn't be introducing duplicate systems. You should either fix the underlying issue (STOP USING COMPONENTS AS MARKERS TO TELL AN ENTITY SYSTEM TO PATCH THE ENTITY, use trait functions instead!), or make the problematic traits mutually exclusive.

Not sure how to fix the underlying issue, but according to 1 user, egglaying which uses the same system was causing an issue with silk weaving. Which would mean Arachne/Arachnid can't use the egglaying trait.

PR #1438, opened 2025-09-23
Age Limit

Raikyr0:
# Description
Just a PR to increase the average max age races can be, reasonable to believe that in 500 years life has improved enough people can live longer.

Dionae and Slimes can now be up to 1000, as they're biologically immortal.

Mostly flavor for station records and peoples roleplay though.

# Changelog
:cl:
- tweak: Max Age for normal races increased to 200, Dionae and Slimes can now be up to 1000. There are no downsides to being old, not in the 25th century.


ashkitten:
on SPLURT this is configurable as "chronological age" and "visual/biological age", i think it'd be nice to do something similar here to support say, vampires who live for hundreds of years but look to be in their twenties (thinking of a specific one on floof)

Mnemotechnician:
This will make it so that 50 year old humans will be considered "young", so it's a no from me.
Besides that, this PR has no positive sides as character age doesn't currently have any effects on your character (and for pure roleplay purposes, you can roleplay being whatever age you want)

Raikyr0:
> This will make it so that 50 year old humans will be considered "young", so it's a no from me. Besides that, this PR has no positive sides as character age doesn't currently have any effects on your character (and for pure roleplay purposes, you can roleplay being whatever age you want)

```
    /// <summary>
    ///     Characters younger than this appear young.
    /// </summary>
    [DataField]
public int YoungAge = 30;
```

Actually it'll remain exactly the same with the max age increased.

M3739 (commented):
I'm not against raising the maximum possible age to 200. But why Diona in particular being able to have their age set to be much higher?

Raikyr0:
> I'm not against raising the maximum possible age to 200. But why Diona in particular being able to have their age set to be much higher?

Many reasons, one of em being tree people, second being they're a weird diona nymph hivemind entity, it would not be weird if they were biologically immortal or close to.

M3739:
> Many reasons, one of em being tree people, second being they're a weird diona nymph hivemind entity, it would not be weird if they were biologically immortal or close to.

Well, that is EE lore IIRC. And we're not following that AFAIK.

This opens up a bit of a gateway: If we do this for Diona, then Vox and Shadekin will require similiar treatments if we're going off of their respective lore.

Raikyr0:
> > Many reasons, one of em being tree people, second being they're a weird diona nymph hivemind entity, it would not be weird if they were biologically immortal or close to.
> 
> Well, that is EE lore IIRC. And we're not following that AFAIK.
> 
> This opens up a bit of a gateway: If we do this for Diona, then Vox and Shadekin will require similiar treatments if we're going off of their respective lore.

I have a vague understanding about Vox lore and its they swap bodies before old age so biologically they're like always young, this is more of a station records needs Rimworld Chronological Age for them.

As for Shadekin, I have no clue what Trystan wants and I ain't touching it lest he gets angry about it.

Mnemotechnician (approved):

PR #1445, opened 2025-09-27
Predict Modify Underwear Verbs and Minor Code Cleanup

TheRealMasterChief117:
# Description

Simplifies the code where I could and makes the code a bit more readable.
Moves the systems to shared so the verbs can be predicted.
Works exactly the same as before.

---

<p>

https://github.com/user-attachments/assets/bc34b5f1-111b-485a-9547-c7bb4995d97c

</p>
</details>

---

# Changelog
It's not player-facing yo


M3739 (approved):
The schema test fails are unrelated nor heisentests. Looks good to me.
Would like @Mnemotechnician's inputs on this, as they are a bit more versed on this sort of stuff than I am.

Content.Shared/_Floof/ModifyUndies/ModifyUndiesSystem.cs
```diff
@@ -0,0 +1,188 @@
...
+    private void AddModifyUndiesVerb(Entity<ModifyUndiesComponent> ent, ref GetVerbsEvent<Verb> args)
+    {
+        if (args.Hands == null || !args.CanAccess || !args.CanInteract)
+            return;
+
+        if (!TryComp<HumanoidAppearanceComponent>(args.Target, out var humApp))
+            return;
+
+        if (args.User != args.Target && _inventory.TryGetSlotEntity(args.Target, "jumpsuit", out _))
+            return; // mainly so people cant just spy on others undies *too* easily
```
Mnemotechnician, line 50: I'd recommend to use the slot blocker here because it's way more versatile. It will also come with the added benefit of preventing a player from toggling someone's underwear if they are e.g. wearing just an outer clothing piece (unless it's tagged with Skirt iirc)

Note that this will require adding some `using` statements, so please don't just blindly use "commit suggestion" here.
```suggestion
        if (!TryComp<HumanoidAppearanceComponent>(args.Target, out var humApp) || !TryComp<InventoryComponent>(args.Target, out var inventory))
            return;

        if (args.User != args.Target && if (_slotBlocker.IsSlotObstructed((args.Target, inventory), null, SlotBlockerSystem.CheckType.IgnoreBlockerPreference, SlotFlags.INNERCLOTHING, out _))
            return; // mainly so people cant just spy on others undies *too* easily
```

Content.Shared/_Floof/ModifyUndies/ModifyUndiesSystem.cs
```diff
@@ -0,0 +1,188 @@
...
+{
+    [Dependency] private readonly MarkingManager _markingManager = default!;
+    [Dependency] private readonly SharedPopupSystem _popupSystem = default!;
+    [Dependency] private readonly SharedAudioSystem _audio = default!;
+    [Dependency] private readonly SharedHumanoidAppearanceSystem _humanoid = default!;
+    [Dependency] private readonly SharedDoAfterSystem _doAfterSystem = default!;
+    [Dependency] private readonly InventorySystem _inventory = default!;
+    [Dependency] private readonly IEntityManager _entMan = default!;
+    [Dependency] private readonly INetManager _net = default!;
+
```
Mnemotechnician, line 29:
```suggestion
    [Dependency] private readonly SlotBlockerSystem _slotBlocker = default!;

```

Mnemotechnician (commented):
Everything seems fine (the code is definitely not the best, but that's whatever). I have a suggestion though.

PR #1458, opened 2025-10-02
Glimmer Event

Raikyr0:
# Description
Adds 2 new Glimmer events which spawns an amount of flora or flesh anomalies on station based on high glimmer.
It uses the glimmer mob rule so it will spawn based on the amount of psychics on station, 1 per 10 psychics. So on average, around 3.

Also reduced the amount of glimmer drained by events, and lowered the threshold in which mites and wisp can spawn.

Why?
People were asking for Glimmer to be more dangerous.
This will give security and epistemics more work. (They begged)

# Changelog
:cl:
- add: Warning. Psionic incursion detected. Realspace manifestation imminent.


Mnemotechnician (changes requested):
1. New prototypes need to be placed under the _Floof namespace
2. I don't particularly agree with reducing the minimum glimmer threshold for glimmer mites. While that would indeed create some work for epi to do, it would be impossible to avoid lest epi do any work at all.

PR #1459, opened 2025-10-02
Healing Word

Raikyr0:
# Description
Adds the healing word trait.

Why?
Glimmer goes up, and psionics is incredibly weak.
The purpose is to SPIKE the glimmer. and cause problems. Give the Mantis work.

In fancy terms, Healing Word trait will add more reliability to psionics, and with that forces the glimmer to be higher. Thus forcing the Mantis to do work.

# Changelog
:cl:
- add: Added Healing Word to Traits


M3739 (commented):
Allowing round start healing word is something I do not approve of. The possibility of medical becoming largely redundant due to the presence of this trait is not something I'd like.

Glimmer being raised due to this addition is not an adequate solution, or even a step to achieving a solution, to the issue that plagues psionics as a whole. An issue that is with the design in it's entirety as it stands.

Raikyr0:
> Allowing round start healing word is something I do not approve of. The possibility of medical becoming largely redundant due to the presence of this trait is not something I'd like.
> 
> Glimmer being raised due to this addition is not an adequate solution, or even a step to achieving a solution, to the issue that plagues psionics as a whole. An issue that is with the design in it's entirety as it stands.

Behavior-wise, from what I saw of Goob's MRP branch when this was added. 
Was Medical ended up looking like a fantasy medbay where every doctor knew healing word. (Bless the Goobmed Cabal was fucking awful at medical) Before the Psionics Refactor where they just let people pick nearly every spell into code whites.

PR #1471, opened 2025-10-06
Increased EG-4 Required Time From 3 Days to 2 Weeks

Raikyr0:
# Description
Apparently the EG-4 has been problematic as every security officer has just been taking it, even tho really it's supposed to be a nice reward unlocked through long service. Which then asks the question, do our security simply play 12 hours everyday as security like a real ass job or is that just edge cases.

Anyway this PR changes the required hours to take an EG-4 from 72 hours to 336 hours or 2 weeks. If the EG-4 is still a problem, I'm going pump it to a month.

# Changelog
:cl:
- tweak: Increased the hours required to take an EG-4

### DIFF
--- a/Resources/Prototypes/Loadouts/Jobs/Security/weapons.yml
+++ b/Resources/Prototypes/Loadouts/Jobs/Security/weapons.yml
@@ -1047,7 +1047,7 @@
     group: LoadoutSecurityWeapons
   - !type:CharacterDepartmentTimeRequirement
     department: Security
-    min: 259200 # 3 days
+    min: 1209600 # 14 days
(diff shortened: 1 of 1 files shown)


Memeji:
Sorry, but I'm going to have to say that the issue with the Energy revolver isn't about the time requirement. Sec will still be able to all have this weapon eventually again and we'll be back to square one. I appreciate wanting to add something that rewards long-time Sec players. The best way to do that through the loadout, imo, is by making a special reskin of a gun that they already have.

Since the Energy Revolver is going to be removed from Sec's loadout, I will go ahead and close this PR.

PR #1473, opened 2025-10-06
Language - Onis

Raikyr0:
# Description
So, we're changing Oni to speak Kagebun (Lore Reason)
And Felinids which are just Humans to speak Sol Common.

I've done something interesting here where, Felinids can still speak Nekomimetic, which is their weird weeb japanese. But Onis and Kitsune also understand it but can't speak it since it's a bastardized Kagebun. The same way a tourist speaks your native language but really awfully, and you understand them but you don't want to stoop to their level.

# Changelog
:cl:
- tweak: Onis to speak Kagebun and Felinids to speak Sol Common.


Chokovit:
Gets a thumbs up from me

Mnemotechnician (approved):
Seems fine.

M3739 (approved):
I don't mind this.

PR #1475, opened 2025-10-07
Re-Adding Xenoglossy Trait

XavierSomething:
# Description

This adds the Xenoglossy trait back as an available trait to Floof. This trait allows you to hear all languages, and you also can speak in a way that is understood by all. 

The reason why this was commented out before are overshadowed in a few ways: 

- First of all, the privacy issue. With the addition of NanoChat, privacy for speaking is easily obtained, and you can also just whisper. Privacy as an excuse to remove the trait is a poor excuse as if you're speaking aloud, anyone that is the same species can possibly understand you anyhow, so saying that people with xenoglossy understanding all languages is privacy-breaking is a poor argument if you're speaking out loud enough in the first place. With NanoChat added, this reason is more mitigated as privacy for speaking to others is much more easily accessible.
- Second, the opting out of the language mechanic. Yes, this does happen if you were to take the language, but I think this offers some RP benefits as well. There is the obvious of being a translator between all groups, which offers a lot of opportunities for xenoglossy'ed individuals. More importantly is consistent characters, having characters who commonly are cataloguers/librarians to keep their xenoglossy as jobs outside of that makes those characters feel more consistent if their character revolves around that job, rather than just only having it when they are in the job in the first place.
- Third, that people can just choose this trait over taking multiple languages if that PR is pulled. While yes, this is true on trait cost alone, the issue is that first, you have to take psionic traits in general to start getting xenoglossy while you can directly take languages without needing psionics, and even so, the downside of this over picking languages directly is that you cannot speak the language directly, so taking the language trait would have that overall benefit both mechanically but mainly RP wise.

In regards to balance, I am willing to move around the trait cost, although going higher than 10 does break the trait. I do think it sits well at 8 since it's just language that it mitigates, which does some some gameplay advantages as communication is incredibly important, so having a high value is normal, but with how little it may come up, I say 8 is a good place to leave it. This is open for discussion, though.

---

# TODO

- [X] Uncommented Xenoglossy
- [X] Checked that it works
---

<p>

Image Showing Xenoglossy Trait

</p>
</details>

---

# Changelog

:cl:
- add: Re-Added Xenoglossy as a Trait

### DIFF
--- a/Resources/Prototypes/Traits/mental.yml
+++ b/Resources/Prototypes/Traits/mental.yml
@@ -319,7 +319,7 @@
 - type: trait
   id: XenoglossyPower
   category: Mental
-  points: -12
+  points: -8
   requirements:
     - !type:CharacterLogicOrRequirement
       requirements:
--- a/Resources/Prototypes/Traits/mental.yml
+++ b/Resources/Prototypes/Traits/mental.yml
@@ -316,41 +316,39 @@
       psionicPowers:
         - MetapsionicPower
 
-# Floof - this is an extreme invasion of privacy and is basically a way to opt out from languages.
-# As such, this is commented out for now.
-#- type: trait
-#  id: XenoglossyPower
-#  category: Mental
-#  points: -12
-#  requirements:
-#    - !type:CharacterLogicOrRequirement
-#      requirements:
-#        - !type:CharacterTraitRequirement
-#          traits:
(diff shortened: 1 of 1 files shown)


XavierSomething:
If it feels too powerful, I can push the trait cost to 10! As stated in the PR, I'm willing to discuss it

PR #1482, opened 2025-10-08
Allow Binding Verb Invocation to Keys

ashkitten:
# Description

Allow `invokeverb self` to be used as a non-admin for bindings.
Add `expand` client-side command to expand certain variables in a command (currently only `$HOVERED`).

Example usage:
```
bind MouseMiddle Command "expand invokeverb self $HOVERED \"Look at\""
```

---

# TODO

- [ ] Should this take a single string arg instead? I wanted to avoid introducing more layers of quoting for no reason.

---

<p>

image

</p>
</details>

---

# Changelog

:cl:
- add: Emote keybind users rejoice! You can now bind right-click verbs to keybinds that will activate on the hovered entity.

### DIFF
--- a/Content.Client/_Floof/Commands/ExpandCommand.cs
+++ b/Content.Client/_Floof/Commands/ExpandCommand.cs
@@ -1,0 +1,68 @@
+using Content.Client.ContextMenu.UI;
+using Content.Client.Gameplay;
+using Content.Client.Viewport;
+using Content.Shared.Administration;
+using Robust.Client.Input;
+using Robust.Client.State;
+using Robust.Client.UserInterface;
+using Robust.Client.UserInterface.CustomControls;
+using Robust.Shared.Console;
+
+namespace Content.Client._Floof.Commands;
+
+[AnyCommand]
+internal sealed class ExpandCommand : LocalizedCommands
+{
+    private const string Hovered = "$HOVERED";
+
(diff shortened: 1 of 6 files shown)


Mnemotechnician (approved):
This should probably be fine

Stealthii:
So there's the potential for consistent pre-processing to be used here, to not only make what we expose safer, but provide a lot easier verbage for people to achieve what they want.

I've personally been using <angle brackets> in my design doc for future extension to chat substitution, as I feel it reads cleaner and are less "programmery" for people to understand.

So my only suggestion is to replace `$HOVERED` with `<hovered>` and/or `<target>`.

Examples of future ideas:

* `"me waves <her> hand in the air"` (auto pronoun replacement for his/her/its/they)
* `"me twirls the <held> in <his> hand"` (auto-insertion of active hand's contents
* `"subtle puts a hand on <target>'s face"` (need I say more)

Mnemotechnician:
1. The $VAR syntax is the standard across all toolshed commands and is used by all non-toolshed commands with variables (`do`, for example)
2. It is the standard across most programming languages
3.  The angle bracket syntax doesn't make it any more intuitive. In fact, it's the opposite because angle brackets are generally used in specifications to denote required syntax.

M3739 (approved):

PR #1492, opened 2025-10-20
Add Resomi Laughs From Rmc14

BasedUser:
# Description

ummmm uhhhh yea apparently goob had these and misattributed them to floof but it was awkward that we didn't actually have this sound, plus there's a lot more resomi characters here than anywhere else i've seen
(best practice is probably to just give in and _RMC those sounds but i'm not feeling it)

---

# Changelog

:cl: router
- add: Ported Resomi laughs from RMC14.


Mnemotechnician:
Sounds like the sound of a car alarm system being engaged

Mnemotechnician (approved):

PR #1493, opened 2025-10-21
Magician's Tophat

Raikyr0:
# Description
MagicTricks
Increases the size of the normal Magician's tophat to hold atleast 1 Andy Stall. 3 by 3.
Adds a Syndicate Magician's tophat which lets you store much more.
Uplink pricing is rather cheap because it's not very space proof or armored.

Now you can do magic tricks and disappearing acts. The potential for trickery.

On a unrelated note, doing this via visual code studio was awful.

# Changelog
:cl:
- add: A magician never reveals their secrets.
- tweak: Magician's tophat has a little more space, fits atleast 1 Andy Stall.


Mnemotechnician (approved):
This is probably fine

PR #1494, opened 2025-10-22
Drinks Vending Machine Update 22/10/25

Raikyr0:
# Description
Adds Water Bottle to all soda machines, because in real life they do actually have bottled water.
Increased the inventory of vending machines by on average 66%, as it's weird the default stock is that low.
Changed the Emagged inventory to a Contraband one, both drinks are very whatever.

To do later, do the same to snack machines.
# Changelog
:cl:
- add: Added Water to Soda Vending Machines. because we've all seen real vending machines with bottled ~~microplastics~~ water.
- tweak: Increased the inventory of soda vending machines to be a little more realistic.
- tweak: Changed the emagged inventory of soda machines to a contraband one.


Chokovit:
What specifically is in the emagged vending machines?

Raikyr0:
> What specifically is in the emagged vending machines?

Changeling Sting, which is just a soft drink with no effects, the Can however changes form. 
And Blood Red Brew, which is ice cold nuclear cola.

Chokovit:
> > What specifically is in the emagged vending machines?
> 
> Changeling Sting, which is just a soft drink with no effects, the Can however changes form. And Blood Red Brew, which is ice cold nuclear cola.

OHHHH okay that's what you meant

Mnemotechnician (approved):
Seeing how soda machines always run out of drinks past the 2 hour mark and how no one refills them, this is probably a good change.

PR #1496, opened 2025-10-22
Katana Buff

Raikyr0:
# Description
Makes Katanas 2 hand-able, and adds a very low chance to parry projectiles (probability wise, you will die before you reflect a single bullet). How practical this will be, is not very.

Banzai!

# Changelog
:cl:
- tweak: Katanas are now 2 hand-able.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Melee/sword.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Melee/sword.yml
@@ -90,12 +90,20 @@
     heavyStaminaCost: 15
     maxTargets: 1
     angle: 20
+  - type: Wieldable # Floof - Banzai!
+  - type: IncreaseDamageOnWield
+    damage:
+      types:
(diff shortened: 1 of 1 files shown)


Mnemotechnician:
I'm not sure about a 160% damage output on wield. This increases the dps output of a katana from 18 to 30, and even more when wielded by an oni.

Raikyr0:
> I'm not sure about a 160% damage output on wield. This increases the dps output of a katana from 18 to 30, and even more when wielded by an oni.

It's intended that you put in more strength by using both hands, forgoing the safety of a shield in the off hand. Of course a Katana, in it's design, is mean to cut down unarmored/lightly armored opponents.

Also yes, while somewhat strong, we need to consider the Samurai did eventually get overshadowed by the GUN. 
TLDR; Personally I wouldn't, but Banzai charging into an L6 or any automatic weapon, while cool, is not a great idea.

PR #1499, opened 2025-10-23
Both Barrels

Raikyr0:
# Description
https://discord.com/channels/1255902263667331193/1255992182880473269/1430992882583605318
(Video)

Adds a Burst Fire option for the Double Barrel shotgun, and the sawn off shotgun which builds out of a double barrel.
It's intended to simulate firing both barrels at once.

# Changelog

:cl:
- add: Added Burst Fire to Double Barrel Shotgun


Mnemotechnician (approved):
that doesnt look safe

PR #1502, opened 2025-10-25
Added Cold Resistance to Coats that Really Should Have Had It by Default

Raikyr0:
# Description
I noticed the Bomber Jacket in the Winterdrobe, didn't have cold resistance.
So I changed the parent.
Then I noticed, a bunch of other coats, which also should have had cold resistance. So I also fixed that.

Note this only does the coats in the base game file, ones added in the DV or any other doesn't get this treatment.

# Changelog
:cl:
- fix: Fixed a bunch of coats to properly be resistant to the cold


Mnemotechnician (approved):
This is probably fine

PR #1503, opened 2025-10-25
Added Natural Lubricant as a Lathe Accelerant

Raikyr0:
# Description
Added Natural Lubricant as an alternative to space lube to speed up lathes.

# Changelog
I'm not documenting this, players can figure this shit out themselves, else people are just gonna do it often.


Mnemotechnician:
Absolutely not until there's anything in place to prevent people from just getting a bucket of this stuff from themselves.

ashkitten:
It's actually really difficult to get in large quantities except from artifacts.

ashkitten:
I don't think it should be quite as good as space lube though.

Raikyr0:
> I don't think it should be quite as good as space lube though.

actually fair it should be worse then actual lube WD-40.
WD-Pussy tho.

Mnemotechnician:
Halving the production times just because someone [did inappropriate act] inside a hyperlathe is too much. If the modifier was something like 0.75 for normal lathes and 0.95 for hypers, then it could pass as a "funny interaction that doesn't really affect gameplay", otherwise it's a pretty exploitable hole in balance.

Raikyr0:
> Halving the production times just because someone [did inappropriate act] inside a hyperlathe is too much. If the modifier was something like 0.75 for normal lathes and 0.95 for hypers, then it could pass as a "funny interaction that doesn't really affect gameplay", otherwise it's a pretty exploitable hole in balance.

Personally, I think lubing lathes is more of QoL thing for normal lathes, it's just slightly faster. But I will agree on the Hyperlathe, due to the fact it does reduce the heat it produces, even if it's just 10% faster.

Mnemotechnician:
It's not "just a qol thing", lube literally increases the production speed 4-fold. This matters a lot when producing power cells and other time-expensive equipment.

ashkitten:
I'm gonna be real, with space lube still being strictly better *and* easier to get, this PR is an easy merge. Natural lubricant doesn't come in the booze-o-mat, and the times I have explicitly put out a bucket trying to solicit donations it took all shift to get maybe 200u. This is a good, easy change and I barely foresee it being used, much less powergamed.

southbridge-fur:
Imagine the Natural Lubricant in Lathes PR being bike shedded like this.

PR #1505, opened 2025-10-27
Logistics - Gun Crate Reorganization - Part 1

Raikyr0:
# Description
I've changed the Logistics labeling for guns and some other things so it's less vague.
So the new naming format I decided on is "Gun Name" "Gun Classification" Crate. 
Laser Rifles now come uncharged in the box, I also wanted the guns to come unloaded but that means making even more prototypes.
WT550 and Enforcer are in single units now, price adjusted accordingly, so it's easier to afford. (and for Logi to carry more crates)
Enforcer crate comes with diversified ammo instead of 3 boxes of lethals, it's now 1 lethals, 1 beanbags and 1 slugs now. (I would have done the same for the WT550 but there's no prototype clip for rubber ammo)
MK58s have been changed to 2000 per pistol down from 2600 per, but still comes in a 2fer, so its a 4000 spesos box for 2. It was costing too much for a supposedly mass produced standard pistol.

Why? It's supposed to make organization and finding things easier.

# Changelog
:cl:
- tweak: WT550 and Enforcer Crates now only come with 1 gun, price adjusted accordingly.
- tweak: MK58s crate is slightly cheaper now, it was costing too much for a mass produced pistol.


Mnemotechnician (changes requested):
Do not change IDs of upstream items. This is just begging to create conflicts with upstream maps/shuttles/cargo items/loot spawners.

Raikyr0:
Fixed the IDs back

PR #1506, opened 2025-10-27
Logistics - Gun Crate Reorganization - Part 2

Raikyr0:
# Description
Part 2 of the Reorganization.
Mark 1, Adjutant and Universals now come 1 per box with price adjusted.
Adjutant Crate comes with various ammo boxes, Universal Crate comes with 2 Rubber clips.
Universal price adjusted from 11250 per gun to 10000 per, just a round down.

It makes it less daunting to buy these guns since they no longer come in bulk. Logi gets to haul more crates.

# Changelog
:cl:
- tweak: Mark 1 Rifle, Adjutant Shotguns and Universal Pistols now come 1 per crate, price adjusted.
- tweak: Universals are now about 10% cheaper.


Mnemotechnician:
I'm not sure about this. The high price is there not because the box contains 2 universals, but to bar cargo techs from casually buying guns for themselves. It's there to make it so that you have to spend a large sum of money at once in order to get even 1 gun.

Raikyr0:
> I'm not sure about this. The high price is there not because the box contains 2 universals, but to bar cargo techs from casually buying guns for themselves. It's there to make it so that you have to spend a large sum of money at once in order to get even 1 gun.

But, Cargo Techs outside of skeleton crew don't have order permissions. So the average cargo tech can't order it normally (since they don't have order perms). Meaning it's the decision of the LO or Inventory Spec.

Raikyr0:
Also reverted IDs to prevent mapping issues.

Mnemotechnician:
Unmarked stale, still needs discussion as Fish is busy.

PR #1511, opened 2025-10-28
Fix Medidrobe

Raikyr0:
# Description
Removed the extra winter boots in the contraband list.

# Changelog
:cl:
- remove: Removed extra winter boots from the medidrobe contraband list.


Mnemotechnician (approved):
weird.

PR #1516, opened 2025-11-03
Fixes Bionic Limbs

Raikyr0:
# Description
This PR adds bionic hands and feet to the respective arm and leg replacement traits, it's very weird that you would have a flesh hand at the end of a bionic arm and this PR fixes that. Installing cybernetics in round already adds the hand and foot.

This PR also disables Bionic Legs for Arachne as it simply does not work due to how the legs are labeled. IE left leg (second).

# Changelog
:cl:
- tweak: Bionic Legs disabled for Arachne as it simply did not work
- fix: Added bionic hands and feet to bionic arms and legs, it didn't make sense you would have a flesh bit attached to the end.

### DIFF
--- a/Resources/Prototypes/Traits/physical.yml
+++ b/Resources/Prototypes/Traits/physical.yml
@@ -902,15 +902,15 @@
       protoId: SpeedLeftLeg
       slotId: "left leg"
     - !type:TraitCyberneticLimbReplacement
-      removeBodyPart: Foot
-      partSymmetry: Left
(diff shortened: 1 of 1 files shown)


Raikyr0:
Minor issue with hands and feet falling off, gonna see WHY this is happening but it's probably just Solidus shitcode.

Raikyr0:
Deemed unlikely to be fixed within the next week

PR #1519, opened 2025-11-05
New Negative Trait - Astral Leak

Raikyr0:
# Description
This PR adds a new mental negative trait, Astral Leak, which turns the player into a Glimmer Source. 
The trait is considered dangerous and likely to cause prosecution from the Mantis.

astralleak

# TODO
- [X] Test It and Produce Media

# Changelog
:cl:
- add: Added New Psionic Negative Trait


Resources/Prototypes/_Floof/psionics.yml
```diff
@@ -25,3 +25,23 @@
...
+  initializeFunctions:
+    - !type:AddPsionicPowerComponents
+      components:
+      - type: GlimmerSource
+    - !type:AddPsionicAssayFeedback
+      assayFeedback: glimmer-leak-power-feedback
+  removalFunctions:
+    - !type:RemovePsionicPowerComponents
+      components:
+      - type: GlimmerSource
```
Mnemotechnician, line 46: I'm not sure how balanced this trait would be in the first place, but if we are adding it, it should produce less glimmer than a prober (maybe like 30% of the normal prober rate? we don't know how many people would decide to take this trait and lock up in dorms)

Raikyr0:
I don't know why it didn't find the field, but the Linter looked like there was a error in the namespace.
Ok I have no clue why it doesn't like the field.

PR #1523, opened 2025-11-07
Unified Backpack Loadout

Raikyr0:
# Description
This PR unifies Backpacks in the Loadout to the single Backpack Tab. With the exception of Security, Command and Paramedic, all backpacks can be found there now. The first 2 will get you searched and the last one is a filled bag with medical tools.

This PR should also fix most issues where you can select both a civilian backpack and a job backpack.
Feature wise, you can now take nearly any backpack as long as you're the job required or have enough hours in the department. (Because there's nothing wrong using your main job bag while taking a break as a passenger.)

Bags
I have no clue why it randomizes in the loadout, but it's evil.

A issue noticed while I was doing this was there roboticist bag in the civilian loadout already but the other 2 weren't which isn't an issue now.

# Changelog
:cl:
- tweak: Backpacks are nearly all in the Backpack Tab


Mnemotechnician:
> I have no clue why it randomizes in the loadout, but it's evil.

Because there's no defined sorting order. If and only if I finish my loadout ui rework, it may get fixed.

PR #1528, opened 2025-11-10
Ported Blinking Emote and Sound with Variation

Jazun:
# Description

Attempt number 2 since I PR'd from master (I barely have a clue what I'm doing 👍 ).
Adds the blinking sound from Delta since it was pretty funny. I also added a higher pitched variant as to not make it too repetitive.

---

<p>

https://github.com/user-attachments/assets/60f75989-3e1f-4283-9756-aad61cfeeabc

</p>
</details>

---

# Changelog

:cl:
- add: Added a sound for blinking


Jazun:
Couple of issues I'm only now seeing is things like the amount of spaces (indents) before some of the lines but I'm not exactly sure or familiar with how much that might affect the performance of the code, if at all

Mnemotechnician (commented):
Has this been tested with other species? I'm not sure we have emote sound parenting yet.

Jazun:
I tested all of the species and it seems to work fine, but it's on EVERY wheel for some reason (including ghosts, which you said was because there's no whitelist / blacklist for the emote) but it doesn't really do anything when it's not meant to (aside from make skeletons scream and mouses squeak)

https://github.com/user-attachments/assets/10fcea27-e97b-4040-b7c1-1cd9bcf6563b

Mnemotechnician (approved):

PR #1531, opened 2025-11-14
Minor Ninja Change

Raikyr0:
# Description
Ninjas now work the Kurokumo clan instead of just the spider clan (flavor so they seem more ninja-ery)
Ninja gear is now also Chameleon Equipment too and they have a voice mask.

Why? Stealth comes in many forms.

# Changelog
:cl:
- tweak: Ninjas have received new lessons and equipment in stealth.


Mnemotechnician:
I don't mind this, but I already constantly hear players complain about how ninjas with pvp experience are basically uncatchable, I'm not sure it's a good idea to make them even more powerful with no drawbacks.

Raikyr0:
> I don't mind this, but I already constantly hear players complain about how ninjas with pvp experience are basically uncatchable, I'm not sure it's a good idea to make them even more powerful with no drawbacks.

Personally, with this I want them to try sneak more or do more spy like stuff, an infiltrator.
That said, they wouldn't be a very good Ninja if they got caught. 
I might consider adding a goal onto Ninja to evac to CC undetected.

PR #1535, opened 2025-11-14
Makes It so Arachne Can Wear the Atmos Fire Suit

Raikyr0:
# Description
Makes it so Arachne can wear the Atmos Fire Suit. 

This is a problem. Either accept we gotta let em wear some work hardsuits or race ban Arachne from Engineering, they're a walking OSHA violation.

# Changelog
:cl:
- tweak: Arachne can now wear the Atmos Fire Suit. Their ass might be a bit toasty tho.


KyuPolaris:
This has been a part of my pr for the Arachne changes since it's conception

PR #1536, opened 2025-11-15
Buff the TEG

BasedUser:
# Description

Flick threatened me with a paddling if I didn't do this. I blame them.
<sub>I am not trapped in a basement waiting for help. Don't come looking for me, I am happy and I'm having fun with my partner. \*sobs\* PLEASE LET ME OU-</sub>

---

<p>

image

</p>
</details>

---

# Changelog

:cl: router
- tweak: Buffed TEG ramping by about 3-ish times. You should no longer see station-wide flickers if Engi turns on 10 hellfire freezers at the same time.

### DIFF
--- a/Resources/Prototypes/Entities/Structures/Power/Generation/teg.yml
+++ b/Resources/Prototypes/Entities/Structures/Power/Generation/teg.yml
@@ -62,7 +62,7 @@
     # It fires processing on behalf of its connected circulators.
     - type: AtmosDevice
     - type: TegGenerator
-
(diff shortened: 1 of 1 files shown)


Mnemotechnician:
Not a fan of this. TEG already has no downsides, there's no reason to buff it further.

BasedUser:
technically this is the one thing that's worse for it than the *AME*, if you ignore needing to know atmos as a downside

i still don't see why this is a bad thing considering 2 RTGs and 13 rad collectors can power almost any station

PR #1537, opened 2025-11-15
Wings - It Works

Raikyr0:
# Description
This PR adds a trait that replaces your arms with harpy wings, which grant flight, however due to unforeseen circumstances, does not change the appearance of your arms to look like harpy wings. Which you can fix by having Wing Markings, also unforeseen, IPCs will appear have no arms or hands. Which I may just disable the trait for IPCs. 

mightbeatadlong
Picture shows the Flight Icon and my character stam crit from flying too long.

# Changelog
:cl:
- add: Added Wings Trait


Raikyr0:
I've certainly gotten it working, it does it deed, grant flight. However, it has the weird issue of, it SHOULD replace your arms with harpy wings in appearance but they appear to be normal arms. Likely an issue from being so far downstream.

But realistically speaking, since it doesn't replace the arms physically in appearance, one can just take Wing Markings and simulate flight. Of course there's no way to check if one has a WING marking.

PR #1542, opened 2025-11-16
New Trait/Loadout Ui Fixes

Mnemotechnician:
# Description
Oopsie.

# Changelog
:cl:
- fix: The trait selections counter no longer shows the same max value as the trait points counter, the loadouts ui now correctly fetches your playtimes when checking if a loadout is usable, and the server now correctly keeps track of trait slots.


Mnemotechnician:
https://github.com/user-attachments/assets/2cb68f90-2e88-4374-94cf-a112acfd0da0

PR #1543, opened 2025-11-16
The Veterinarian Update.

Enturian:
Now most animals and station pets are valid targets for surgery.

List of affected animals and station pets:

Station pets affected:

Ian
Old Ian
Lisa
Puppy Ian
Runtime
Exception
Floppa
Bandito
Bingus
McGriff
Paperwork
Morty (Both new and old)
Poppy
Morticia
Alexander
Renault
Hamlet
Laika
Siobhan

Animals affected (not station pets):

Ferrets
Corgis
Cats (Including syndiecats, I dont know why someone would try this)
Ferrets
Sloths
Possums
Raccoons
Pigs
Cows
Hamsters
Kangaroos

-----------------------------

This is my first PR, first of many.

Explanation on changes:

Using common sense, one would think, "Just add surgery target to all the parent mobs from the station pets," and that's true. The problem is that, apparently, childs don't inherit the surgery interface if they are not sharing the same YAML file, so almost all station pets needed the surgery interface hard-coded in them, with some exceptions like IAN and all his variants.

### DIFF
--- a/Resources/Prototypes/DeltaV/Entities/Mobs/NPCs/pets.yml
+++ b/Resources/Prototypes/DeltaV/Entities/Mobs/NPCs/pets.yml
@@ -18,6 +18,12 @@
     tags:
     - CannotSuicide
     - VimPilot
+  - type: SurgeryTarget
+  - type: UserInterface
+    interfaces:
+      enum.SurgeryUIKey.Key:
+        type: SurgeryBui
+
 
 - type: entity
   name: Laika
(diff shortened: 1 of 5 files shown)


Mnemotechnician:
Please include some media in this PR. This includes (but not limited to) screenshots or recordings of your changes being functional.

Mnemotechnician (changes requested):
Also, this apparently removes the StrippingUiKey from all affected mobs. If you want to preserve it AND add the surgery bui, you need to extend MobStrippableBase, include surgery and stripping bui keys there, and use that new prototype on all affected mobs (just make sure that in the list of parents it comes before any prototype that adds MobStrippableBase)

PR #1552, opened 2025-11-20
Balancing 20/11/25

Raikyr0:
# Description
Rebalances TC cost.

China Lake more expensive, L6 affordable, Sleeping Carp affordable.
China Lake space usage larger.
Added Normal Surplus Crate to uplink.
Minor adjustments to surplus internal value.

Why?
Got a complaint that China Lake was too small (for balancing reason, not realism). So I increased the size of it, and increased the price of it due to how prominent it was as an easy win button. Decreased L6 and Sleeping Carp to be affordable to boost how often we see those as they were greatly overpriced compared to China Lake. China Lakes time in the Limelight is over.

# Changelog
:cl:
- add: Added Normal Surplus Crate for gambling addicts
- tweak: Rebalanced TC cost of unused items and increased cost of China Lake
- tweak: Increased the size of a China Lake in bags, it's shotgun sized but due to balancing it's gotta be bigger

### DIFF
--- a/Resources/Locale/en-US/store/uplink-catalog.ftl
+++ b/Resources/Locale/en-US/store/uplink-catalog.ftl
@@ -241,10 +241,10 @@ uplink-zombie-bundle-name = Syndicate Zombie Bundle
 uplink-zombie-bundle-desc = An all-in-one kit for unleashing the undead upon a station.
 
 uplink-surplus-bundle-name = Surplus Crate
-uplink-surplus-bundle-desc = Contains 50 telecrystals worth of completely random Syndicate items. It can be useless junk or really good.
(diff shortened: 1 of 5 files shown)


Mnemotechnician:
I don't think it's a good idea to lower the cost of the surplus crate. Its whole idea is that you have to pair up with another traitor and gamble all your money in hopes of win big, not just a free way to get 5x the value out of your tc budget.

Raikyr0:
> I don't think it's a good idea to lower the cost of the surplus crate. Its whole idea is that you have to pair up with another traitor and gamble all your money in hopes of win big, not just a free way to get 5x the value out of your tc budget.

This isn't the super surplus, this is the normal surplus. And the value is like 3x.

So, personally, being able to buy a normal surplus, god bless I love gambling, for all your TC is like committing to the god decides my build today and usually like gambling, you get all suits, all allies or all bombs.

PR #1555, opened 2025-11-21
Makes It so Healing Word and Greater Healing Word Can Revive Again

Raikyr0:
# Description
Simple 1 line change to make it so Revivify and Healing Word can revive unrevivable people. 
The description says can't be through conventional means. Unless magic is conventional this should work.

# Changelog
:cl:
- tweak: Tweaked Revivify/Healing Word to work on unrevivable people again.

### DIFF
--- a/Content.Server/Abilities/Psionics/Abilities/HealOtherPowerSystem.cs
+++ b/Content.Server/Abilities/Psionics/Abilities/HealOtherPowerSystem.cs
@@ -133,7 +133,6 @@ public sealed class RevivifyPowerSystem : EntitySystem
             _damageable.TryChangeDamage(args.Target.Value, args.HealingAmount * args.ModifiedAmplification, true, false, damageableComponent, uid);
 
         if (!args.DoRevive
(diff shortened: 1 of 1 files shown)


Arbitrated:
As somebody who uses the Unrevivable trait unironically, we strongly oppose this change. It makes character building with traits even easier (Healing Word is still quite common on stations, even if it can't be directly chosen, which means Unrevivable becomes almost comparable to Unclonable in how often it would be relevant), and on an RP level we would very much prefer that people have to actually work around these sorts of limitations.

Raikyr0:
> As somebody who uses the Unrevivable trait unironically, we strongly oppose this change. It makes character building with traits even easier (Healing Word is still quite common on stations, even if it can't be directly chosen, which means Unrevivable becomes almost comparable to Unclonable in how often it would be relevant), and on an RP level we would very much prefer that people have to actually work around these sorts of limitations.

it can be changed so only greater healing revives honestly, personally i'd prefer if someone who was true unrevivable to have a death acidifer

Mnemotechnician:
I won't even bother explaining, no.

PR #1577, opened 2026-01-15
Fix 15/01/26

Raikyr0:
# Description
Bugfixes that I'm able to fix from the bug report list, the rest seem either esoteric C# issues, FTL jank, and faction issues which I'm unwilling to fix at the moment.

# Changelog
:cl:
- fix: Fixed Tarantula names showing up with only numbers.
- fix: Fixed Exploding Pens not being able to be triggered like a grenade.
- fix: Fixed Vim Harness having no crafting cost.
- fix: Fixed Vim Mech's walkcycle having the wrong noise.
- fix: Fixed Spray Cans having half the amount of spray they should.
- tweak: Allows HAMTR to move in space so hamsters don't just drift off.


Resources/Prototypes/Entities/Objects/Weapons/Bombs/pen.yml
```diff
@@ -1,7 +1,60 @@
...
+  id: PenEmbeddableNoClick
+  abstract: true
+  suffix: NoClick
+  components:
+  - type: LandAtCursor
+  - type: DamageOtherOnHit
+    damage:
+      types:
+        Piercing: 5
+
```
Mnemotechnician: You should change the pen prototype to remove clicking, change its id it to PenNoClick, then introduce another prototype, give it the original pen id and clicking, and make it inherit from PenNoClick. This is just code duplication.

Raikyr0: Ok I should have fixed it like this.

